PR #1164, opened 2022-03-09
Add drop_events.drop_when documentation

sbayer55:
### Description
Updated `drop_events` readme to include `drop_when` and `handle_failed_events` options
Added documentation on Data Prepper Expression syntax.
 
### Check List

### DIFF
--- a/data-prepper-plugins/drop-events-processor/README.md
+++ b/data-prepper-plugins/drop-events-processor/README.md
@@ -24,6 +24,10 @@ Create the following file named `logs_json.log` and replace the `path` in the fi
 
 When run, the processor will filter out and drop all messages.
 
+## Conditional Drop Events
+The `when` parameter can be used to drop selected events. The `when` parameter takes a single String expression. See
+[Data Prepper Expression syntax](../../docs/expression_syntax) for a complete list of features supported by a Data Prepper Expression.
+
 ## Developer Guide
 This plugin is compatible with Java 14. See
 - [CONTRIBUTING](https://github.com/opensearch-project/data-prepper/blob/main/CONTRIBUTING.md)
--- a/data-prepper-plugins/drop-events-processor/README.md
+++ b/data-prepper-plugins/drop-events-processor/README.md
@@ -6,7 +6,8 @@ This is a processor that drops all messages that are passed into it.
 ### Starting in Data Prepper v1.3.0
 * Added support for new configurations options `drop_when` and `handle_failed_events`. See [Configuration Options](#Configuration-Options).
 
-Upgrade information can be found at [how to upgrade from a pre-release version to 1.3.0](#upgrade-to-1.3.0).
+Upgrade information can be found at 
+[how to upgrade from a pre-release version to 1.3.0](#How-to-upgrade-from-a-pre-release-version-to-130).
 
 
 ## Basic Usage
@@ -63,7 +64,6 @@ This plugin is compatible with Java 14. See
 
 ## How To Upgrade Guides
 
-<a name="upgrade-to-1.3.0"></a>
 ### How to upgrade from a pre-release version to 1.3.0
 
 When upgrading only adding `drop_when: true` is required for compatability with Data Prepper 1.3.0. After adding `drop_when: true`
--- a/data-prepper-plugins/drop-events-processor/README.md
+++ b/data-prepper-plugins/drop-events-processor/README.md
@@ -1,6 +1,14 @@
 # Drop Events Processor
 This is a processor that drops all messages that are passed into it.
 
+## Changes
+
+### Starting in Data Prepper v1.3.0
+* Added support for new configurations options `drop_when` and `handle_failed_events`. See [Configuration Options](#Configuration-Options).
+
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/drop-events-processor/README.md
```diff
@@ -1,6 +1,15 @@
 # Drop Events Processor
 This is a processor that drops all messages that are passed into it.
 
+## Changes
+
+### Starting in Data Prepper v1.3.0
```
dlvenable, line 6: The `drop` processor is brand-new in 1.3.0, so we don't need to discuss changes.

docs/expression_syntax.md
```diff
@@ -0,0 +1,195 @@
+**Is your feature request related to a problem? Please describe.**
```
dlvenable, line 1: I think we should just describe the currently supported expression syntax. We can direct users who want to know about upcoming changes to the RFC at the bottom of this document.

docs/expression_syntax.md
```diff
@@ -0,0 +1,195 @@
...
+As part of a larger feature (#522) to support complex condition statements in data prepper there is a need to define a syntax for conditional statements. A conditional statement is a String that is evaluated at runtime and may reference fields within a record.
+
+**Describe the solution you'd like**
+_Terms used throught this document are defined in the Definitions section_
+
+## Supported Operators
```
dlvenable, line 7: I think we can make this documentation quite a bit more straighforward for pipeline authors.

data-prepper-plugins/drop-events-processor/README.md
```diff
@@ -12,19 +21,65 @@ drop-pipeline:
...
+[Data Prepper Expression syntax](../../docs/expression_syntax) for a complete list of features supported by a Data Prepper Expression.
+
 ## Developer Guide
 This plugin is compatible with Java 14. See
 - [CONTRIBUTING](https://github.com/opensearch-project/data-prepper/blob/main/CONTRIBUTING.md)
 - [monitoring](https://github.com/opensearch-project/data-prepper/blob/main/docs/monitoring.md)
+
+## How To Upgrade Guides
+
+### How to upgrade from a pre-release version to 1.3.0
```
dlvenable, line 67: I don't think we should document changes within pre-releases. I think it can cause confusion and add bulk to the important documentation.

docs/expression_syntax.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Reserved for possible future functionality
+Reserved symbol set: `^`, `*`, `/`, `%`, `+`, `-`, `xor`, `=`, `+=`, `-=`, `*=`, `/=`, `%=`, `++`, `--`, `${<text>}`
+
+## Set Initialiser
+Defines a set or term and/or expressions.
+
+Examples
+```
+{1, 2, 3}
```
dlvenable, line 30: For the documentation, I'd like to use more realistic examples. e.g. comparing against status code.

dlvenable (changes requested):

data-prepper-plugins/drop-events-processor/README.md
```diff
@@ -12,19 +21,65 @@ drop-pipeline:
...
+
+When upgrading only adding `drop_when: true` is required for compatability with Data Prepper 1.3.0. After adding `drop_when: true`
```
sshivanii, line 69: Typo: `compatibility`

docs/expression_syntax.md
```diff
@@ -0,0 +1,195 @@
+**Is your feature request related to a problem? Please describe.**
+As part of a larger feature (#522) to support complex condition statements in data prepper there is a need to define a syntax for conditional statements. A conditional statement is a String that is evaluated at runtime and may reference fields within a record.
+
+**Describe the solution you'd like**
+_Terms used throught this document are defined in the Definitions section_
+
+## Supported Operators
+In order of evaluation priority. _(top to bottom, left to right)_
+
+| Operator             | Description              | Data Prepper Version |
```
sshivanii, line 10: Can we have this priority table in an increasing/decreasing order of Data prepper version?

sbayer55: I think the real problem is the table has too much information. The primary goal of this section is to document order of operations. Removing the version and including only supported operators for 1.3.0 will help here.

Similar to the Java Operator Precedence table
image

@sshivanii let me know your thoughts on this.
@dlvenable you also commented, I'd like to hear your input also.

docs/expression_syntax.md
sbayer55: A better solution could be to remove the sections referring to operators not yet supported in 1.3.0

docs/expression_syntax.md
sshivanii: @sbayer55 I like this Java Operator Precedence approach, we can have the supported operators for only v1.3.0 and update the table based on more operators as and when we add it.

sbayer55: I updated the table to fit this suggestion.

dlvenable (approved):

sshivanii (approved):
Thanks for the update, looks good!

PR #1178, opened 2022-03-11
FIX: coercion on json pointer values

### LINKED ISSUE #1003: Data Prepper Expression Evaluator

chenqi0805:
Signed-off-by: Chen <19492223+chenqi0805@users.noreply.github.com>

### Description
This PR fixes coercion on corner cases in jsonpointer value, especially
* Double value -> convert to float
* non literal value -> exception
 
### Issues Resolved
Resolves #1003 
 
### Check List

### DIFF
--- a/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
+++ b/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
@@ -11,9 +11,7 @@ import org.opensearch.dataprepper.expression.antlr.DataPrepperExpressionParser;
 
 import javax.inject.Named;
 import java.io.Serializable;
-import java.util.Arrays;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 
--- a/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
+++ b/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
@@ -12,7 +12,6 @@ import org.opensearch.dataprepper.expression.antlr.DataPrepperExpressionParser;
 import javax.inject.Inject;
 import javax.inject.Named;
 import java.io.Serializable;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Function;
 
--- a/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
+++ b/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
@@ -9,6 +9,7 @@ import com.amazon.dataprepper.model.event.Event;
 import org.antlr.v4.runtime.tree.TerminalNode;
 import org.opensearch.dataprepper.expression.antlr.DataPrepperExpressionParser;
 
+import javax.inject.Inject;
 import javax.inject.Named;
 import java.io.Serializable;
 import java.util.HashMap;
@@ -17,15 +18,11 @@ import java.util.function.Function;
 
 @Named
 class ParseTreeCoercionService {
-    private static final Map<Class<? extends Serializable>, Function<Object, Object>> literalTypes;
+    private final Map<Class<? extends Serializable>, Function<Object, Object>> literalTypeConversions;
 
-    static {
-        literalTypes = new HashMap<>();
-        literalTypes.put(String.class, Function.identity());
-        literalTypes.put(Boolean.class, Function.identity());
-        literalTypes.put(Integer.class, Function.identity());
-        literalTypes.put(Float.class, Function.identity());
-        literalTypes.put(Double.class, v -> ((Double) v).floatValue());
+    @Inject
+    public ParseTreeCoercionService(final Map<Class<? extends Serializable>, Function<Object, Object>> literalTypeConversions) {
+        this.literalTypeConversions = literalTypeConversions;
     }
 
     public Object coercePrimaryTerminalNode(final TerminalNode node, final Event event) {
@@ -62,8 +59,8 @@ class ParseTreeCoercionService {
         final Object value = event.get(jsonPointer, Object.class);
         if (value == null) {
             return null;
-        } else if (literalTypes.containsKey(value.getClass())) {
-            return literalTypes.get(value.getClass()).apply(value);
+        } else if (literalTypeConversions.containsKey(value.getClass())) {
+            return literalTypeConversions.get(value.getClass()).apply(value);
         } else {
             throw new ExpressionCoercionException("Unsupported type for value " + value);
         }
--- a/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
+++ b/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
@@ -12,16 +12,23 @@ import org.opensearch.dataprepper.expression.antlr.DataPrepperExpressionParser;
 import javax.inject.Named;
 import java.io.Serializable;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
 
 @Named
 class ParseTreeCoercionService {
-    private final List<? extends Class<? extends Serializable>> literalTypes = Arrays.asList(
-            String.class,
-            Boolean.class,
-            Integer.class,
-            Float.class
-    );
+    private static final Map<Class<? extends Serializable>, Function<Object, Object>> literalTypes;
+
+    static {
+        literalTypes = new HashMap<>();
+        literalTypes.put(String.class, Function.identity());
+        literalTypes.put(Boolean.class, Function.identity());
+        literalTypes.put(Integer.class, Function.identity());
+        literalTypes.put(Float.class, Function.identity());
+        literalTypes.put(Double.class, v -> ((Double) v).floatValue());
+    }
 
     public Object coercePrimaryTerminalNode(final TerminalNode node, final Event event) {
         final int nodeType = node.getSymbol().getType();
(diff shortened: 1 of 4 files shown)


data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
```diff
@@ -10,18 +10,28 @@
 import org.opensearch.dataprepper.expression.antlr.DataPrepperExpressionParser;
 
 import javax.inject.Named;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
 
 @Named
 class ParseTreeCoercionService {
+    private final List<? extends Class<? extends Serializable>> literalTypes = Arrays.asList(
```
dlvenable, line 19: This can be static.

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
```diff
@@ -42,4 +52,17 @@ public Object coercePrimaryTerminalNode(final TerminalNode node, final Event eve
...
         throw new ExpressionCoercionException("Unable to cast " + obj.getClass().getName() + " into " + clazz.getName());
     }
+
+    private Object resolveJsonPointerValue(final String jsonPointer, final Event event) {
+        final Object value = event.get(jsonPointer, Object.class);
+        if (value == null) {
+            throw new ExpressionCoercionException("Unsupported type for value " + value);
+        } else if (literalTypes.contains(value.getClass())) {
+            return value;
+        } else if (value instanceof Double) {
```
dlvenable, line 62: This approach would be easier to extend if we have other possible types.


```
private static final Map<Class, Function<Object, Object>> literalTypes;

static {
  literalTypes = new HashMap<>();
  literalTypes.put(String.class, Function.identity());
  literalTypes.put(Boolean.class, Function.identity());
  literalTypes.put(Integer.class, Function.identity());
  literalTypes.put(Float.class, Function.identity());
  literalTypes.put(Double.class, v -> ((Double) value).floatValue());
}
```

And here:
```
else if (literalTypes.containsKey(value.getClass())) {
             return literalTypes.get(value.getClass()).apply(value);
```

But, this is all internal and later refactoring is light. So I don't have a strong opinion on it.

dlvenable (changes requested):

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
```diff
@@ -42,4 +52,17 @@ public Object coercePrimaryTerminalNode(final TerminalNode node, final Event eve
         }
         throw new ExpressionCoercionException("Unable to cast " + obj.getClass().getName() + " into " + clazz.getName());
     }
+
+    private Object resolveJsonPointerValue(final String jsonPointer, final Event event) {
+        final Object value = event.get(jsonPointer, Object.class);
+        if (value == null) {
+            throw new ExpressionCoercionException("Unsupported type for value " + value);
```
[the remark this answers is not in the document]
chenqi0805: Right now there is no null value support in the grammar literal types so it is impossible for cx to check jsonpointer value is null in conditional expression. We have two options for null appearing in jsonPointer value:
* throw Exception
* return null and let each operator complain about the arg type.
I think both should work and I do not favor one over the other. @sbayer55 thoughts?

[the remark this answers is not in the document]
dlvenable: So in the current design `/message == 200` would have to fail if `/message` is `null`? I think this is clearly a false expression. I think if we returned a special `NullValue` class the rest of the code would work as expected. Right?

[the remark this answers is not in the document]
dlvenable: Looking at this code (currently in PR), I think the equals operator would work correctly if the `ParseTreeCoercsionService` provided a `new Object()`.

https://github.com/opensearch-project/data-prepper/pull/1177/files#diff-083f9ffc9761303a542e0904c2de3bc5ba2ec7ef612278087bf6da29208ab961R34

We could even do something like this:

```
class NullValue {
  private static NULL_VALUE = new NullValue();

  private NullValue() { }

  public static nullValue() {
    return NULL_VALUE;
  }
}
```

And return:

```
return NullValue.nullValue()
```

dlvenable (dismissed):

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
```diff
@@ -10,18 +10,33 @@
...
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
 
 @Named
 class ParseTreeCoercionService {
+    private static final Map<Class<? extends Serializable>, Function<Object, Object>> literalTypes;
+
+    static {
```
sbayer55, line 22: Could this be moved to a constructor argument and generated in a bean configuration function? Static variable can make classes difficult to unit test.

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
```diff
@@ -10,18 +10,33 @@
...
+        literalTypes = new HashMap<>();
+        literalTypes.put(String.class, Function.identity());
```
sbayer55, line 24: Are all of these keys needed? Could we use something like
```
if (literalTypes.hasKey(value.getClass()))
    return literalTypes..get(value.getClass()).apply(value);
else
    return value;
```

data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/ParseTreeCoercionServiceTest.java
```diff
@@ -97,36 +96,43 @@ void testCoerceTerminalNodeBooleanType() {
...
     }
 
-    @Test
-    void testCoerceTerminalNodeJsonPointerTypeMissingKey() {
-        final String testMissingKey = "missingKey";
-        final String testJsonPointerKey = "/" + testMissingKey;
-        final Event testEvent = createTestEvent(new HashMap<>());
+    @ParameterizedTest
+    @MethodSource("provideUnSupportedJsonPointerValues")
+    void testCoerceTerminalNodeJsonPointerTypeUnSupportedValues(final Object testValue) {
```
sbayer55, line 121: Could `testValue` type be changed to Event and create the events in the `provideUnSupportedJsonPointerValues` method? I think it would make the test more readable.

sbayer55 (changes requested):

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
chenqi0805: Works for me.

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/ParseTreeCoercionService.java
chenqi0805: I think that will not work for Long type which we want to throw exception. Also, for other custom types, it will just return literal objects which I am not sure if we want to support. e.g.
```
/code1 == /code2              | {"code1": StatusCode, "code2": StatusCode}
```

data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/ParseTreeCoercionServiceTest.java
chenqi0805: Actually the same method is used for both jsonpointer and escape jsonpointer test case in which keys are separate. So providing the value only will probably make it more compact and focused on the difference (testValue). Thought?

sbayer55 (approved):

dlvenable (approved):

PR #1335, opened 2022-04-26
Support remaining OpenTelemetry Metrics proto spec features

### LINKED ISSUE #1310: Support latest OTEL metrics features
PR #242 introduced OTEL metrics support for Data-Prepper. However, due to scope constraints and an [outdated Java OTEL protocol library](https://github.com/opensearch-project/data-prepper/issues/1269) not all features could be fully supported. With the upgraded library the following [metrics features](https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto) can be implemented and should thus be supported by Data Prepper:

- Exponential Histograms
- Flags
- SchemaUrl
- ScopeMetrics
- Exemplars

kmssap:
### Description
Add support for remaining OpenTelemetry Metrics features.
 
### Issues Resolved
#1310 
 
### Check List

### DIFF
--- a/build-resources.gradle
+++ b/build-resources.gradle
@@ -7,7 +7,7 @@
 ext.versionMap = [
         junitJupiter : '5.8.2',
         mockito : '3.11.2',
-        opentelemetryProto : '1.7.1-alpha',
+        opentelemetryProto : '0.16.0-alpha',
         opensearchVersion : '1.3.5',
         armeria: '1.19.0',
         armeriaGrpc: '1.19.0',
--- a/data-prepper-plugins/otel-metrics-raw-processor/README.md
+++ b/data-prepper-plugins/otel-metrics-raw-processor/README.md
@@ -10,18 +10,18 @@ processor:
 ```
 
 ## Configuration
-It is possible to create explicit representations of histogram buckets and their boundaries with this plugin. This can be controlled with the following parameters:
+It is possible to create explicit representations of histogram buckets and their boundaries. This feature can be controlled with the following parameters:
 
 ```yaml
   processor:
     - otel_metrics_raw_processor:
-        calculateHistogramBuckets: true
-        calculateExponentialHistogramBuckets: true
+        calculate_histogram_buckets: true
+        calculate_exponential_histogram_buckets: true
 ```
-There are two parameters: `calculateHistogramBuckets` and `calculateExponentialHistogramBuckets`.
-If each parameter is not given it defaults to `false` and buckets are not calculated.
+There are two parameters: `calculate_histogram_buckets` and `calculate_exponential_histogram_buckets`.
+If a parameter is not provided it defaults to `false`.
 
-If `calculateHistogramBuckets` is set to `true`, the following JSON will be added to every histogram JSON:
+If `calculate_histogram_buckets` is set to `true`, the following JSON will be added to every histogram JSON:
 
 ```json
  "buckets": [
@@ -53,7 +53,7 @@ This is an explicit form of the more dense OpenTelemetry representation that is 
 ```
 
 
-If `calculateExponentialHistogramBuckets` is set to `true`, the following JSON will be added to every histogram JSON:
+If `calculate_exponential_histogram_buckets` is set to `true`, the following JSON will be added to every histogram JSON:
 ```json
 
     "negativeBuckets": [
--- a/data-prepper-plugins/otel-metrics-raw-processor/README.md
+++ b/data-prepper-plugins/otel-metrics-raw-processor/README.md
@@ -106,8 +106,8 @@ a scale parameter, offset and list of bucket counts:
 
 The `exponential_histogram_max_allowed_scale` parameter defines the maximum allowed scale for the exponential histogram. Increasing this parameter will increase potential
 memory consumption. See [the spec](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto) for more information on exponential histograms and their computational complexity.
-All exponential histograms that have a scale that is above 10 (or the configured parameter) will be discarded and logged with error level.
 
+All exponential histograms that have a scale that is above the configured parameter (by default, 10) will be discarded and logged with error level.
 **Note**: the absolute scale value is used for comparison, so a scale of -11 will be treated equally to 11 and thus exceed the configured value of 10 - and be discarded.
 
 ## Metrics
--- a/data-prepper-plugins/otel-metrics-raw-processor/README.md
+++ b/data-prepper-plugins/otel-metrics-raw-processor/README.md
@@ -9,7 +9,7 @@ processor:
     - otel_metrics_raw_processor
 ```
 
-## Configuration
+## Configurations
 It is possible to create explicit representations of histogram buckets and their boundaries. This feature can be controlled with the following parameters:
 
 ```yaml
@@ -17,11 +17,14 @@ It is possible to create explicit representations of histogram buckets and their
     - otel_metrics_raw_processor:
         calculate_histogram_buckets: true
         calculate_exponential_histogram_buckets: true
+        exponential_histogram_max_allowed_scale: 10
 ```
-There are two parameters: `calculate_histogram_buckets` and `calculate_exponential_histogram_buckets`.
-If a parameter is not provided it defaults to `false`.
 
-If `calculate_histogram_buckets` is set to `true`, the following JSON will be added to every histogram JSON:
+There are three possible parameters: `calculate_histogram_buckets`, `calculate_exponential_histogram_buckets` and `exponential_histogram_max_allowed_scale`
+If `calculate_histogram_buckets` and `calculate_exponential_histogram_buckets` are not provided they default to `false`. 
+If `exponential_histogram_max_allowed_scale` is not provided it defaults to 10.
+
+If `calculate_histogram_buckets` is not set to `false`, the following JSON will be added to every histogram JSON:
 
 ```json
  "buckets": [
@@ -53,7 +56,7 @@ This is an explicit form of the more dense OpenTelemetry representation that is 
 ```
 
 
-If `calculate_exponential_histogram_buckets` is set to `true`, the following JSON will be added to every histogram JSON:
+If `calculate_exponential_histogram_buckets` is not set to `false`, the following JSON will be added to every histogram JSON:
 ```json
 
     "negativeBuckets": [
@@ -100,6 +103,12 @@ a scale parameter, offset and list of bucket counts:
     "negativeOffset" : 0,
     "positiveOffset : 1
 ```
+
+The `exponential_histogram_max_allowed_scale` parameter defines the maximum allowed scale for the exponential histogram. Increasing this parameter will increase potential
+memory consumption. See [the spec](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto) for more information on exponential histograms and their computational complexity.
+All exponential histograms that have a scale that is above 10 (or the configured parameter) will be discarded and logged with error level.
+
+**Note**: the absolute scale value is used for comparison, so a scale of -11 will be treated equally to 11 and thus exceed the configured value of 10 - and be discarded.
 
 ## Metrics
 This plugin uses all common metrics in [AbstractProcessor](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-api/src/main/java/com/amazon/dataprepper/model/processor/AbstractProcessor.java), and does not currently introduce custom metrics.
--- a/data-prepper-plugins/otel-metrics-raw-processor/README.md
+++ b/data-prepper-plugins/otel-metrics-raw-processor/README.md
@@ -9,6 +9,98 @@ processor:
     - otel_metrics_raw_processor
 ```
 
+## Configuration
+It is possible to create explicit representations of histogram buckets and their boundaries with this plugin. This can be controlled with the following parameters:
+
+```yaml
+  processor:
+    - otel_metrics_raw_processor:
+        calculateHistogramBuckets: true
+        calculateExponentialHistogramBuckets: true
+```
+There are two parameters: `calculateHistogramBuckets` and `calculateExponentialHistogramBuckets`.
+If each parameter is not given it defaults to `false` and buckets are not calculated.
+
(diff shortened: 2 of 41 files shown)


dlvenable:
Thank you @kmssap for this contribution! I will take a look at this.

data-prepper-plugins/otel-proto-common/src/main/java/com/amazon/dataprepper/plugins/otel/codec/OTelProtoCodec.java
```diff
@@ -457,4 +486,291 @@ protected AnyValue objectToAnyValue(final Object obj) throws UnsupportedEncoding
...
+    public static List<Quantile> getQuantileValues(List<SummaryDataPoint.ValueAtQuantile> quantileValues) {
+        return quantileValues.stream()
+                .map(q -> new DefaultQuantile(q.getQuantile(), q.getValue()))
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Create the buckets, see <a href="https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto">
+     *     the OTel metrics proto spec</a>
+     * <p>
```
dlvenable, line 677: The Gradle build is failing because these Javadocs are invalid. Please see #1271 for a possible solution.

https://github.com/opensearch-project/data-prepper/runs/6182249051?check_suite_focus=true

build-resources.gradle
```diff
@@ -7,7 +7,7 @@
 ext.versionMap = [
         junitJupiter : '5.8.2',
         mockito : '3.11.2',
-        opentelemetryProto : '1.7.1-alpha',
+        opentelemetryProto : '0.16.0-alpha',
```
dlvenable, line 10: Did you have to make any code changes to support this? I tried making this change a few weeks ago. But, some of the tests were failing with an exception and I didn't get to look further into it.

data-prepper-plugins/otel-metrics-raw-processor/src/main/java/com/amazon/dataprepper/plugins/processor/otelmetrics/OtelMetricsRawProcessorConfig.java
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.amazon.dataprepper.plugins.processor.otelmetrics;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class OtelMetricsRawProcessorConfig {
+
+    @JsonProperty("calculateHistogramBuckets")
```
dlvenable, line 12: We are aiming to have Data Prepper use snake_case for all configurations to promote consistency. (It's not quite there yet). But, for new properties we want to use snake_case only. The plugin framework will automatically use snake_case. So removing these `@JsonProperty` annotations (or removing the value from them) will fix this.

data-prepper-plugins/otel-metrics-raw-processor/README.md
```diff
@@ -9,6 +9,98 @@ processor:
     - otel_metrics_raw_processor
 ```
 
+## Configuration
+It is possible to create explicit representations of histogram buckets and their boundaries with this plugin. This can be controlled with the following parameters:
+
+```yaml
+  processor:
+    - otel_metrics_raw_processor:
+        calculateHistogramBuckets: true
```
dlvenable, line 18: Per my other comment, please update the README with the snake_case properties.

dlvenable (changes requested):

data-prepper-plugins/otel-metrics-raw-processor/README.md
```diff
@@ -9,6 +9,98 @@ processor:
     - otel_metrics_raw_processor
 ```
 
+## Configuration
+It is possible to create explicit representations of histogram buckets and their boundaries. This feature can be controlled with the following parameters:
+
+```yaml
+  processor:
+    - otel_metrics_raw_processor:
+        calculate_histogram_buckets: true
```
dlvenable, line 18: I keep wondering about the default here. One of the goals of Data Prepper is to perform as much processing as possible during streaming so that OpenSearch does not need to do it. So I'd think that Data Prepper would typically calculate both bucket types. And if that is the case, then I think the default should be true.

What do you think?

data-prepper-plugins/otel-metrics-raw-processor/src/test/java/com/amazon/dataprepper/plugins/processor/otelmetrics/MetricsPluginGaugeTest.java
```diff
@@ -10,43 +10,60 @@
...
+import java.util.concurrent.TimeUnit;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.entry;
 
-
 @RunWith(MockitoJUnitRunner.class)
 public class MetricsPluginGaugeTest {
 
+    private static final Clock CLOCK = Clock.fixed(Instant.ofEpochSecond(1_700_000_000), ZoneOffset.UTC);
```
dlvenable, line 49: What is this 1.7 billion? Maybe extract it into a constant with a readable name. Or perhaps rename the `CLOCK` to include the meaning of what it provides.

data-prepper-plugins/otel-proto-common/src/main/java/com/amazon/dataprepper/plugins/otel/codec/OTelProtoCodec.java
```diff
@@ -457,4 +486,294 @@ protected AnyValue objectToAnyValue(final Object obj) throws UnsupportedEncoding
...
+    /**
+     * Maps a List of {@link io.opentelemetry.proto.metrics.v1.ExponentialHistogramDataPoint.Buckets} to an
+     * internal representation for Data Prepper.
+     * See <a href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#exponential-buckets">data model</a>
+     *
+     * @param buckets the list of buckets
+     * @param scale the scale of the exponential histogram
+     * @return a mapped list of Buckets
+     */
+    public static List<Bucket> createExponentialBuckets(ExponentialHistogramDataPoint.Buckets buckets, int scale) {
```
dlvenable, line 766: Should this method have any assertions on the `scale` to avoid issues with large values? With scale=30 the data point would have 1 billion buckets.

dlvenable (changes requested):
Overall this looks great! I left a few inline comments for discussion.

data-prepper-plugins/otel-metrics-raw-processor/README.md
kmssap: I was wandering about the same thing. And because the calculation of the bucket boundaries is not something that a user would typically expect because it adds something on top of what OpenTelemetry provides I initially thought it might be better to turn it off. But I can see the point of offloading the processing and this line of reasoning makes more sense in this context. I will turn it on by default.

data-prepper-plugins/otel-proto-common/src/main/java/com/amazon/dataprepper/plugins/otel/codec/OTelProtoCodec.java
```diff
@@ -66,6 +89,12 @@ public class OTelProtoCodec {
      */
     public static final Function<String, String> SPAN_ATTRIBUTES_REPLACE_DOT_WITH_AT = i -> SPAN_ATTRIBUTES + DOT + i.replace(DOT, AT);
     public static final Function<String, String> RESOURCE_ATTRIBUTES_REPLACE_DOT_WITH_AT = i -> RESOURCE_ATTRIBUTES + DOT + i.replace(DOT, AT);
+    public static final Function<String, String> PREFIX_AND_METRIC_ATTRIBUTES_REPLACE_DOT_WITH_AT = i -> METRIC_ATTRIBUTES + DOT + i.replace(DOT, AT);
+    public static final Function<String, String> PREFIX_AND_RESOURCE_ATTRIBUTES_REPLACE_DOT_WITH_AT = i -> RESOURCE_ATTRIBUTES + DOT + i.replace(DOT, AT);
+    public static final Function<String, String> PREFIX_AND_EXEMPLAR_ATTRIBUTES_REPLACE_DOT_WITH_AT = i -> EXEMPLAR_ATTRIBUTES + DOT + i.replace(DOT, AT);
+
+
+    private static final Map<Integer, double[]> EXPONENTIAL_BUCKET_BOUNDS = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
kmssap: Yes, it would be possible to introduce memory issues by chance or even malicious intent. Should we perhaps limit the scale to +/- 10 and introduce a config parameter that enables the user to adjust it ? Would 10 be a sensible choice ?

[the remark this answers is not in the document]
dlvenable: @kmssap ,

Yes, using a scale configuration makes sense. And I think 10 is a reasonable default. It will take up 32KB of memory per my calculations.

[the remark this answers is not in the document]
kmssap: Done

data-prepper-plugins/otel-proto-common/src/main/java/com/amazon/dataprepper/plugins/otel/codec/OTelProtoCodec.java
```diff
@@ -457,4 +486,294 @@ protected AnyValue objectToAnyValue(final Object obj) throws UnsupportedEncoding
...
+                                Hex.encodeHexString(exemplar.getSpanId().toByteArray()),
+                                Hex.encodeHexString(exemplar.getTraceId().toByteArray()),
+                                unpackExemplarValueList(exemplar.getFilteredAttributesList())))
+                .collect(Collectors.toList());
+    }
+
+
+    static double[] calculateBoundaries(int scale) {
+        int len = 1 << Math.abs(scale);
+        double[] boundaries = new double[len + 1];
```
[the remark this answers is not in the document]
kmssap: You are right, thanks for pointing this out. Currently we only have the base calculation. What needs to be added is the bucket location of measurements between the successive powers of 2...

data-prepper-plugins/otel-metrics-raw-processor/README.md
```diff
@@ -101,6 +104,12 @@ a scale parameter, offset and list of bucket counts:
     "positiveOffset : 1
 ```
 
+The `exponential_histogram_max_allowed_scale` parameter defines the maximum allowed scale for the exponential histogram. Increasing this parameter will increase potential
+memory consumption. See [the spec](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto) for more information on exponential histograms and their computational complexity.
+All exponential histograms that have a scale that is above 10 (or the configured parameter) will be discarded and logged with error level.
```
dlvenable, line 109: > that is above 10 (or the configured parameter)

I think this might be somewhat clearer.

> that is above the configured parameter (by default, 10).

data-prepper-plugins/otel-metrics-raw-processor/src/main/java/com/amazon/dataprepper/plugins/processor/otelmetrics/OTelMetricsRawProcessor.java
```diff
@@ -222,6 +227,14 @@ private List<? extends Record<? extends Metric>> mapHistogram(final io.opentelem
 
     private List<? extends Record<? extends Metric>> mapExponentialHistogram(io.opentelemetry.proto.metrics.v1.Metric metric, String serviceName, Map<String, Object> ils, Map<String, Object> resourceAttributes, String schemaUrl) {
         return metric.getExponentialHistogram().getDataPointsList().stream()
+                .filter(dp -> {
+                    if (otelMetricsRawProcessorConfig.getExponentialHistogramMaxAllowedScale() < Math.abs(dp.getScale())){
```
dlvenable, line 231: Should this allow any scale to pass through if the `calculateExponentialHistogramBuckets` is set to false? The scale is a problem only when performing the processing, right?

This might be more appropriate:

```
if (otelMetricsRawProcessorConfig.getCalculateExponentialHistogramBuckets() &&
                            otelMetricsRawProcessorConfig.getExponentialHistogramMaxAllowedScale() < Math.abs(dp.getScale())){
```

What do you think?

dlvenable (changes requested):
Thanks for adding that new configuration! I think this is almost complete. I left a couple small comments. Also, checkstyle is failing. I think you just need to remove unnecessary imports (optimize imports in IntelliJ).

> Error: eckstyle] [ERROR] /home/runner/work/data-prepper/data-prepper/data-prepper-plugins/otel-metrics-raw-processor/src/main/java/com/amazon/dataprepper/plugins/processor/otelmetrics/OTelMetricsRawProcessor.java:10:8: Unused import - com.amazon.dataprepper.model.configuration.PipelineModel. [UnusedImports]

data-prepper-plugins/service-map-stateful/build.gradle
```diff
@@ -19,7 +19,7 @@ dependencies {
     implementation 'commons-codec:commons-codec:1.15'
     implementation 'io.micrometer:micrometer-core'
     implementation 'com.fasterxml.jackson.core:jackson-databind'
-    implementation("io.opentelemetry:opentelemetry-proto:${versionMap.opentelemetryProto}") {
+    implementation("io.opentelemetry.io:opentelemetry-proto:${versionMap.opentelemetryProto}") {
```
dlvenable, line 22: @kmssap , I think this should be `io.opentelemetry.proto`. That may be why the build is failing.

kaisternad: Thanks, this has been fixed in the latest commit.

kmssap: Thanks, this has been fixed with the latest commit.

dlvenable (approved):

dlvenable:
Thank you @kmssap for this great PR and your work in updating it for our recent changes!

Issue #1614, opened 2022-07-25
Implement CSVProcessor parsing


asifsmohammed:
Implement CSV processor logic in `doExecute()`

PR #1627, opened 2022-07-28
Implemented CSV Processor w unit tests, added validation that delimit…

finnroblin:
…er and quote char are different

### Description
Implementation of the CSV Processor & Unit Tests. Also added a minor validation condition to the Config that the delimiter and quote character must be different (otherwise parsing behavior may be strange). 
 
### Issues Resolved
[Implemented CSVProcessor parsing #1614 ](https://github.com/opensearch-project/data-prepper/issues/1614)
 
### Check List


data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -12,22 +12,88 @@
...
+import java.util.List;
 
 /**
  * Processor to parse CSV data in Events.
  *
  */
 @DataPrepperPlugin(name="csv", pluginType = Processor.class, pluginConfigurationType = CSVProcessorConfig.class)
 public class CSVProcessor extends AbstractProcessor<Record<Event>, Record<Event>> {
+    private final List<String> header;
+    private final String quote_char;
```
travisbenedict, line 32: By convention Java uses camelcase for variable names. More info on style conventions here: https://www.javatpoint.com/java-naming-conventions#:~:text=CamelCase%20in%20Java%20naming%20conventions,%2C%20ActionEvent%2C%20ActionListener%2C%20etc.

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -12,22 +12,88 @@
...
+        this.headerSourceKey = config.getColumnNamesSourceKey();
+        this.isDeleteHeaderAfterProcessing = config.isDeleteHeader();
+
+        this.hasUserSpecifiedHeader = (this.header != null);
+        this.hasHeaderSource = (this.headerSourceKey != null);
     }
 
     @Override
     public Collection<Record<Event>> doExecute(final Collection<Record<Event>> records) {
+        CsvMapper mapper = createCsvMapper();
```
travisbenedict, line 56: variables should be marked as `final` to prevent reassignment

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -12,22 +12,88 @@
...
+
+        for (Record<Event> record : records) {
+
+            Event event = record.getData();
+
+            String message = event.get(this.source_key, String.class);
+            boolean this_event_has_header_source = this.hasHeaderSource && event.containsKey(this.headerSourceKey);
+
+            try {
+                MappingIterator<List<String>> message_it = mapper.readerFor(List.class).with(schema).readValues(message);
```
travisbenedict, line 67: `messageIterator` would be a better name here since the abbreviation "it" could lead to some confusion

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -45,4 +111,44 @@ public boolean isReadyForShutdown() {
...
     }
+
+    private CsvMapper createCsvMapper() {
+        CsvMapper mapper = new CsvMapper();
+        mapper.enable(CsvParser.Feature.WRAP_AS_ARRAY); // allows mapper to read with empty schema
+        return mapper;
+    }
+
+    private CsvSchema createCsvSchema() {
+        char delimiter_as_char = this.delimiter.charAt(0); // safe due to config input validations
```
travisbenedict, line 122: Could delimiter and quote be parsed as `char`s instead of Strings?

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessorConfig.java
```diff
@@ -66,4 +66,10 @@ boolean isValidDelimiter() {
     boolean isValidQuoteCharacter() {
         return quoteCharacter.length() == 1;
     }
+
+    @AssertTrue(message = "quote_character and delimiter cannot be the same character")
+    boolean areDelimiterAndQuoteCharacterDifferent() {
+        return !(delimiter.equals(quoteCharacter));
```
travisbenedict, line 72: Good catch!

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -12,22 +12,88 @@
...
+                        List<String> header_from_event_source = header_it.nextValue();
+                        putDataInEvent(event, header_from_event_source, row);
+                    } else {
+                        if (this.hasUserSpecifiedHeader) {
+                            putDataInEvent(event, this.header, row);
+                        }
+                        else {
+                            List<String> emptyHeader = new ArrayList<String>();
+                            putDataInEvent(event, emptyHeader, row);
+                        }
```
travisbenedict, line 86: I think you could simplify this section and make it more readable by moving the header processing into a helper method. You could also use `else if` to get rid of the nested if statements. I'm thinking something like

```
private List<String> parseHeader(....) {
    if (this_event_has_header_source) {
        ....
        return header_it.nextValue();
    } else if (this.hasUserSpecifiedHeader) {
        return this.header;
    } else {
       return new ArrayList<String>();
    }
}
```


This section would then become 
```
final List<String> header = parseHeader(....);
putDataInEvent(event, header, row);
```

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -45,4 +111,44 @@ public boolean isReadyForShutdown() {
...
+
+    private CsvSchema createCsvSchema() {
+        char delimiter_as_char = this.delimiter.charAt(0); // safe due to config input validations
+        char quote_char_as_char = this.quote_char.charAt(0); // safe due to config input validations
+        CsvSchema schema = CsvSchema.emptySchema().withColumnSeparator(delimiter_as_char).withQuoteChar(quote_char_as_char);
+        return schema;
+    }
+
+    private void putDataInEvent(Event event, List<String> header, List<String> data) {
+        // open question: We decided that if there is a user-specified header and a header source event key then we use the header source
```
travisbenedict, line 129: We shouldn't commit a comment like this. This would be a good thing to move to the Github issue

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
finnroblin: That's much cleaner, thanks!

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
finnroblin: Agreed. I'll change `header_it` to `headerIterator` as well.

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
finnroblin: Good callout! I had these as `char`s in the config originally. Asif and I talked about the config offline and decided to not use any primitive types. I'm not exactly sure but I believe using primitives in the config can cause weird behavior around null values. I think it was a bigger problem for `boolean` vs `java.lang.Boolean`. Do you forsee any problems with the String conversion implementation?

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
finnroblin: Good catch! I removed the code comment and clarified in the GitHub issue (#1081).

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
finnroblin: Thanks for the resource!

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
travisbenedict: The point about null conversion makes sense. I think the String conversion approach should be fine as is.

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -12,22 +12,60 @@
...
     public Collection<Record<Event>> doExecute(final Collection<Record<Event>> records) {
+        final CsvMapper mapper = createCsvMapper();
+        final CsvSchema schema = createCsvSchema();
+
+        for (final Record<Event> record : records) {
+
+            final Event event = record.getData();
+
+            final String message = event.get(config.getSource(), String.class);
+            final boolean userDidSpecifyHeaderEventKey = (config.getColumnNamesSourceKey() != null);
```
travisbenedict, line 49: I think you could use `Objects.nonnull()` instead of `!= null` to be more declarative and help with readability

https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#nonNull-java.lang.Object-

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -12,22 +12,60 @@
...
+            final boolean thisEventHasHeaderSource = userDidSpecifyHeaderEventKey && event.containsKey(config.getColumnNamesSourceKey());
+
+            try {
+                final MappingIterator<List<String>> messageIterator = mapper.readerFor(List.class).with(schema).readValues(message);
+
+                // otherwise the message is empty
+                if (messageIterator.hasNextValue()) {
```
travisbenedict, line 56: Would it make sense for this to be `while` instead of `if` are we always guaranteed that there would be just one row? How would this work for parsing a whole CSV file at one?

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -12,22 +12,60 @@
...
+                    final List<String> row = messageIterator.nextValue();
+                    final List<String> header = parseHeader(event, thisEventHasHeaderSource, mapper, schema);
+                    putDataInEvent(event, header, row);
+                }
+
+                if (thisEventHasHeaderSource && config.isDeleteHeader()) {
+                    event.delete(config.getColumnNamesSourceKey());
+                }
+            } catch (IOException e) {
+                System.out.println(e.toString());
```
travisbenedict, line 66: I think this should be an `error` log instead of printing to stdout. Here are some references on that

Instantiate logger: https://github.com/opensearch-project/data-prepper/blob/0a5910b789e004a0b69b8c053660db8b3a2da835/data-prepper-plugins/grok-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/grok/GrokPrepper.java#L58

Write error log: https://github.com/opensearch-project/data-prepper/blob/0a5910b789e004a0b69b8c053660db8b3a2da835/data-prepper-plugins/grok-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/grok/GrokPrepper.java#L124

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -45,4 +83,65 @@ public boolean isReadyForShutdown() {
...
+                final MappingIterator<List<String>> headerIterator = mapper.readerFor(List.class).with(schema)
+                        .readValues(headerUnprocessed);
+                // if header is empty, behaves correctly since columns are autogenerated.
+                final List<String> headerFromEventSource = headerIterator.nextValue();
+                return headerFromEventSource;
+            }
+            catch (IOException e) {
+                final List<String> emptyHeader = new ArrayList<String>();
+                return emptyHeader;
+            }
```
travisbenedict, line 114: Not important but you might consider moving this into its own function as well. I usually try to avoid having a `try/catch` block within an `if` if there are multiple branches already.

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -45,4 +83,65 @@ public boolean isReadyForShutdown() {
...
+    }
+
+    private CsvSchema createCsvSchema() {
+        final char delimiterAsChar = config.getDelimiter().charAt(0); // safe due to config input validations
+        final char quoteCharAsChar = config.getQuoteCharacter().charAt(0); // safe due to config input validations
+        final CsvSchema schema = CsvSchema.emptySchema().withColumnSeparator(delimiterAsChar).withQuoteChar(quoteCharAsChar);
+        return schema;
+    }
+
+    private List<String> parseHeader(Event event, final boolean thisEventHasHeaderSource, final CsvMapper mapper,
```
travisbenedict, line 100: nit: missed a couple `final`s for method arguments. `event` here and in `putDataInEvent` and `colNumber` for `generateColumn`. Exceptions within a `catch` clause can be `final` too - `catch (final IOException e) {`

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -45,4 +83,65 @@ public boolean isReadyForShutdown() {
...
+            for (; i < headerSize; i++) {
+                event.put(header.get(i),data.get(i));
+            }
+            for (int j = i; j < dataSize; j++) {
+                event.put(generateColumn(j),data.get(j));
+            }
+        }
+    }
+
+    private String generateColumn(int colNumber) {
```
travisbenedict, line 143: `generateColumnHeader` might be a more accurate name

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
```diff
@@ -45,4 +83,65 @@ public boolean isReadyForShutdown() {
...
+                event.put(header.get(i),data.get(i));
+            }
+        } else {
+            int i = 0;
+            for (; i < headerSize; i++) {
+                event.put(header.get(i),data.get(i));
+            }
+            for (int j = i; j < dataSize; j++) {
+                event.put(generateColumn(j),data.get(j));
+            }
```
travisbenedict, line 139: I think this could work for `dataSize <= headerSize` too right? If `headerSize >= dataSize` then the second loop just wouldn't be executed. Cutting out the `if` would help simplify this function. 

You may also want to consider creating more informative variable names than `i` and `j`. It took me a few reads to understand what they were being used for. Maybe rewriting this function to something like this would make sense:

```
int providedHeadersColIdx = 0;
for (providedHeadersColIdx; providedHeadersColIdx < headers.size(); providedHeadersColIdx++) {
    event.put(headers.get(providedHeadersColIdx), data.get(providedHeadersColIdx));
}
for (int remainingColsIdx = providedHeadersColIdx; remainingColsIdx < data.size(); remainingColsIdx++) {
    event.put(generateColumn(remainingColsIdx),data.get(remainingColsIdx));
}
```
There might be some clever naming that you could do here. The names I have here are pretty verbose

travisbenedict (commented):
Most of my comments are non blocking, stylistic things. The main things that I'm concerned about are the `while` vs `if` and logging comments

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
finnroblin: This is a good callout and raises more questions about the behavior when there is a multiline newline-delimited CSV file in the `message` field. Right now we assume that splitting the CSV file into a collection of single line records is handled upstream by the codec, so the processor will receive Events with a single row record. With the current implementation, only the first line will be parsed. If we change to `while` instead of `if`, the Event will be overwritten by every line so it will eventually only contain the last line. So neither solution works great for multiline CSV files.

I can think of two ways to handle multiline CSV files off the top of my head:

1. Split each data column into a separate field:
`{"message":"1,2\n3,4"}` => `{"col1":"1,3","col2":"2,4"}`
2. Emit multiple outgoing events. Parse each line of the original `message` field into a separate Event: 
`{"message":"1,2\n3,4"}` => `[{"col1":"1","col2":"2"}, {"col1":"3","col2":"4"}]`
I think that 2 is a better solution because each CSV row record is kept together, which makes the data more interpretable since users can compare the data in different fields on each event.

If we want to support header autodetection in the CSV file in the Event field, we will need to add another configuration option so that users can indicate that the incoming Event will contain a multiline CSV file with a header (and possibly another configuration option for the newline delimiter, which would be `\n` be default). 

What are your thoughts on handling multiline CSV files and adding configuration options for entire CSV file header autodetection?

finnroblin: Travis and I discussed processing multiline CSV files offline and decided to table multiline processing for now because the main ways of ingesting CSV files (through the `file` source or `S3 Source`) already split the data into rows. The current processor and config can be extended to handle multiline CSV parsing for a single event in a follow up PR if there's a need for it.

data-prepper-plugins/csv-processor/src/main/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessor.java
finnroblin: Great suggestion! We can do something like: 
```
int providedHeaderColIdx = 0;
for (; providedHeaderColIdx < header.size() && providedHeaderColIdx < data.size(); providedHeaderColIdx++) {
    event.put(header.get(providedHeaderColIdx), data.get(providedHeaderColIdx));
}
```
To prevent the index being out of bounds if headerSize > dataSize. Then we can cut the `if` and keep it to two `for` loops.

travisbenedict (dismissed):

data-prepper-plugins/csv-processor/src/test/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessorTest.java
```diff
@@ -6,34 +6,316 @@
...
+        final List<Record<Event>> editedEvents = (List<Record<Event>>) csvProcessor.doExecute(Collections.singletonList(eventUnderTest));
+        final Event parsedEvent = getSingleEvent(editedEvents);
+        assertThat(parsedEvent.containsKey("message"), equalTo(true));
+        assertThat(parsedEvent.containsKey("column1"), equalTo(false));
+    }
+
+    @Test
+    public void test_when_delimiterIsTab_then_parsedCorrectly() {
+        when(processorConfig.getDelimiter()).thenReturn("\t");
+        csvProcessor = createObjectUnderTest();
```
asifsmohammed, line 68: You don't have to create this object in every test since it's been created in `setup()` method which is annotated with `@BeforeEach`.

data-prepper-plugins/csv-processor/src/test/java/com/amazon/dataprepper/plugins/processor/csv/CSVProcessorTest.java
```diff
@@ -6,34 +6,316 @@
...
+        assertThatKeyEquals(parsedEvent, "column3", "3");
+    }
+
+    @Test
+    public void test_when_deleteHeaderAndHeaderSourceDefined_then_headerIsDeleted() {
+        when(processorConfig.isDeleteHeader()).thenReturn(true);
+        when(processorConfig.getColumnNamesSourceKey()).thenReturn("header");
+        csvProcessor = createObjectUnderTest();
+
+        final Map<String, Object> eventData = new HashMap();
```
asifsmohammed, line 84: nit: `new HashMap<>();`
Please make this change in all the tests.

asifsmohammed (changes requested):

asifsmohammed (approved):

travisbenedict (approved):

Issue #1615, opened 2022-07-25
Modify newline codec to add CSV header auto detection support


asifsmohammed:
Update existing newline codec in s3 source

PR #1640, opened 2022-08-02
Enhanced Newline Codec w optional header_destination to add first lin…

finnroblin:
…e as header to outgoing events

### Description
Adds to the S3 Source `newline` codec. Adds a `header_destination` field to NewlineDelimitedConfig. If `header_destination` is specified, all outgoing Events will have a header field attached to them. The header is the first line in the `inputStream` after `skipLines` have been skipped. If a user does not specify `header_destination` then the `newline` functionality and interface are the exact same.
 
### Issues Resolved
Resolves [#1615](https://github.com/opensearch-project/data-prepper/issues/1615)
 
### Check List

### DIFF
--- a/data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedCodec.java
+++ b/data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedCodec.java
@@ -16,6 +16,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 import java.util.function.Consumer;
 
@@ -23,6 +25,7 @@ import java.util.function.Consumer;
 public class NewlineDelimitedCodec implements Codec {
     private static final String MESSAGE_FIELD_NAME = "message";
     private final int skipLines;
+    private final String headerDestination;
 
     @DataPrepperPluginConstructor
     public NewlineDelimitedCodec(final NewlineDelimitedConfig config) {
@@ -32,6 +35,8 @@ public class NewlineDelimitedCodec implements Codec {
         if (skipLines < 0) {
             throw new IllegalArgumentException("skipLines must be non-negative.");
         }
+
+        headerDestination = config.getHeaderDestination();
     }
 
     @Override
@@ -42,18 +47,34 @@ public class NewlineDelimitedCodec implements Codec {
     }
 
     private void parseBufferedReader(final BufferedReader reader, final Consumer<Record<Event>> eventConsumer) throws IOException {
+        final boolean doAddHeaderToOutgoingEvents = Objects.nonNull(this.headerDestination);
+        boolean hasReadHeader = false;
+        String header = "";
+
         int linesToSkip = skipLines;
         String line;
         while ((line = reader.readLine()) != null) {
 
-            if (linesToSkip > 0) {
-                linesToSkip--;
+            if (linesToSkip > 0 || (doAddHeaderToOutgoingEvents && !hasReadHeader)) {
+                if (linesToSkip > 0) {
+                    linesToSkip--;
+                } else {
+                    header = line;
+                    hasReadHeader = true;
+                }
                 continue;
             }
 
-            final Event event = JacksonLog.builder()
-                    .withData(Collections.singletonMap(MESSAGE_FIELD_NAME, line))
-                    .build();
+            final Map<String, String> eventData;
+            if (doAddHeaderToOutgoingEvents) {
+                eventData = new HashMap<>();
+                eventData.put(this.headerDestination, header);
+                eventData.put(MESSAGE_FIELD_NAME, line);
+            } else {
+                eventData = Collections.singletonMap(MESSAGE_FIELD_NAME, line);
+            }
+
+            final Event event = JacksonLog.builder().withData(eventData).build();
             eventConsumer.accept(new Record<>(event));
         }
     }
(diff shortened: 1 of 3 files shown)


data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedCodec.java
```diff
@@ -42,18 +47,34 @@ public void parse(final InputStream inputStream, final Consumer<Record<Event>> e
...
+        boolean hasReadHeader = false;
+        String header = "";
+
         int linesToSkip = skipLines;
         String line;
         while ((line = reader.readLine()) != null) {
 
-            if (linesToSkip > 0) {
-                linesToSkip--;
+            if (linesToSkip > 0 || (doAddHeaderToOutgoingEvents && !hasReadHeader)) {
```
travisbenedict, line 58: I think if you split this clause into two separate `if` statements you could simplify this and remove some branching. As it is now it seems like there would be a bug if you wanted to skip multiple lines since `hasReadHeader` would be set to `true` on the first iteration of the loop

I'm thinking something like:
```
if (linesToSkip > 0) {
    linesToSkip--;
    continue;
}

final Map<String, String> eventData = new HashMap<>();

if (doAddHeaderToOutgoingEvents && !hasReadHeader) {
    header = line;
    hasReadHeader = true;
    eventData.put(this.headerDestination, header);
}

eventData.put(MESSAGE_FIELD_NAME, line);
final Event event = JacksonLog.builder().withData(eventData).build();

**Review by travisbenedict — CHANGES_REQUESTED** [review]

File: data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedCodec.java (hunk 1 above)
Comment (travisbenedict) [reply data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedCodec.java:58]: Ignore the above suggestion I wasn't properly accounting for the `continue`. I think what's here is fine.

You might consider using some variables to break down the boolean logic and make things more readable. Something like:
```
final boolean stillNeedToReadHeader = doAddHeaderToOutgoingEvents && !hasReadHeader;
final boolean shouldSkipLine = linesToSkip > 0 || stillNeedToReadHeader;
if (shouldSkipLine) {
    ...
}
```


I think the end of the method could also be rewritten to make the code shorter and more declarative by getting rid of the `else` but that's just a personal preference. 
```
final Map<String, String> eventData = new HashMap<>(); 
if (doAddHeaderToOutgoingEvents) { 
    eventData.put(this.headerDestination, header); 
} 

eventData.put(MESSAGE_FIELD_NAME, line); 
final Event event = JacksonLog.builder().withData(eventData).build();
```

**Review by travisbenedict — APPROVED** [review]

File: data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedCodec.java (hunk 1 above)
Comment (finnroblin) [reply data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedCodec.java:58]: Thanks for your feedback! I agree that the boolean logic needs some work. I took your suggestion and added some variables to make the first statement more readable. I also agree that getting rid of the `else` is more declarative and I've made that change. I'll push these changes to this branch along with the next review to avoid triggering a re-review.

File: data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedCodec.java [hunk 2]
```diff
@@ -42,18 +47,34 @@ public void parse(final InputStream inputStream, final Consumer<Record<Event>> e
     }
 
     private void parseBufferedReader(final BufferedReader reader, final Consumer<Record<Event>> eventConsumer) throws IOException {
+        final boolean doAddHeaderToOutgoingEvents = Objects.nonNull(this.headerDestination);
```
asifsmohammed, line 50: What should be the expected behavior if  headerDestination is empty String `""`?
nit: you can remove this keyword in line 50 & 73 to access `headerDestination`.

data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedConfig.java
```diff
@@ -6,12 +6,15 @@
...
+import jakarta.validation.constraints.AssertTrue;
+
+import java.util.Objects;
 
 /**
  * Configuration class for the newline delimited codec.
  */
 public class NewlineDelimitedConfig {
-    private final int skipLines = 0;
+    private int skipLines = 0;
```
finnroblin, line 17: I removed the `final` keyword here because I wasn't able to specify > 0 in a YAML file and have the lines skipped properly. I think the unit tests only pass for different skipLines because they mock `getSkipLines`, not because they actually set the skipLines variable (which can't be changed from `0` because of `final`). Removing `final` caused expected behavior. I'm happy to move this issue somewhere else, but I also think it's a minor change and within the scope of the `newline` codec enhancements of this review.

asifsmohammed (approved):
LGTM

data-prepper-plugins/s3-source/src/main/java/com/amazon/dataprepper/plugins/source/codec/NewlineDelimitedConfig.java
travisbenedict: Nice catch

travisbenedict (approved):

PR #1688, opened 2022-08-23
ParseJsonProcessor initial implementation

finnroblin:
### Description
Starting point to implement the `parse_json` processor. This PR adds a configuration class, unit tests, and implementation of core features of the `parse_json` processor. With these changes the processor can deserialize JSON strings and will deserialize all nested fields. A JSON pointer feature to parse only a part of the JSON still needs to be added. The behavior of deserialized JSON Arrays still needs to be tested and some modifications to the logic might be necessary.
 
### Issues Resolved
Step towards resolving #831 
 
### Check List


data-prepper-plugins/parse-json-processor/src/main/java/com/amazon/dataprepper/plugins/processor/parsejson/ParseJsonProcessorConfig.java
```diff
@@ -0,0 +1,41 @@
...
+        return source;
+    }
+
+    /**
+     * The destination that the parsed JSON is written to. Defaults to the root of the Event.
+     * If the destination field already exists, it will be overwritten.
+     *
+     * @return The name of the destination field.
+     */
+    public String getDestination() {
```
finnroblin, line 38: I think it would be better for the default `destination` (Event root) to be `/` instead of `null`. Then we avoid some null values and my opinion is that `/` is more declarative. What do others think?

travisbenedict: If we used `/` as the root what would the expected behavior behavior be if someone had multiple `/`s in the destination? One might expect it to create a nested field similar to directory syntax . 

ex: Would `destination: /foo/bar` become `foo.bar: <parsed_json>` where `bar` is a nested field?

That might be useful, but I don't think that's what we're going for here so maybe `/` would be misleading. I do think your point about a more declarative default is valid though

travisbenedict (approved):

data-prepper-plugins/parse-json-processor/src/main/java/com/amazon/dataprepper/plugins/processor/parsejson/ParseJsonProcessorConfig.java
finnroblin: Right now the behavior if someone had multiple `/` in the destination (`/foo/bar`) is that the parsed json will be written to the most nested field in the destination. So if the json is `{"key":"value"}` and the destination is (`/foo/bar`), `event.get("/foo/bar", Map.class) == {"key":"value"}` and `event.get("/foo/bar/key", String.class) == "value"`. Right now the `parse_json` processor doesn't flatten nested fields (so `/foo/bar` is not flattened to `foo.bar`), it just puts any nested json into the event without changing it. 

I do see how `/` is misleading. It's also misleading for another reason: calling `event.put("/", value)` will not put `value` into the event root because the `JacksonEvent` implementation trims leading slashes. So it would not write `value` to the event at all. 

I think we can stick with a `null` default and be more declarative by mentioning that `destination is null -> write to root` in documentation.

data-prepper-plugins/parse-json-processor/src/main/java/com/amazon/dataprepper/plugins/processor/parsejson/ParseJsonProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public Collection<Record<Event>> doExecute(final Collection<Record<Event>> records) {
+        final ObjectMapper objectMapper = new ObjectMapper();
+        final boolean doWriteToRoot = Objects.isNull(destination);
+
+        for (final Record<Event> record : records) {
+            final Event event = record.getData();
+            final String message = event.get(source, String.class);
+            try {
+                TypeReference<HashMap<String, Object>> hashMapTypeReference = new TypeReference<HashMap<String, Object>>() {};
```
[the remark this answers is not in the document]
finnroblin: Thanks for catching! I'll add `final` in the next revision or the next `parse_json` PR to avoid triggering re-review.

data-prepper-plugins/parse-json-processor/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+dependencies {
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.1'
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.8.1'
+
+    implementation project(':data-prepper-api')
+    implementation project(':data-prepper-plugins:common')
+    implementation 'com.fasterxml.jackson.core:jackson-databind'
+}
+
+test {
```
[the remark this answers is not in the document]
finnroblin: Thanks for the recommendation! 

I set the threshold to 96% but it's slightly higher in reality because there aren't any unit tests for the `prepareForShutdown`, `isReadyForShutdown`, and `shutdown` methods. These methods are needed to extend `AbstractProcessor` but they're empty because `parse_json` doesn't have any special behavior for shutdown. The implemented line and branch coverage are 100%.

data-prepper-plugins/parse-json-processor/src/main/java/com/amazon/dataprepper/plugins/processor/parsejson/ParseJsonProcessorConfig.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import jakarta.validation.constraints.NotEmpty;
+import jakarta.validation.constraints.NotNull;
+
+public class ParseJsonProcessorConfig {
+    static final String DEFAULT_SOURCE = "message";
+
+    @NotEmpty
+    @NotNull
```
asifsmohammed, line 16: nit: @NotEmpty also checks for not null. I think @NotNull is required in this case.

data-prepper-plugins/parse-json-processor/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+plugins {
+    id 'java'
+}
+
+group 'org.opensearch.dataprepper'
+version '2.0.0-SNAPSHOT'
```
asifsmohammed, line 11: Please remove these 2 lines. it's not required

data-prepper-plugins/parse-json-processor/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.1'
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.8.1'
```
asifsmohammed, line 19: You can remove these dependencies. I think these are used from parent project.

travisbenedict (approved):

sbayer55 (approved):
Looks good, thanks Finn!

PR #1707, opened 2022-08-31
Remove deprecated prepper plugin type

oeyh:
### Description
This change removes the support for the deprecated `Prepper` plugin type and the support of the `prepper` name in configuration yaml; updates samples and documentations accordingly. 
 
### Issues Resolved
Contributes to #619  
 
### Check List


dlvenable (dismissed):
This looks good. Thanks!

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
```diff
@@ -150,7 +149,7 @@ private List<Processor> newProcessor(final PluginSetting pluginSetting) {
             LOG.warn(
                     "No plugin of type Processor found for plugin setting: {}, attempting to find comparable Prepper plugin.",
                     pluginSetting.getName());
-            return newProcessor(pluginSetting, Prepper.class)
+            return newProcessor(pluginSetting, Processor.class)
                     .stream()
                     .map(prepper -> (Processor) prepper)
                     .collect(Collectors.toList());
```
cmanning09, line 155: I believe this whole code block can be changed to the following now:

```
return newProcessor(pluginSetting, Processor.class);
```

cmanning09 (commented):
Overall looks good and eliminates the the prepper interface and api support. I did find a few references to prepper now in the code base but those are just variables. or comments (See [ServiceMapStatefulPrepper](https://github.com/opensearch-project/data-prepper/search?q=ServiceMapStatefulPrepper#L68)) but those don't need to be addressed now.

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
cmanning09: Actually, this whole catch logic can be removed. We should no longer be searching for preppers.

oeyh: Good catch! I further simplified the two `newProcessor()` methods into one on top of your comment since the second parameter `baseClass` will always be `Processor.class` now that Prepper is removed. See changes in [3b466c0](https://github.com/opensearch-project/data-prepper/pull/1707/commits/3b466c0c23a1b5c7954ab4b0b6d7a46533b8c827).

oeyh:
@dlv @cmanning09 I have added the changes to rename `PrepperState` to this PR as well (in [bff9ded](https://github.com/opensearch-project/data-prepper/pull/1707/commits/bff9ded0e8332a385ffb20f764df5fb626dd3cc4)). 

For the other changes we discussed, i.e. renaming class names in `data-prepper-plugins` like GrokPrepper, ServiceMapStatefulPrepper, also including MapDbPrepperState, etc., I'd like to add in a separate PR to keep this one from growing too large.

dlvenable (dismissed):

cmanning09 (approved):

dlvenable:
Something about this rebase doesn't seem right. The diff shows that changes that should have already been merged into `main`.

dlvenable (dismissed):

oeyh:
@dlvenable @cmanning09 This is ready for another look. The previous approvals were dismissed due to new changes to address merge conflicts. Sorry that those new changes are not obvious and buried in the commit history, but they are all pretty trial changes.

dlvenable (approved):

dlvenable (changes requested):

e2e-test/trace/src/integrationTest/resources/raw-span-e2e-pipeline-latest-release.yml
```diff
@@ -14,12 +14,12 @@ raw-pipeline:
   source:
     pipeline:
       name: "entry-pipeline"
-  prepper:
+  processor:
     - otel_trace_raw_prepper:
```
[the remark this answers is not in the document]
oeyh: I did changed it to `otel_trace_raw` once and noticed that it caused the end-to-end test to fail. I think this is testing on the released docker image and thus still taking the old configuration name.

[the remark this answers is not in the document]
dlvenable: If this is causing merge issues, we can fix it in a follow-on PR as well.

dlvenable (approved):

cmanning09 (approved):

PR #1719, opened 2022-09-01
Acm private key password is nullable

asifsmohammed:

### Description
`ACMCertificatProvider` generates a passphrase if not provided.
https://github.com/opensearch-project/data-prepper/blob/caebb1964233c0d400ef40c8c7c24353ee3091ff/data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/certificate/acm/ACMCertificateProvider.java#L82
 
### Check List


dlvenable (approved):
This is not a valid configuration, so we're removing it, right?

cmanning09 (approved):

asifsmohammed:
> This is not a valid configuration, so we're removing it, right?

Yes. private key password can be null but previously we had a check that it can't be null.

PR #1750, opened 2022-09-12
Added Local peer forwarder and updated http servoce class to write to…

asifsmohammed:
… buffer

### Description
- Updated WireEvents to have pipeline name
- Updated HTTP service to write deserialized events to buffer
- Added a getter for buffer map in `PeerForwarderProvider` 
- Added `LocalPeerForwarder` which is used by plugins which don't implement `RequiresPeerForwarding`
- Added `REQUEST_ENTITY_TOO_LARGE` response to response handler if request size is greater than buffer size
 
### Check List

### DIFF
--- a/data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
+++ b/data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
@@ -106,9 +106,6 @@ public class PipelineParser {
 
             LOG.info("Building processors for the pipeline [{}]", pipelineName);
             final int processorThreads = pipelineConfiguration.getWorkers();
-            final List<List<Processor>> processorSets = pipelineConfiguration.getProcessorPluginSettings().stream()
-                    .map(this::newProcessor)
-                    .collect(Collectors.toList());
 
             final Map<List<Processor>, String> processorMap = pipelineConfiguration.getProcessorPluginSettings().stream()
                     .collect(Collectors.toMap(this::newProcessor, PluginSetting::getName));
@@ -117,8 +114,6 @@ public class PipelineParser {
                     .map(processorMapEntry -> processorMapEntry.getKey().stream()
                             .map(processor -> {
                                 if (processor instanceof RequiresPeerForwarding) {
-                                    // TODO: Create buffer per stateful processor and store map of processor, buffer
-                                    // TODO: get plugin id from PipelineParser
                                     return new PeerForwardingProcessorDecorator(
                                             processor, peerForwarderProvider, pipelineName, processorMapEntry.getValue()
                                     );
--- a/data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
+++ b/data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
@@ -107,21 +107,23 @@ public class PipelineParser {
             LOG.info("Building processors for the pipeline [{}]", pipelineName);
             final int processorThreads = pipelineConfiguration.getWorkers();
 
-            final Map<List<Processor>, String> processorMap = pipelineConfiguration.getProcessorPluginSettings().stream()
-                    .collect(Collectors.toMap(this::newProcessor, PluginSetting::getName));
+            final List<List<IdentifiedComponent<Processor>>> processorSets = pipelineConfiguration.getProcessorPluginSettings().stream()
+                    .map(this::newProcessor)
+                    .collect(Collectors.toList());
 
-            final List<List<Processor>> decoratedProcessorSets = processorMap.entrySet().stream()
-                    .map(processorMapEntry -> processorMapEntry.getKey().stream()
-                            .map(processor -> {
-                                if (processor instanceof RequiresPeerForwarding) {
+            final List<List<Processor>> decoratedProcessorSets = processorSets.stream()
+                    .map(processorComponentList -> processorComponentList.stream()
+                            .map(processorComponent -> {
+                                if (processorComponent.getComponent() instanceof RequiresPeerForwarding) {
                                     return new PeerForwardingProcessorDecorator(
-                                            processor, peerForwarderProvider, pipelineName, processorMapEntry.getValue()
+                                            processorComponent.getComponent(), peerForwarderProvider, pipelineName, processorComponent.getName()
                                     );
                                 }
-                                return processor;
+                                return processorComponent.getComponent();
                             })
                             .collect(Collectors.toList())
-                    ).collect(Collectors.toList());
+                        ).collect(Collectors.toList());
+
 
             final int readBatchDelay = pipelineConfiguration.getReadBatchDelay();
 
@@ -141,13 +143,17 @@ public class PipelineParser {
 
     }
 
-    private List<Processor> newProcessor(final PluginSetting pluginSetting) {
-        return pluginFactory.loadPlugins(
+    private List<IdentifiedComponent<Processor>> newProcessor(final PluginSetting pluginSetting) {
+        final List<Processor> processors = pluginFactory.loadPlugins(
                 Processor.class,
                 pluginSetting,
                 actualClass -> actualClass.isAnnotationPresent(SingleThread.class) ?
                         pluginSetting.getNumberOfProcessWorkers() :
                         1);
+
+        return processors.stream()
+                .map(processor -> new IdentifiedComponent<>(processor, pluginSetting.getName()))
+                .collect(Collectors.toList());
     }
 
     private Optional<Source> getSourceIfPipelineType(
@@ -233,6 +239,24 @@ public class PipelineParser {
             pipelineMap.remove(pipelineName);
             sourceConnectorMap.remove(pipelineName);
             removeConnectedPipelines(pipelineName, pipelineConfigurationMap, pipelineMap);
+        }
+    }
+
+    private static class IdentifiedComponent<T> {
+        private final T component;
+        private final String name;
+
+        private IdentifiedComponent(final T component, final String name) {
+            this.component = component;
+            this.name = name;
+        }
+
+        T getComponent() {
+            return component;
+        }
+
+        String getName() {
+            return name;
         }
     }
 }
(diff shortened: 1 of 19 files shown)


data-prepper-core/src/test/java/org/opensearch/dataprepper/peerforwarder/server/PeerForwarderServerProxyTest.java
```diff
@@ -58,11 +71,22 @@ void stop_should_stop_server_if_server_started() throws ExecutionException, Inte
...
+        when(peerForwarderProvider.isPeerForwardingRequired()).thenReturn(true);
 
         final PeerForwarderServerProxy objectUnderTest = createObjectUnderTest();
         objectUnderTest.start();
         objectUnderTest.stop();
         verify(server).stop();
     }
 
+    @Test
+    void stop_should_do_nothing_if_noop_server_is_started() {
```
chenqi0805, line 83: This test name is confusing. Could you explain what it is testing? Looks like server should not be started or stopped since isPeerForwardingRequired() returns false. Maybe `no_server_interaction_if_peer_forwarding_not_required`?

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
```diff
@@ -110,13 +110,18 @@ private void buildPipelineFromConfiguration(
                     .map(this::newProcessor)
                     .collect(Collectors.toList());
 
-            final List<List<Processor>> decoratedProcessorSets = processorSets.stream()
-                    .map(processorSet -> processorSet.stream()
+            final Map<List<Processor>, String> processorMap = pipelineConfiguration.getProcessorPluginSettings().stream()
```
dlvenable, line 113: The order here is very important.

I'd much rather have a wrapper around `Processor` rather than trying to force a `Map` here.

data-prepper-core/src/test/java/org/opensearch/dataprepper/peerforwarder/RemotePeerForwarderTest.java
```diff
@@ -78,7 +78,7 @@ void test_forwardRecords_with_two_local_ips_should_process_record_two_record_loc
     void test_forwardRecords_with_one_local_ip_and_one_remote_ip_should_process_record_one_record_locally() {
         AggregatedHttpResponse aggregatedHttpResponse = mock(AggregatedHttpResponse.class);
         when(aggregatedHttpResponse.status()).thenReturn(HttpStatus.OK);
-        when(peerForwarderClient.serializeRecordsAndSendHttpRequest(anyCollection(), anyString(), anyString())).thenReturn(aggregatedHttpResponse);
+        when(peerForwarderClient.serializeRecordsAndSendHttpRequest(anyCollection(), anyString(), anyString(), anyString())).thenReturn(aggregatedHttpResponse);
```
dlvenable, line 81: These tests should verify the actual input strings.

data-prepper-core/src/main/java/org/opensearch/dataprepper/peerforwarder/discovery/DiscoveryMode.java
```diff
@@ -13,7 +13,8 @@
 public enum DiscoveryMode {
     STATIC(StaticPeerListProvider::createPeerListProvider),
     DNS(DnsPeerListProvider::createPeerListProvider),
-    AWS_CLOUD_MAP(AwsCloudMapPeerListProvider::createPeerListProvider);
+    AWS_CLOUD_MAP(AwsCloudMapPeerListProvider::createPeerListProvider),
+    LOCAL(LocalPeerListProvider::createPeerListProvider);
```
dlvenable, line 17: I think it might make sense to name this `SINGLE_NODE` and also `SingleNodePeerListProvider`. Thoughts?

data-prepper-core/src/main/java/org/opensearch/dataprepper/peerforwarder/PeerForwarderProvider.java
```diff
@@ -5,30 +5,84 @@
...
+        }
+        else {
+            return new LocalPeerForwarder();
         }
+    }
+
+    private PeerForwarderReceiveBuffer<Record<?>> createBufferPerPipelineProcessor(final String pipelineName, final String pluginId) {
+        final PeerForwarderReceiveBuffer<Record<?>> peerForwarderReceiveBuffer = new
+                PeerForwarderReceiveBuffer<>(peerForwarderConfiguration.getBufferSize(), peerForwarderConfiguration.getBatchSize());
+        if (pipelinePeerForwarderReceiveBufferMap.containsKey(pipelineName)) {
```
dlvenable, line 55: You can make this code more readable using `computeIfAbsent`.

```
Map<String, PeerForwarderReceiveBuffer<Record<?>>> pluginsBufferMap = pipelinePeerForwarderReceiveBufferMap.computeIfAbsent(pipelineName, new HashMap<>());

pluginsBufferMap.put(pluginId, peerForwarderReceiveBuffer);
```

dlvenable (changes requested):

data-prepper-core/src/main/java/org/opensearch/dataprepper/peerforwarder/discovery/DiscoveryMode.java
chenqi0805: IMO local is fine. SingleNode is not very indicative of local instance/node unless I misunderstand this DiscoveryMode

asifsmohammed: `SINGLE_NODE` can be a confusing customer experience as they can define single node even in static list. Maybe a `LOCAL_NODE`?

data-prepper-core/src/test/java/org/opensearch/dataprepper/peerforwarder/RemotePeerForwarderTest.java
asifsmohammed: We can check if records exists in the test on line 64 as both records in the test are processed locally. In this test I'm not sure how we can test it as we are not sure of which record will be processed locally. 
Is there a hamcrest method for it?

data-prepper-core/src/test/java/org/opensearch/dataprepper/peerforwarder/server/PeerForwarderServerProxyTest.java
asifsmohammed: Yeah, it's a bit confusing. peerForwarderProvider.isPeerForwardingRequired() return false so it'll create a `NoOpPeerForwarderServer`. I can rename the test

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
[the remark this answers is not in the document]
asifsmohammed: It's a good catch, I'll update it.

asifsmohammed: Like have a wrapper which returns processor name? This again leads to having plugin id's, which makes this logic simple here.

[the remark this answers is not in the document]
asifsmohammed: Actually, we can't make this map<processorName,  List<Processor> because we allow multiple processors of same type.

data-prepper-core/src/main/java/org/opensearch/dataprepper/peerforwarder/discovery/DiscoveryMode.java
dlvenable: I'm fine with `LOCAL_NODE` as well. And we can change it any time prior to the release if we want.

As a point of reference, OpenSearch uses the "single node" concept: `discovery.type=single-node`

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
dlvenable: Something like this:


```
 private static class IdentifiedComponent<T> {
        private final T component;
        private final String name;

        private IdentifiedComponent(final T component, final String name) {
            this.component = component;
            this.name = name;
        }

        T getComponent() {
            return component;
        } 
        
        String getName() {
            return name;
        }
    }

    private List<IdentifiedComponent<Processor>> newProcessor(final PluginSetting pluginSetting) {
        final List<Processor> processors = pluginFactory.loadPlugins(
                Processor.class,
                pluginSetting,
                actualClass -> actualClass.isAnnotationPresent(SingleThread.class) ?
                        pluginSetting.getNumberOfProcessWorkers() :
                        1);
        
        return processors.stream()
                .map(p -> new IdentifiedComponent<>(p, pluginSetting.getName()))
    }
```

You will need to change how to handle the data coming back from `newProcessors`, but that shouldn't be too challenging.

asifsmohammed: I made the changes but the static class will not be used once we have plugin ids.

data-prepper-core/src/test/java/org/opensearch/dataprepper/peerforwarder/server/PeerForwarderHttpServiceTest.java
```diff
@@ -92,7 +98,21 @@ void test_doPost_with_bad_HTTP_request_should_return_BAD_REQUEST() throws Execut
         assertThat(aggregatedHttpResponse.status(), equalTo(HttpStatus.INTERNAL_SERVER_ERROR));
     }
 
-    private AggregatedHttpRequest generateRandomValidHTTPRequest() throws JsonProcessingException,
+    @Test
+    void test() throws ExecutionException, JsonProcessingException, InterruptedException {
```
chenqi0805, line 102: A more specific test name?

dlvenable (dismissed):
Please update that `test` name as pointed out by @chenqi0805 . Thanks!

chenqi0805 (approved):

PR #1760, opened 2022-09-13
Load pipeline configurations from directory

### LINKED ISSUE #1736: Load pipeline files from the pipelines directory

oeyh:
### Description
With this change, data prepper loads pipeline configurations from `pipelines/` folder under data prepper home directory. The configurations can be specified in one yaml file or in multiple yaml files as specified in #305. 

Smoke tests were also updated.
 
### Issues Resolved
Resolves #1736  
 
### Check List

### DIFF
--- a/data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
+++ b/data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
@@ -91,6 +91,7 @@ public class PipelineParser {
             }
             return pipelineMap;
         } catch (IOException e) {
+            LOG.error("Failed to parse the configuration file {}", pipelineConfigurationFileLocation);
             throw new ParseException(format("Failed to parse the configuration file %s", pipelineConfigurationFileLocation), e);
         }
     }
@@ -104,22 +105,28 @@ public class PipelineParser {
             FileFilter yamlFilter = pathname -> (pathname.getName().endsWith(".yaml") || pathname.getName().endsWith(".yml"));
             List<InputStream> configurationFiles = Stream.of(configurationLocation.listFiles(yamlFilter))
                     .map(file -> {
+                        InputStream inputStream;
                         try {
-                            return new FileInputStream(file);
+                            inputStream = new FileInputStream(file);
+                            LOG.info("Reading pipeline configuration from {}", file.getName());
                         } catch (FileNotFoundException e) {
-                            return null;
+                            inputStream = null;
+                            LOG.warn("Pipeline configuration file {} not found", file.getName());
                         }
+                        return inputStream;
                     })
                     .filter(Objects::nonNull)
                     .collect(Collectors.toList());
 
             if (configurationFiles.isEmpty()) {
+                LOG.error("Pipelines configuration file not found at {}", pipelineConfigurationFileLocation);
                 throw new ParseException(
                         format("Pipelines configuration file not found at %s", pipelineConfigurationFileLocation));
             }
 
             return new SequenceInputStream(Collections.enumeration(configurationFiles));
         } else {
+            LOG.error("Pipelines configuration file not found at {}", pipelineConfigurationFileLocation);
             throw new ParseException(format("Pipelines configuration file not found at %s", pipelineConfigurationFileLocation));
         }
     }
--- a/data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
+++ b/data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
@@ -95,28 +95,29 @@ public class PipelineParser {
         }
     }
 
-    private InputStream mergePipelineConfigurationFiles() {
+    private InputStream mergePipelineConfigurationFiles() throws IOException {
         final File configurationLocation = new File(pipelineConfigurationFileLocation);
 
         if (configurationLocation.isFile()) {
-            try {
-                return new FileInputStream(configurationLocation);
-            } catch (FileNotFoundException e) {
-                throw new ParseException(format("Pipeline configuration file %s not found.", configurationLocation), e);
-            }
+            return new FileInputStream(configurationLocation);
         } else if (configurationLocation.isDirectory()) {
             FileFilter yamlFilter = pathname -> (pathname.getName().endsWith(".yaml") || pathname.getName().endsWith(".yml"));
-            List<InputStream> configurationFiles = Stream.of(configurationLocation.listFiles(yamlFilter)).map(file -> {
-                try {
-                    return new FileInputStream(file);
-                } catch (FileNotFoundException e) {
-                    throw new ParseException("Pipelines configuration file not found", e);
-                }
-            }).collect(Collectors.toList());
+            List<InputStream> configurationFiles = Stream.of(configurationLocation.listFiles(yamlFilter))
+                    .map(file -> {
+                        try {
+                            return new FileInputStream(file);
+                        } catch (FileNotFoundException e) {
+                            return null;
+                        }
+                    })
+                    .filter(Objects::nonNull)
+                    .collect(Collectors.toList());
+
             if (configurationFiles.isEmpty()) {
                 throw new ParseException(
                         format("Pipelines configuration file not found at %s", pipelineConfigurationFileLocation));
             }
+
             return new SequenceInputStream(Collections.enumeration(configurationFiles));
         } else {
             throw new ParseException(format("Pipelines configuration file not found at %s", pipelineConfigurationFileLocation));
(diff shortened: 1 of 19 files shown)


release/archives/linux/data-prepper-x64.sh
```diff
@@ -5,6 +5,15 @@
...
 
+if [[ $# -gt 0 ]]
+  then
+    echo
+    echo "Data Prepper expects no command line arguments. Configuration files should be placed in "
+    echo "the application home directory and will be read automatically."
+    echo
+    exit 1
+fi
+
```
oeyh, line 16: This is to address this comment: https://github.com/opensearch-project/data-prepper/pull/1737#discussion_r969938840

I'm also thinking about adding back the support for the old command line arguments as I noticed it might still benefits certain use cases. It's easier to switch between different configs from command line arguments than finding the right directories and swapping config files there.

dlvenable (changes requested):
Thanks! This looks good, and I just have a couple small comments.

release/archives/linux/data-prepper-x64.sh
oeyh: Cool. I will add it back in this PR then.

data-prepper-core/src/test/java/org/opensearch/dataprepper/parser/PipelineParserTests.java
```diff
@@ -121,6 +121,28 @@ void testMultipleProcessors() {
...
+    @Test
+    void parseConfiguration_from_directory_with_single_file_creates_the_correct_pipelineMap() {
+        final PipelineParser pipelineParser = new PipelineParser(TestDataProvider.SINGLE_FILE_PIPELINE_DIRECTOTRY, pluginFactory, peerForwarderProvider);
+        final Map<String, Pipeline> actualPipelineMap = pipelineParser.parseConfiguration();
+        assertThat(actualPipelineMap.keySet(), equalTo(TestDataProvider.VALID_MULTIPLE_PIPELINE_NAMES));
+    }
+
+    @Test
+    void parseConfiguration_from_directory_with_no_files_should_throw() {
+        final PipelineParser pipelineParser = new PipelineParser(TestDataProvider.EMPTY_PIPELINE_DIRECTOTRY, pluginFactory, peerForwarderProvider);
```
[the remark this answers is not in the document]
oeyh: I changed the filename and added some content to the file (which will not get parsed) in [01b44b0](https://github.com/opensearch-project/data-prepper/pull/1760/commits/01b44b023d92662cd45885233fedb7545708a624).

release/archives/linux/data-prepper-x64.sh
oeyh: Did it in [f0adae3](https://github.com/opensearch-project/data-prepper/pull/1760/commits/f0adae387c780d28f776a3ff02401d42cbc738d8)

dlvenable (dismissed):

cmanning09 (commented):
Can we update some of our documentation / README's to highlight this change?

- https://github.com/opensearch-project/data-prepper/blob/main/docs/getting_started.md#running
- https://github.com/opensearch-project/data-prepper/blob/fbe6873564c36eb1851a0fcacf752c6b018af669/docs/configuration.md#configuration

There are more docs but I would just focus on these. We highlight how to run data prepper in the main docs and we can update the other ones when we deprecate the older functionality later.

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
```diff
@@ -88,6 +95,35 @@ public Map<String, Pipeline> parseConfiguration() {
...
+            return new FileInputStream(configurationLocation);
+        } else if (configurationLocation.isDirectory()) {
+            FileFilter yamlFilter = pathname -> (pathname.getName().endsWith(".yaml") || pathname.getName().endsWith(".yml"));
+            List<InputStream> configurationFiles = Stream.of(configurationLocation.listFiles(yamlFilter))
+                    .map(file -> {
+                        try {
+                            return new FileInputStream(file);
+                        } catch (FileNotFoundException e) {
+                            return null;
+                        }
```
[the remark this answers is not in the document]
dlvenable: This is used only within the stream and the nulls are immediately filtered out. Using `null` here seems quite reasonable to me.

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
```diff
@@ -88,6 +95,35 @@ public Map<String, Pipeline> parseConfiguration() {
...
     }
 
+    private InputStream mergePipelineConfigurationFiles() throws IOException {
+        final File configurationLocation = new File(pipelineConfigurationFileLocation);
+
+        if (configurationLocation.isFile()) {
+            return new FileInputStream(configurationLocation);
+        } else if (configurationLocation.isDirectory()) {
+            FileFilter yamlFilter = pathname -> (pathname.getName().endsWith(".yaml") || pathname.getName().endsWith(".yml"));
+            List<InputStream> configurationFiles = Stream.of(configurationLocation.listFiles(yamlFilter))
```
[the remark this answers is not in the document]
dlvenable: Great catch! Maybe we can add this as a future feature?

oeyh:
>Can we update some of our documentation / README's to highlight this change?

Yes. We do have a sub task to update documentation, but it makes sense to start doing some of them here.

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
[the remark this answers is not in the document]
oeyh: It doesn't support recursive loading here. I opened an issue #1769 as a future feature to consider.

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/PipelineParser.java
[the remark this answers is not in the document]
oeyh: Added more logging in [5c7ab86](https://github.com/opensearch-project/data-prepper/pull/1760/commits/5c7ab86f5e1da8b8a159c08d61d683cbb6c10573). The pipeline files loaded will show up in the logs like this:
image

shared-config/log4j2-rolling.properties
```diff
@@ -33,11 +33,11 @@ rootLogger.level = warn
 rootLogger.appenderRef.stdout.ref = STDOUT
 rootLogger.appenderRef.file.ref = RollingFile
 
-logger.pipeline.name = com.amazon.dataprepper.pipeline
+logger.pipeline.name = org.opensearch.dataprepper.pipeline
```
dlvenable, line 36: Thanks for catching these!

dlvenable (approved):

release/archives/linux/data-prepper-jdk-x64.sh
```diff
@@ -5,6 +5,25 @@
...
+if [[ $# == 0 ]]; then
+    echo "Reading pipelines and data-prepper configuration files from Data Prepper home directory."
+elif [[ $# == 2 ]]; then
+    echo
+    echo "Data Prepper now supports reading pipeline and data-prepper configuration files"
+    echo "from Data Prepper home directory automatically."
+    echo "You can continue to specify paths to configuration files as command line arguments,"
+    echo "but that support will be dropped in a future release."
+    echo
+else
```
asifsmohammed, line 17: I think we accept single argument as well.

dlvenable: Do we in 1.x? I think the user must supply both pipeline and configuration files.

asifsmohammed (approved):

release/archives/linux/data-prepper-jdk-x64.sh
asifsmohammed: > Do we in 1.x? I think the user must supply both pipeline and configuration files.

No we don't. But we just need to clean up `DataPrepperArgs` class.

PR #1798, opened 2022-09-23
removed pipelines dir is smoke test container

sbayer55:

### Description
Removed pipelines directory from docker-compose file, in Data Prepper volumes.
 
### Issues Resolved
Fix broken smoke test
 
### Check List

### DIFF
--- a/release/smoke-tests/docker-compose.yml
+++ b/release/smoke-tests/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     working_dir: /usr/share/data-prepper/
     volumes:
       - ./data-prepper/config/data-prepper-config.yaml:/usr/share/data-prepper/config/data-prepper-config.yaml
-      - ./data-prepper/config/pipelines.yaml:/usr/share/data-prepper/pipelines/pipelines.yaml
+      - ./data-prepper/config/pipelines.yaml:/usr/share/data-prepper/pipelines.yaml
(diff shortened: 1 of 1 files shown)


release/smoke-tests/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
     working_dir: /usr/share/data-prepper/
     volumes:
       - ./data-prepper/config/data-prepper-config.yaml:/usr/share/data-prepper/config/data-prepper-config.yaml
-      - ./data-prepper/config/pipelines.yaml:/usr/share/data-prepper/pipelines/pipelines.yaml
+      - ./data-prepper/config/pipelines.yaml:/usr/share/data-prepper/pipelines.yaml
```
oeyh, line 10: The `pipelines/` dir was added to accommodate changes in 2.0. Is it causing you trouble to test a previous version? 

If that's the case, can we add a line without `pipelines/` dir instead? That might make it work for both cases.

sbayer55: Should the current main branch read from `usr/share/data-prepper/pipelines/pipelines.yaml` or is that a planned change?

oeyh: Current main branch should do that.

sbayer55: Thanks! I found the issue. When running the smoke test, it used an image I built previously.

PR #1819, opened 2022-09-27
Fix jdk version check in script

oeyh:

### Description
This PR makes a small change to the jdk version parser to handle cases that contain more than 3 numbers in version spec.

Before this PR, a version like "17.0.4.1" would be parsed as "17.0.4" and causing a `(standard_in) 1: syntax error` during comparison with minimum required version. With the change, "17.0.4.1" will be parsed as "17.0" so it can be compared numerically.
 
### Issues Resolved
N/A
Issue was mentioned here: https://forum.opensearch.org/t/latest-data-prepper-not-working/11055
 
### Check List


dlvenable (approved):
Thanks for improving the behavior of this script!

asifsmohammed (approved):

PR #2062, opened 2022-12-02
Setup boilerplate for RSS Source Plugin

### LINKED ISSUE #2041: Setup RSS Source boilerplate code
Add project setup for RSS Source.

sshivanii:
### Description
* Adds `Document` EventType to data-prepper-api model
* Adds module `rss-source` to `data-prepper-plugins`
* Creates a skeleton `RSSSource` class extending Source interface
* Adds `RSSSourceConfig` with the pipeline config options in the first iteration of rss-source
 
### Issues Resolved
#2041 
 
### Check List


sshivanii:
Noticed the build failure due to coverage ratio in `data-prepper-api`, looking into it

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/document/Document.java
```diff
@@ -0,0 +1,16 @@
...
+
+package org.opensearch.dataprepper.model.document;
+
+import org.opensearch.dataprepper.model.event.Event;
+
+/**
+ * A document event in Data Prepper represents a document type.
+ * @since 2.1
+ */
+public interface Document extends Event {
```
cmanning09, line 14: Why do we need a new Document event type? What feature's are we planning to add here? Are there special Document features we are planning to support and need standardized? 

I want to make sure we aren't building one off event types that will not be re-used. The S3 source plugin read documents from S3, yet we are using the generic event type. Ref: https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3ObjectWorker.java#L66

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/event/EventType.java
```diff
@@ -11,14 +11,16 @@
 import java.util.stream.Collectors;
 
 /**
- * An enumm to represent the event types supported in Data Prepper.
+ * An enum to represent the event types supported in Data Prepper.
```
cmanning09, line 14: Thank you!

data-prepper-plugins/rss-source/src/main/java/org/opensearch/dataprepper/plugins/source/rss/RSSSource.java
```diff
@@ -0,0 +1,43 @@
...
+public class RSSSource implements Source {
+
+    private static final Logger LOG = LoggerFactory.getLogger(RSSSource.class);
+
+    private final PluginMetrics pluginMetrics;
+
+    private final RSSSourceConfig rssSourceConfig;
+
+    @DataPrepperPluginConstructor
+    public RSSSource(PluginMetrics pluginMetrics, final RSSSourceConfig rssSourceConfig) {
```
cmanning09, line 27: minor: parameters be final. 

Same for `start(...)`

data-prepper-plugins/rss-source/src/main/java/org/opensearch/dataprepper/plugins/source/rss/RSSSource.java
```diff
@@ -0,0 +1,43 @@
...
+import org.opensearch.dataprepper.model.annotations.DataPrepperPlugin;
+import org.opensearch.dataprepper.model.annotations.DataPrepperPluginConstructor;
+import org.opensearch.dataprepper.model.buffer.Buffer;
+import org.opensearch.dataprepper.model.source.Source;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@DataPrepperPlugin(name = "rss", pluginType = Source.class, pluginConfigurationType =  RSSSourceConfig.class)
+public class RSSSource implements Source {
```
cmanning09, line 18: Take it or leave it: I would recommend adding the unit tests as well. There are some simple config validation test cases you could implement now. Adding them later when the plugin is implemented works as well. I just don't want us to forget.

data-prepper-plugins/rss-source/src/main/java/org/opensearch/dataprepper/plugins/source/rss/RSSSourceConfig.java
```diff
@@ -0,0 +1,31 @@
...
+public class RSSSourceConfig {
+
+    static final Duration DEFAULT_POLLING_FREQUENCY = Duration.ofMinutes(5);
+
+    @JsonProperty("url")
+    @NotBlank(message = "RSS Feed URL cannot be null or empty")
+    private String url;
+
+    @JsonProperty("pollingFrequency")
+    private Duration pollingFrequency = DEFAULT_POLLING_FREQUENCY;
```
cmanning09, line 22: Is there a range we plan on supporting?

data-prepper-plugins/rss-source/src/main/java/org/opensearch/dataprepper/plugins/source/rss/RSSSourceConfig.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class RSSSourceConfig {
+
+    static final Duration DEFAULT_POLLING_FREQUENCY = Duration.ofMinutes(5);
+
+    @JsonProperty("url")
+    @NotBlank(message = "RSS Feed URL cannot be null or empty")
+    private String url;
+
+    @JsonProperty("pollingFrequency")
```
cmanning09, line 21: What is our expected format for `pollingFrequency`? Can the JsonProperty annotation handle the Duration object conversion?  Can you add a unit test demonstrating this?

data-prepper-plugins/rss-source/src/main/java/org/opensearch/dataprepper/plugins/source/rss/RSSSourceConfig.java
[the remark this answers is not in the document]
sshivanii: I understand the definition difference. However, the reason it's being called `pollingFrequency` is due to retrieving the RSS feed URL every n minutes, or how frequently we retrieve the refreshed feed. I'd avoid calling it Period related since period in Java usually refers to date-based values. We can sync-up if you still think it should be renamed.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/document/Document.java
sshivanii: Good question. The reason we want to add `Document` event type is to support full-text search use case. The converted RSS items that the user would want to use are documents which correspond to webpages from the RSS feed. Currently, the `EventType` enum supports only 3 use cases: Log, Trace, Metric. The RSS docs do not fit into any of these categories. By introducing a Document type, we can fit other future text use cases in this EventType, making it extensible. Details about which plugins to support are still under way but it did come up in a design discussion.

As for S3, while building the `Event` it does end up building a `JacksonLog`. There's 3 implementations of the parse method that ends up creating instances of `JacksonLog`:
* [JsonCodec](https://github.com/opensearch-project/data-prepper/blob/8cd0f870afe74dca34187a6d31034c166ec44d48/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/codec/JsonCodec.java#L57)
* [CsvCodec](https://github.com/opensearch-project/data-prepper/blob/8cd0f870afe74dca34187a6d31034c166ec44d48/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/codec/CsvCodec.java#L85)
* [NewlineDelimitedCodec](https://github.com/opensearch-project/data-prepper/blob/8cd0f870afe74dca34187a6d31034c166ec44d48/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/codec/NewlineDelimitedCodec.java#L76)

kkondaka: Thanks for the explanation. I have couple of questions Since the underlying object is still an Event(which is essentially a map), what would be stored in the map for the document? How does that help with "full text search"

sshivanii: @kkondaka - Each RSS `item` will be a corresponding `Event` as mentioned in the Design doc. Since we need a new `EventType` to categorize RSS documents, I think it makes sense to have a corresponding Document interface to extend `Event` even though there's no required feature at the moment.

data-prepper-plugins/rss-source/src/main/java/org/opensearch/dataprepper/plugins/source/rss/RSSSourceConfig.java
sshivanii: Added a simple unit test asserting the default here.

data-prepper-plugins/rss-source/src/main/java/org/opensearch/dataprepper/plugins/source/rss/RSSSourceConfig.java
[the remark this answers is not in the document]
chenqi0805: Discussed offline. The format will be a notation string, e.g.

```
pollingFrequency: 5m
```

chenqi0805 (approved):

data-prepper-plugins/rss-source/src/main/java/org/opensearch/dataprepper/plugins/source/rss/RSSSourceConfig.java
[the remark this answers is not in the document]
sshivanii: This will be called out in the RSS Source documentation. Duration values will support ISO-8601 format so `PT5M` for 5 minutes as well as notation Strings for seconds and milliseconds (e.g. 60s or 3600ms)

data-prepper-plugins/rss-source/src/main/java/org/opensearch/dataprepper/plugins/source/rss/RSSSourceConfig.java
cmanning09: I am not concerned about the default. I am concerned about the JSON object being parsed into this object. I would like to see a test demonstrating that. I am concerned the current format will not support parsing the config correctly. Feel free to add it in a follow-up PR.

cmanning09 (approved):

PR #2198, opened 2023-01-27
Add Support for retry when Sink fails to initialize 

### LINKED ISSUE #936: Data Prepper requires OpenSearch is available or will fail

A Data Prepper pipeline with the OpenSearch sink requires that OpenSearch be running and reachable in order to run. This is a rather confusing user experience. It is unclear if OpenSearch is unreachable in general, or just starting up.

Data Prepper pipelines with OpenSearch as a sink should be able to start and wait for the OpenSearch sink rather than fail.

One approach that could work well is to have a concept of an un-started pipeline. Data Prepper can create the different sinks and then wait for them to be ready before starting up the sources. In this way, Data Prepper won't accept data in the source until the sink is ready.

This would probably require adding new methods to `Sink`:

```
void initialize();
boolean isReady();
```

Data Prepper could just run and let the buffer fill up until the sink is ready. This is simpler than the concept proposed above.

**Additional context**

Here is a forum post with a user of Data Prepper who was help up because of the current behavior.

https://discuss.opendistrocommunity.dev/t/data-preeper-plugin/8319

kkondaka:
### Description
Add Support for retry when Sink fails to initialize.
During dataprepper initialization, isReady() call is made to each pipeline to make sure each pipeline is ready before starting the source. Source is not started even if one of the pipelines is not ready. 

Two new methods (`initialzie` `isReady`) are added to Sink interface. A retry thread is started if Sink is not ready. This retry thread periodically invokes `initialize` to initialize Sink until it is successful.
 
### Issues Resolved
Resolves #936 
 
### Check List


data-prepper-api/src/main/java/org/opensearch/dataprepper/model/sink/AbstractSink.java
```diff
@@ -43,4 +54,13 @@ public void output(Collection<T> records) {
      * @param records Records to be output
      */
     public abstract void doOutput(Collection<T> records);
+
+    @Override
+    public void shutdown() {
+        retryThread.stop();
+    }
+
+    public Thread.State getRetryThreadState() {
```
dlvenable, line 63: Do we need this? If it is only for testing, please make it package protected (remove the `public` modifier).

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/sink/AbstractSink.java
```diff
@@ -17,14 +17,25 @@
...
     protected final PluginMetrics pluginMetrics;
     private final Counter recordsInCounter;
     private final Timer timeElapsedTimer;
+    private final Thread retryThread;
 
     public AbstractSink(final PluginSetting pluginSetting) {
         this.pluginMetrics = PluginMetrics.fromPluginSetting(pluginSetting);
         recordsInCounter = pluginMetrics.counter(MetricNames.RECORDS_IN);
         timeElapsedTimer = pluginMetrics.timer(MetricNames.TIME_ELAPSED);
+        Integer numRetries = (Integer) pluginSetting.getAttributeFromSettings(NUM_RETRIES);
```
dlvenable, line 31: I tend to think that Data Prepper core should be responsible for determining the number of allowed retries, or a global time to success.

If we do want to stick with having the sinks configure them, we should still have each sink configuration provide the setting and then pass it into the `AbstractSink` class. We are removing `PluginSetting` in favor or plugin configuration classes. The current work is somewhat incompatible with that upcoming change.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/sink/AbstractSink.java
```diff
@@ -17,14 +17,25 @@
  * This class implements the Sink interface and records boilerplate metrics
  */
 public abstract class AbstractSink<T extends Record<?>> implements Sink<T> {
+    static private final String NUM_RETRIES = "num_retries";
```
dlvenable, line 20: Please rename to `number_retries` to support our configuration consistency efforts. Using full words helps maintain consistency.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/sink/AbstractSink.java
```diff
@@ -17,14 +17,25 @@
...
 
     public AbstractSink(final PluginSetting pluginSetting) {
         this.pluginMetrics = PluginMetrics.fromPluginSetting(pluginSetting);
         recordsInCounter = pluginMetrics.counter(MetricNames.RECORDS_IN);
         timeElapsedTimer = pluginMetrics.timer(MetricNames.TIME_ELAPSED);
+        Integer numRetries = (Integer) pluginSetting.getAttributeFromSettings(NUM_RETRIES);
+        if (numRetries == null) {
+            numRetries = NUM_DEFAULT_RETRIES;
+        }
+        retryThread = new Thread(new SinkThread(this, numRetries));
```
dlvenable, line 35: I think this class design could be re-worked some.

Data Prepper core should call `initialize()` on the sink.

The `AbstractSink` can then provide a `doInitialize()` method which sub-classes will implement. Thus, `AbstractSink` can implement `initialize()` to have the code you have here:

```
retryThread = new Thread(...);
if(!isReady()) {
  retryThread.start();
}
```

data-prepper-core/src/main/java/org/opensearch/dataprepper/DataPrepper.java
```diff
@@ -71,6 +75,28 @@ public DataPrepper(
...
+            while (pipelineIter.hasNext()) {
+                String pipelineName = (String)pipelineIter.next();
+                if (!transformationPipelines.get(pipelineName).isReady()) {
+                    failedPipelines.add(pipelineName);
+                }
+            }
+            pipelineNames = failedPipelines;
+            if (pipelineNames.size() > 0) {
+                try {
+                    Thread.sleep(5000);
```
dlvenable, line 92: This could be configurable in data-prepper-config.yaml. But, that could come as a later improvement.

data-prepper-core/src/main/java/org/opensearch/dataprepper/DataPrepper.java
```diff
@@ -71,6 +75,28 @@ public DataPrepper(
      */
     public boolean execute() {
         peerForwarderServer.start();
+        Set<String> pipelineNames = transformationPipelines.keySet();
+        int numRetries = 0;
+        while (pipelineNames.size() > 0 && numRetries++ < MAX_RETRIES) {
+            Set<String> failedPipelines = new HashSet<String>();
```
dlvenable, line 81: `failedPipelines` seems a little inaccurate. Perhaps `uninitializedPipelines`? At first glance it seems that these are failed for good.

data-prepper-core/src/main/java/org/opensearch/dataprepper/DataPrepper.java
```diff
@@ -71,6 +75,28 @@ public DataPrepper(
...
+            }
+            pipelineNames = failedPipelines;
+            if (pipelineNames.size() > 0) {
+                try {
+                    Thread.sleep(5000);
+                } catch (Exception e){}
+            }
+        }
+        if (pipelineNames.size() > 0) {
+            LOG.info("One or more Sinks are not ready even after {} retries.", numRetries);
```
dlvenable, line 97: This class is dealing with pipelines not sinks. It may be good to say "One or more Pipelines are not ready...".

data-prepper-core/src/main/java/org/opensearch/dataprepper/DataPrepper.java
```diff
@@ -71,6 +75,28 @@ public DataPrepper(
      */
     public boolean execute() {
         peerForwarderServer.start();
+        Set<String> pipelineNames = transformationPipelines.keySet();
```
dlvenable, line 78: This could use a better name to improve readability. Perhaps `waitingPipelineNames`?

dlvenable (changes requested):

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/sink/AbstractSink.java
kkondaka: Yes, it is only for testing. Will remove the "public"

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/sink/AbstractSink.java
kkondaka: I am OK with setting it to fixed value in the AbstractSink class. Shall we go with that approach?

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/sink/AbstractSink.java
kkondaka: As mentioned above. I could remove this completely and use a fixed (not configurable) value.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/sink/AbstractSink.java
kkondaka: I thought about that approach but derived Sink class's (like OpenSearch Sink) constructor would not have got called by the time Abstract Sink called initialized() on the OpenSearch Sink. So, calling initialize() in OpenSearchSInk before it's constructor got invoked doesn't make sense. Correct me if my understanding is wrong.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -87,37 +90,41 @@ public OpenSearchSink(final PluginSetting pluginSetting) {
     this.routingField = openSearchSinkConfig.getIndexConfiguration().getRoutingField();
     this.action = openSearchSinkConfig.getIndexConfiguration().getAction();
     this.indexManagerFactory = new IndexManagerFactory();
+    this.initialized = false;
+    this.lock = new ReentrantLock(true);
+  }
 
+  @Override
+  public void initialize() {
```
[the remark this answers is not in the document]
kkondaka: Will add check to make sure initialize() is skipped if it is already initialized.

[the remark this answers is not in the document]
graytaylor0: It's less about skipping intialize if it's already been called. It's more in the case where initialization fails halfway through, and making sure that each step that was run before is a no-op when retried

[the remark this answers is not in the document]
kkondaka: I think, in that respect the function is idempotent. We do check to see if index is present or not before creating it.

data-prepper-plugins/opensearch/README.md
```diff
@@ -88,6 +88,8 @@ Default is null.
 
 - `index_type` (optional): a String from the list [`custom`, `trace-analytics-raw`, `trace-analytics-service-map`, `management-disabled`], which represents an index type. Defaults to `custom`. This index_type instructs Sink plugin what type of data it is handling. 
 
+- `num_retries` (optional): a Number indicating the number of times the sink initialization should be tried before failing the pipeline. The retry mechanism waits for 1 second before each retry. Default value 600.
```
dlvenable, line 91: Please remove this now that it is no longer configurable.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -87,16 +90,20 @@ public OpenSearchSink(final PluginSetting pluginSetting) {
...
     try {
-      initialize();
-    } catch (final IOException e) {
-      this.shutdown();
-      throw new RuntimeException(e.getMessage(), e);
+        doInitialize();
+    } catch (IOException e) {
+        LOG.warn("Failed to initialize OpenSearch sink " + e.getMessage());
+    } catch (Exception e) {
+        this.shutdown();
```
dlvenable, line 100: Why do you shutdown this sink here? I think we want to keep trying right?

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/sink/AbstractSinkTest.java
```diff
@@ -69,7 +96,55 @@ public void doOutput(Collection<Record<String>> records) {
...
+        }
+
+        @Override
+        public void shutdown() {
+            super.shutdown();
+        }
+
+        @Override
+        public void doInitialize() throws Exception {
+            if (++initCount == 3) {
```
dlvenable, line 138: I think you are trying to make this initialize on the last retry right? You can make the static constant package protected in `AbstractSink` and then use it here. Just remove the `private` modifier there.

If somebody changes that value, this test might end up failing.

data-prepper-plugins/common/src/main/java/org/opensearch/dataprepper/plugins/sink/FileSink.java
```diff
@@ -54,12 +55,14 @@ public FileSink() {
     public FileSink(final String outputFile) {
         this.outputFilePath = outputFile == null ? SAMPLE_FILE_PATH : outputFile;
         isStopRequested = false;
+        initialized = false;
```
dlvenable, line 58: It might be better to move the code in between these two changes into the `initialize()` method. This will make it a better example for any contributors who look at how the code works.

dlvenable (changes requested):

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
kkondaka: Not in this case. This is not case where it cannot be retried. When it is possible to retry, we don't throw exception (like in case of `catch (IOException e) {`

dlvenable: We are catching a generic `Exception`. How do we know that it is not retryable?

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/sink/AbstractSinkTest.java
kkondaka: In `AbstractSink`? That's not test code. This is specific to Testing, right? I can make it static code but this number needs to be smaller than the number of retries I have above, which is 10. I will make these number as static constants with init count less than max retries count.

graytaylor0 (dismissed):

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/sink/AbstractSink.java
```diff
@@ -17,14 +17,31 @@
...
+    public abstract void doInitialize() throws Exception;
+
+    @Override
+    public void initialize() {
+        try {
+            doInitialize();
+        } catch (Exception e) {
+        }
+        if (!isReady()) {
+            retryThread = new Thread(new SinkThread(this, NUMBER_OF_RETRIES));
```
asuresh8, line 42: Why would you retry in a separate thread vs retrying in the code right here?

kkondaka: Because we want to check for multiple sinks in multiple pipelines in parallel.

engechas (dismissed):

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
dlvenable: This is throwing for a generic `Exception`. This may or may not be retryable. I think you can shutdown on the `OpenSearchSinkException`. But, any other exception is probably too much.

dlvenable (changes requested):

data-prepper-plugins/common/src/main/java/org/opensearch/dataprepper/plugins/sink/FileSink.java
kkondaka: Made the changes as suggested.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
kkondaka: Only IOExceptions and very few OpenSearchExceptions are retryable as per this check
```
public boolean canRetry(final Exception e) {
        return (e instanceof IOException ||
                (e instanceof OpenSearchException &&
                        !NON_RETRY_STATUS.contains(((OpenSearchException) e).status().getStatus())));
    }
```

dlvenable: I think the problem  here is that we still calling `doInitialize` in the constructor. I think if you remove this and then add the exception handling into `doInitialize` it would work.

```
public void doInitialize() {
  try {
    ...
  } catch(OpenSearchSinkException ex) {
     shutdown();
  } catch(Exception ex) { 
     // Let it retry.
     LOG.warn(...)
  }
}
```

dlvenable (approved):
Looks good. Thanks for making this important change!

engechas (approved):

PR #2246, opened 2023-02-06
Add back LinkedBlockingQueue::poll to avoid busy wait

engechas:

### Description
Avoids busy waiting when the pipeline is idle by re-adding the LinkedBlockingQueue::poll functionality. 
 
### Check List


data-prepper-core/src/main/java/org/opensearch/dataprepper/peerforwarder/PeerForwarderReceiveBuffer.java
```diff
@@ -84,24 +84,49 @@ public void writeAll(final Collection<T> records, final int timeoutInMillis) thr
...
+    // TODO - consolidate duplicate logic in BlockingBuffer
     @Override
     public Map.Entry<Collection<T>, CheckpointState> read(final int timeoutInMillis) {
-        final List<T> records = new ArrayList<>();
+        final List<T> records = new ArrayList<>(batchSize);
+        int recordsRead = 0;
 
         if (timeoutInMillis == 0) {
-            blockingQueue.drainTo(records, batchSize);
+            final T record = pollForBufferEntry(5, TimeUnit.MILLISECONDS);
```
asuresh8, line 94: Use a longer poll time than 5ms. I suggested 1 minute, but really Is there any penalty to making this an infinitely long poll time?

engechas: I tested with 5 seconds and it choked the throughput. I don't have an explanation for this as my understanding is also that there should not be a penalty for longer durations.

The idle CPU utilization is single digit %s even with 5ms here. 
```
32466 root      20   0   32.8g  24.8g  27056 S   4.0  6.6  22:59.48 java
32465 root      20   0   32.8g  28.8g  27192 S   3.7  7.7  19:10.67 java
```

I would like to raise the poll time as well but I don't have an answer as to why it lowers the performance significantly. Perhaps there is something I am missing?

asuresh8: Ah I see that poll takes a long for the up to the duration of timeout: http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/LinkedBlockingQueue.java#l456. Since the buffer can have multiple producers and consumers this will result in a lot of time spent waiting on the lock. 5ms look good for now, and we can mark this as something to optimize further in the future.

asuresh8 (approved):

data-prepper-core/src/main/java/org/opensearch/dataprepper/peerforwarder/PeerForwarderReceiveBuffer.java
engechas: Scratch the last comment - the explanation is that in the case of uneven load in a multi-node scenario, we end up bottlenecking on the pipeline buffer poll call.

Example:
2 node pipeline, all of the load is sent to node 1. Node one processes the data quickly and forwards half the records to node 2. Node 2 has empty pipeline buffers since it has no direct traffic, but its CPF buffers are filling. Each iteration of `ProcessWorker::doRun` on node 2 has to wait for the poll timeout before being able to process the data it received from CPF.

The same issue begins to occur on node 1 as its CPF buffer is always empty.

There shouldn't be a legitimate use case where traffic is drastically uneven enough to cause the above scenario, but there seems no harm in having a low poll timeout since it still solves the busy wait issue

engechas: I just tested with even to both nodes and poll of 5s, it still throttles the throughput pretty significantly. I think your comment is correct

dlvenable (approved):

PR #2250, opened 2023-02-07
Refactor to remove stream in RemotePeerForwarder as micro optimization

engechas:
### Description
Streams are less efficient than ordinary collection methods. Since `forwardBatchedRecords` is executed many times, the slight optimization of using ordinary collection methods makes a non-negligible difference in CPU utilization

 
### Check List


dlvenable (approved):
Did you see an actual improvement from this change?

Also, I think there may be an opportunity to avoid using `Map` here entirely. In the end you just iterate over the objects.

engechas:
> Did you see an actual improvement from this change?
> 
> Also, I think there may be an opportunity to avoid using `Map` here entirely. In the end you just iterate over the objects.

Yes, it dropped the CPU cycle% by 1/3 for the `forwardBatchedRecords` method. This PR won't make a significant difference in performance, but it was low hanging fruit

asifsmohammed (approved):

PR #2265, opened 2023-02-13
Long int fix

kkondaka:
### Description
Data Prepper should support long integer literals . Without this support conditionals like `/durationInNanos >=100000000` are failing with the error 
```
2023-02-10T23:04:34,582 [span-pipeline-processor-worker-5-thread-1] ERROR org.opensearch.dataprepper.expression.ParseTreeEvaluator - Unable to evaluate event
org.opensearch.dataprepper.expression.ExpressionCoercionException: Unsupported type for value 530159171
        at org.opensearch.dataprepper.expression.ParseTreeCoercionService.resolveJsonPointerValue(ParseTreeCoercionService.java:67) ~[data-prepper-expression-2.1.0-SN
APSHOT.jar:?]
        at org.opensearch.dataprepper.expression.ParseTreeCoercionService.coercePrimaryTerminalNode(ParseTreeCoercionService.java:35) ~[data-prepper-expression-2.1.0-
SNAPSHOT.jar:?]
```
 
### Check List

### DIFF
--- a/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/OperatorConfiguration.java
+++ b/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/OperatorConfiguration.java
@@ -29,37 +29,83 @@ class OperatorConfiguration {
         final Map<Class<? extends Number>, BiFunction<Object, Object, Boolean>> intOperations =
                 Map.of(
                         Integer.class, (lhs, rhs) -> (Integer) lhs > (Integer) rhs,
-                        Float.class, (lhs, rhs) -> (Integer) lhs > (Float) rhs
+                        Float.class, (lhs, rhs) -> (Integer) lhs > (Float) rhs,
+                        Long.class, (lhs, rhs) -> (Integer) lhs > (Long) rhs,
+                        Double.class, (lhs, rhs) -> (Integer) lhs > (Double) rhs
                 );
         final Map<Class<? extends Number>, BiFunction<Object, Object, Boolean>> floatOperations =
                 Map.of(
                     Integer.class, (lhs, rhs) -> (Float) lhs > (Integer) rhs,
-                    Float.class, (lhs, rhs) -> (Float) lhs > (Float) rhs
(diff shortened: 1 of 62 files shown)


data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/OperatorConfiguration.java
```diff
@@ -29,37 +29,83 @@ public NumericCompareOperator greaterThanOperator() {
...
+
+        final Map<Class<? extends Number>, BiFunction<Object, Object, Boolean>> longOperations =
+                Map.of(
+                    Integer.class, (lhs, rhs) -> (Long) lhs > (Integer) rhs,
+                    Float.class, (lhs, rhs) -> (Long) lhs > (Float)rhs,
+                    Long.class, (lhs, rhs) -> (Long) lhs > (Long) rhs,
+                    Double.class, (lhs, rhs) -> (Long) lhs > (Double)rhs
+                );
+
+        final Map<Class<? extends Number>, BiFunction<Object, Object, Boolean>> doubleOperations =
```
graytaylor0, line 52: Seems like there could be a less manual way to create these combinations

graytaylor0 (dismissed):

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/OperatorConfiguration.java
kkondaka: It may be possible. But that's something we should try as a separate issue. Given the urgency to fix this issue, I took this approach.

dlvenable (changes requested):
I think it might make sense to include an integration test in `ConditionalExpressionEvaluatorIT` which exactly reproduces your current issue.

Ideally, you could even test this without your change and see that the test fails.

kkondaka:
@dlvenable added tests to `ConditionalExpressionEvaluatorIT` to test long integer behavior.

dlvenable (approved):

PR #2295, opened 2023-02-21
adding backwards compatibility support in versioning check

cmanning09:
### Description
This is a follow on of #2292

Adds backward compatibility support to version check. In any major version, we should have migration fields to help with migrations. So a pipeline created for 1.5 could be compatible for 2.0. But, not always. This change will make it easier for migrating between major versions.

### Issues Resolved
None
 
### Check List

### DIFF
--- a/data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/DataPrepperVersion.java
+++ b/data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/DataPrepperVersion.java
@@ -58,9 +58,8 @@ public class DataPrepperVersion {
     /**
      * Determines if the provided Data Prepper Version is compatible with this.
      *
-     * The initial implementation is a basic implementation that enforces equivalent versions, any shorthand format version
-     * compared with full format with equivalent major versions and equivalent major version but the comparing minor version
-     * is less than this minor version are compatible.
+     * The initial implementation is a basic implementation that will return true only when the comparing version is
+     * less than or equal to this version.
      *
      * @param o - the other DataPrepperVersion to compare with
      * @return return true if the versions are compatible, otherwise false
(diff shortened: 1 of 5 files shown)


data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/DataPrepperVersion.java
```diff
@@ -58,21 +58,20 @@ public Optional<Integer> getMinorVersion() {
...
+     * less than or equal to this version.
      *
      * @param o - the other DataPrepperVersion to compare with
      * @return return true if the versions are compatible, otherwise false
      * @since 2.1
      */
     public boolean compatibleWith(DataPrepperVersion o) {
 
-        if (this.majorVersion != o.getMajorVersion()) {
+        if (this.majorVersion < o.getMajorVersion()) {
```
graytaylor0, line 70: Does the logic to return true if `this.majorVersion > o.getMajorVersion()` make sense? If it's arbitrary comparison why would 1 < 2 return false/incompatible and 2 < 1 return true/compatible?

dlvenable (approved):

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/DataPrepperVersion.java
cmanning09: We are trying to determine if `this` is compatible with the version defined in `o`. `this` should be backwards compatible and support older versions and return true if the version when `o` is smaller and false when `o` is greater.

graytaylor0: Is this assuming that this did not contain breaking changes to older versions?

cmanning09: Yes, it is assuming. Its a best effort.

graytaylor0 (approved):

PR #2348, opened 2023-03-01
Add grok pattern validator to data-prepper-api

graytaylor0:
Signed-off-by: Taylor Gray tylgry@amazon.com

### Description
Add support validate grok patterns from data-prepper-api
### Issues Resolved
 
### Check List


dlvenable (changes requested):
NIce! I like the idea here. But, this is adding a grok dependency to all plugins. I'd like to keep `data-prepper-api` using only the minimum dependencies.

A few possible alternatives:
* Create a `grok-api` library which supports any validations needed for Grok processing. this has the added benefit of being able to coordinate code in that library and in the `grok-processor` project.
* Create a new validations library which is only used by select projects. This seems to end up splitting logic between processors the validator library.

PR #2395, opened 2023-03-22
Wire in source_coordination config and SourceCoordinator interface to Source plugins

graytaylor0:
### Description
This adds boilerplate and wiring for taking a `source_coordination` configuration from the `data-prepper-config.yaml` and instantiating the correct type of `SourceCoordinator`

The `SourceCoordinator` interface is provided to sources that implement `RequiresSourceCoordination`. This interface is not implemented or finalized, but is a starting point to allow for calls to this interface from the source.

The S3Source now implements `RequiresSourceCoordination` as a proof of concept for passing the `SourceCoordinator` interface (I can remove this from the S3Source before merging since it is unused)
 
### Check List


data-prepper-core/src/main/java/org/opensearch/dataprepper/sourcecoordination/DynamoDbSourceCoordinator.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private final SourceCoordinationConfig sourceCoordinationConfig;
+
+    public DynamoDbSourceCoordinator(final SourceCoordinationConfig sourceCoordinationConfig) {
+        this.sourceCoordinationConfig = sourceCoordinationConfig;
+    }
+
+    @Override
+    public void createPartitions(List<String> partitionKeys) {
+
```
chenqi0805, line 31: Nit: TODO

data-prepper-core/src/main/java/org/opensearch/dataprepper/sourcecoordination/DynamoDbSourceCoordinator.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public Optional<SourcePartition> getNextPartition() {
+        return Optional.empty();
+    }
+
+    @Override
+    public void completePartition(String partitionKey) {
+
```
chenqi0805, line 41: same as above

data-prepper-core/src/main/java/org/opensearch/dataprepper/sourcecoordination/DynamoDbSourceCoordinator.java
graytaylor0: I can throw an `UnsupportedOperationException` as well instead if that makes sense

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3Source.java
```diff
@@ -14,17 +14,20 @@
...
+import org.opensearch.dataprepper.model.source.RequiresSourceCoordination;
 import org.opensearch.dataprepper.model.source.Source;
+import org.opensearch.dataprepper.model.source.SourceCoordinator;
 import org.opensearch.dataprepper.plugins.source.codec.Codec;
 import org.opensearch.dataprepper.plugins.source.ownership.BucketOwnerProvider;
 import org.opensearch.dataprepper.plugins.source.ownership.ConfigBucketOwnerProviderFactory;
 
 @DataPrepperPlugin(name = "s3", pluginType = Source.class, pluginConfigurationType = S3SourceConfig.class)
-public class S3Source implements Source<Record<Event>> {
+public class S3Source implements Source<Record<Event>>, RequiresSourceCoordination {
```
dlvenable, line 25: I actually like the interface idea. But, seeing this here I realize that the source coordination is variably needed.

The S3 source does not require it for the SQS notification type. It only requires when it is configured to scan. Only in those cases do we want it.

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/model/sourcecoordination/SourceCoordinationStoreConfig.java
```diff
@@ -0,0 +1,21 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+/**
+ * The interface to be implemented to support different database store configurations
+ * for {@link org.opensearch.dataprepper.model.source.SourceCoordinator} implementations
+ * @since 2.2
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)
+@JsonSubTypes({ @JsonSubTypes.Type(DynamoDBSourceCoordinationStoreConfig.class)})
+public interface SourceCoordinationStoreConfig {
```
dlvenable, line 18: This needs to move to a shared library to make this pluggable outside of data-prepper-core.

Perhaps we should consider having a new library for extending core functionality. Another possible target for this would be extending the metric providers. But, I'm also fine with continuing to use data-prepper-api.

dlvenable (changes requested):

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/SourceCoordinator.java
```diff
@@ -0,0 +1,56 @@
...
+import java.time.Duration;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Interface to be passed to {@link org.opensearch.dataprepper.model.source.Source} to enable distribution and coordination of work
+ * between source plugins, and to track progress and save state for these partitions of work
+ * @since 2.2
+ */
+public interface SourceCoordinator {
```
dlvenable, line 17: This can be generic which would make development easier for sources.

This snippet shows the change that would make it generic.
```
public interface SourceCoordinator<T> {
  Optional<SourcePartition<T>> getNextPartition();
  void saveStateForPartition(final String partitionKey, final T partitionState);
}
```

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/SourcePartition.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.Objects;
+
+/**
+ * The class that will be provided to {@link org.opensearch.dataprepper.model.source.Source} plugins
+ * that implement {@link org.opensearch.dataprepper.model.source.RequiresSourceCoordination} to identify the partition of
+ * data that the source should process
+ * @since 2.2
+ */
+public class SourcePartition {
```
dlvenable, line 16: This also can be generic.

```
public class SourcePartition<T> {
  public T getPartitionState();
  public static class Builder<T> {
    public Builder withPartitionState(final T partitionState)
  }
}
```

dlvenable (changes requested):

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3Source.java
graytaylor0: Yes it is not always needed. If no `source_coordination` config is specified in the `data-prepper-config.yaml`, then the `SourceCoordinator` class that is passed to the source will be null. The source can then decide whether it actually needs it or not.

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/model/sourcecoordination/DynamoDBSourceCoordinationStoreConfig.java
```diff
@@ -0,0 +1,40 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+/**
+ * the configuration for utilizing DynamoDB as the distributed store for {@link org.opensearch.dataprepper.model.source.SourceCoordinator}
+ * @since 2.2
+ */
+public class DynamoDBSourceCoordinationStoreConfig implements SourceCoordinationStoreConfig {
```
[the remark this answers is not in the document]
graytaylor0: I would consider this to be pluggable wouldn't you? Unless you mean it is actually creating as a `@DataPrepperPlugin`.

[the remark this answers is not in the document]
dlvenable: Yes, make this a `@DataPrepperPlugin`.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/SourceCoordinator.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * A partition will be created for each of the partitionKeys passed.
+     * @param partitionKeys
+     */
+    void createPartitions(final List<String> partitionKeys);
```
dlvenable, line 23: Should these partitions have some sort of ordering?

For reading from S3 this probably is the string provided here. But, for an OpenSearch index, we might want to provide the `creation_date`.

Maybe this should be provided as an additional string?

```
class Partition {
  String key;
  String orderKey;
}

...

createPartitions(final<Partition> partitions);
```

In this way, we could list indexes in OpenSearch and we know the latest created index.

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/model/sourcecoordination/SourceCoordinationStoreConfig.java
dlvenable: I think it would look this way:

```
data-prepper-api/
  SourceCoordinationStoreConfig

data-prepper-core/
  SourceCoordinationConfig

data-prepper-plugins/
  dynamodb-source-coordinator/
    DynamoDBSourceCoordinationStoreConfig
    DynamoStoreSettings
```

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3Source.java
dlvenable: That makes sense. So if the source needs it and it is not available, the source will throw an exception.

I did just realize though - why not just make this injectable via the `@DataPrepperConstructor`? We won't even need the interface.

dlvenable (changes requested):

data-prepper-core/src/main/java/org/opensearch/dataprepper/parser/model/sourcecoordination/SourceCoordinationStoreConfig.java
dlvenable: Correction - I don't think we need the `SourceCoordiationStoreConfig`. Once this uses the plugin interface you probably won't need that class at all.

dlvenable (changes requested):

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3Source.java
```diff
@@ -57,4 +60,14 @@ public void start(Buffer<Record<Event>> buffer) {
     public void stop() {
         sqsService.stop();
     }
+
+    @Override
```
dlvenable, line 64: I think perhaps we should not yet include this. It helps to see it, but until the S3 source actually needs it we might want to keep it out.

dlvenable (changes requested):
This is looking great! My only comment is I think we should remove this from the S3 source until it is fully ready and needed.

dlvenable (approved):
Thanks!

kkondaka (approved):
Looks good.

PR #2451, opened 2023-04-05
adding data-time patterns to key-path-prefix and creating readme for S3 DLQ

### LINKED ISSUE #2298: Support an S3 DLQ in OpenSearch

cmanning09:
### Description
1. adding data-time patterns to key-path-prefix and creating readme for s3 dlq.
2. adding detailed README for S3 DLQ Plugin
 
### Issues Resolved
- resolves #2298
 
### Check List

### DIFF
--- a/data-prepper-plugins/failures-common/src/main/java/org/opensearch/dataprepper/plugins/dlq/s3/KeyPathGenerator.java
+++ b/data-prepper-plugins/failures-common/src/main/java/org/opensearch/dataprepper/plugins/dlq/s3/KeyPathGenerator.java
@@ -1,0 +1,86 @@
+package org.opensearch.dataprepper.plugins.dlq.s3;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.TimeZone;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class KeyPathGenerator {
+
+    private static final String TIME_PATTERN_INTERNAL_EXTRACTOR_REGULAR_EXPRESSION  = "%\\{(.*?)\\}";
+    private static final String TIME_PATTERN_STARTING_SYMBOLS = "%{";
+    private static final String TIME_PATTERN_REGULAR_EXPRESSION  = "%\\{.*?\\}";
+    static final ZoneId UTC_ZONE_ID = ZoneId.of(TimeZone.getTimeZone("UTC").getID());
+    private static final Set<Character> INVALID_CHARS = ImmutableSet.of('#', '\\', '/', '*', '?', '"', '<', '>', '|', ',', ':');
+    private static final Set<Character> UNSUPPORTED_TIME_GRANULARITY_CHARS = ImmutableSet.of('m', 's', 'S', 'A', 'n', 'N');
+
+    private final String keyPathPrefix;
+    private List<DateTimeFormatter> dateFormatters;
+    public KeyPathGenerator(final String keyPathPrefix) {
+        this.keyPathPrefix = keyPathPrefix;
+        this.dateFormatters = keyPathPrefix == null ? Collections.emptyList() : getDatePatternFormatter(keyPathPrefix);
+    }
+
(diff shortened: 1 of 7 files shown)


cmanning09:
Checkstyle failures on aggregate processor is failing the builds

cmanning09:
> Checkstyle failures on aggregate processor is failing the builds

Addressed: https://github.com/opensearch-project/data-prepper/commit/0312caf322ffa304abca45aef744ecff94f19f9c

data-prepper-plugins/failures-common/src/main/java/org/opensearch/dataprepper/plugins/dlq/s3/KeyPathGenerator.java
```diff
@@ -0,0 +1,86 @@
+package org.opensearch.dataprepper.plugins.dlq.s3;
```
kkondaka, line 1: Looks like this is somewhat duplicate of the code in OpenSearchSink. Also Wipro is writing some code that seems to be similar to this. We should have a plan to consolidate this.

kkondaka (approved):

data-prepper-plugins/failures-common/src/main/java/org/opensearch/dataprepper/plugins/dlq/s3/KeyPathGenerator.java
cmanning09: 100%, it is slightly similar but there are differences. I think we should standardize this. There was talk here #2424 as well for providing a way to provide patterns for fetching keys. I am thinking we should you at this holistically at how to we build a framework or tooling that allows any plugin to leverage some of these features.

Issue #2412, opened 2023-03-29
Source Coordination in Data Prepper


graytaylor0:
**Is your feature request related to a problem? Please describe.**
Data Prepper has many push based sources, such as `http`, `otel_trace_source`, etc. Distributing data between multiple instances of data prepper can easily be solved with a load balancer.

However, pull based sources of Data Prepper do not have a Data Prepper internal way to coordinate which work is done between different instances of Data Prepper in a multi-node scenario. For example, pulling data from something like an OpenSearch cluster with 5 nodes of Data Prepper would result in all 5 nodes pulling the entirety of the data and processing it 5 times total.

**Describe the solution you'd like**
A core data prepper solution for pull based sources to distribute data between multiple instances of data prepper, and a way to track the progress of the data that is pulled to skip processing of duplicate data.

This solution could use a distributed store to coordinate and track progress of the data. The store could be pluggable and configured in the `data-prepper-config.yaml`. The store type could range from Remote/Local File DB, Apache Zookeeper, MySQL, DynamoDB, and more.

**Describe alternatives you've considered (Optional)**

**Additional context**

PR #2460, opened 2023-04-07
Lease based source coordinator

graytaylor0:
### Description
This change implements the `LeaseBasedSourceCoordinator`, which is responsible for using `SourceCoordinationStores` to manage source partitions
 
### Issues Resolved
Related to #2412 
 
### Check List


data-prepper-api/build.gradle
```diff
@@ -16,11 +16,11 @@ dependencies {
 }
 
 jacocoTestCoverageVerification {
-    dependsOn jacocoTestReport
+    dependsOn(jacocoTestReport)
     violationRules {
         rule { //in addition to core projects rule - this one checks for 100% code coverage for this project
             limit {
-                minimum = 1.0 //keep this at 100%
+                minimum = 0.95
```
graytaylor0, line 23: Unfortunately, this is completely necessary. Even after my new code contained 100% test coverage, the build was still failing. I created the `@SkipTestCoverageGenerated` annotation and even after adding to spots that shouldn't be included it was still failing at 1.0

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/coordinator/SourceCoordinator.java
```diff
@@ -20,7 +20,7 @@
     /**
```
engechas, line 20: What do you think about simplifying this interface? A consumer of this interface is only concerned with 2 high-level ideas: identifying which partition it is assigned and notifying that is has finished processing a partition.

The underlying logic for constructing partitions, reserving them, and then returning the `SourcePartition` object would occur as part of `getPartition`. The interface could be
```
public interface SourceCoordinator<T> {
  public Optional<SourcePartition<T>> getPartition(final Supplier<List<PartitionIdentifier>> partitionIdentifiersSupplier);
  public void deletePartition(final PartitionIdentifier partitionIdentifier);
}
```
And the LeaseBasedSourceCoord implementation would be along the lines of
```
public class LeaseBasedSourceCoordinator<T> {
  public Optional<SourcePartition<T>> getPartition(final Supplier<List<PartitionIdentifier>> partitionIdentifiersSupplier) {
    // Check existing partitions for availability
    final List<SourcePartition<T>> existingSourcePartitions = sourceCoordinationStore.listPartitions();
    final List<SourcePartition<T>> existingSourcePartitions.stream()
          .filter(this::isPartitionAvailable)
          .collect(...);

    // If no partitions are available, try to build more
    final List<PartitionIdentifier> partitionIdentifiers = partitionIdentifiersSupplier.get();
    // filter the above list against the existingSourcePartitions fetched above
    // update the store with the new partitions, reserving one for the current node
    sourceCoordinationStore.updatePartitions(...);
    // return the reserved partition if present, otherwise empty if none were available
  }

  public void deletePartition(final PartitionIdentifier partitionIdentifier) {
    sourceCoordinationStore.deletePartition(partitionIdentifier);
  }

  private boolean isPartitionAvailable(final SourcePartition<T> sourcePartition) {
    if (isLeaseExpired) {
      evictLease(sourcePartition);
      return true;
    }

    return sourcePartition.owner() == null;
  }

  private boolean isLeaseExpired(final SourcePartition<T> sourcePartition) {
    return time.now - LEASE_TIMEOUT > sourcePartition.reservationTime();
  }

  // This would have to be done in some sort of transactional way to avoid concurrent mods
  private void evictLease(final SourcePartition<T> sourcePartition) {
    sourceCoordinationStore.updatePartition(updateRequest);
  }
}
```

The `SourceCoordinationStore` would then be more of a CRUD accessor that facilitates transactions to the underlying data store. The `SourceCoordinator` implementations would handle the business logic surrounding partition selection, reservation, updates, deletion, etc. There is some complexity around concurrent modifications/reads with this approach.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/coordinator/SourcePartition.java
```diff
@@ -15,18 +15,18 @@
  */
 public class SourcePartition<T> {
```
engechas, line 16: There may need to be more metadata in this object. Info about who owns it and when they reserved it, etc. Even for a future non-lease based approach, knowing which node tried to process it and when would be helpful for debugging.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/coordinator/SourceCoordinator.java
```diff
@@ -29,44 +29,56 @@
...
+     *
+     * @param partitionKey - The partition key that uniquely identifies the partition of work
+     * @param reopenAfter  - The duration from the current time to wait before this partition should be processed further at a later date
+     * @return Whether the partition was closed successfully or not. If this returns false, the operation is safe to retry.
+     * @throws org.opensearch.dataprepper.model.source.coordinator.exceptions.PartitionNotFoundException if the partition key could not be found in the distributed store
+     * @throws org.opensearch.dataprepper.model.source.coordinator.exceptions.PartitionNotOwnedException if the partition is not owned by this instance of SourceCoordinator
      * @since 2.2
      */
-    void closePartition(final PartitionIdentifier partitionIdentifier, final Duration reopenAfter);
+    boolean closePartition(final String partitionKey, final Duration reopenAfter);
```
engechas, line 60: Could you elaborate on the use case for closing a partition before it is fully processed?

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/coordinator/SourcePartitionStoreItem.java
```diff
@@ -20,4 +20,12 @@ public interface SourcePartitionStoreItem {
     Instant getPartitionOwnershipTimeout();
```
engechas, line 20: How do the `SourcePartitionStoreItem` and `SourcePartition` differ in usage? At a glance it seems like they are both representations of data in the partition store, I'm wondering if there's room to reconcile them into a single object

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/coordinator/SourceCoordinator.java
graytaylor0: Yes. So closing a partition should only happen to a point where the whole partition has been processed that can be up to this point in time. So in the case of an s3 object, we probably want to just complete the partition as we processed the whole thing.

However, this gives sources the option to not mark the partition as completed, and say that more data will be included in this partition in the future, and at some point we should reopen it to process the new data. 

An example of this could be an OpenSearch index that we process all the documents for that exist at a certain point in time, and then at some point in the future, we want to process this index again to pick up new documents that have been indexed after the intitial processing of the index.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/coordinator/SourcePartition.java
graytaylor0: This object is supplied to the sources. I could see value in providing reserve time and maybe the time that this partition ownership will timeout.

However, by definition if this is returned to the source then the owner is this instance, and I don't think the source should concern itself with this. the core source coordination logic will always be tracking and logging where needed who owns a partition and who processes it

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/coordinator/SourceCoordinator.java
graytaylor0: I think ideally the store would be a CRUD operation, but like you said this adds a lot of complexity with the lease management in `LeaseBasedSourceCoordinator`, as we would have to synchronize, which is very difficult to do without having the store itself handle conflicts. So with this class design, yes the `SourceCoordinationStore` is responsible for handling race condition conflicts between nodes, but it is still similar to a CRUD accessor with tryCreate, tryUpdate, and get operations. If we can implement a `LeaseBasedSourceCoordinator` that was able to just call `create`, `update`, and `get` and handle synchronization on its end, that'd be great. But it would be much more difficult and definitely increase development time by a lot.

As far as the interface simplification, I think the only real simplification possible that you're suggesting is combining the function for `createPartitions` and `getPartition`. I don't think we can get rid of `saveState`, `closePartition`, `completePartition`, or `giveUpPartitions` (for shutdown). I think they all serve an important purpose here.

I also don't think I am in favor of combining the creation of partitions with the `getPartition` call. `createPartitions` and `getPartition` should in most cases be called on a different schedule depending on the source, and there is also a strong consideration depending on performance to only have a single instance of a data prepper cluster be the one creating partitions, which is made easier as a separate function.

engechas (dismissed):

data-prepper-core/src/main/java/org/opensearch/dataprepper/sourcecoordination/LeaseBasedSourceCoordinator.java
```diff
@@ -5,58 +5,273 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(LeaseBasedSourceCoordinator.class);
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    static final Duration DEFAULT_LEASE_TIMEOUT = Duration.ofMinutes(10);
+
+    private static final String hostName;
+
     private final SourceCoordinationConfig sourceCoordinationConfig;
     private final SourceCoordinationStore sourceCoordinationStore;
+    private final PartitionManager<T> partitionManager;
```
kkondaka, line 42: Do we not need any Locks to protect these? I assume this is called by multiple threads at the same time?

data-prepper-core/src/main/java/org/opensearch/dataprepper/sourcecoordination/LeaseBasedSourceCoordinator.java
```diff
@@ -5,58 +5,273 @@
...
-        throw new UnsupportedOperationException();
+    public Optional<SourcePartition<T>> getNextPartition(final Supplier<List<PartitionIdentifier>> partitionsCreatorSupplier) {
+        if (partitionManager.getActivePartition().isPresent()) {
+            return partitionManager.getActivePartition();
+        }
+
+        Optional<SourcePartitionStoreItem> ownedPartitions = sourceCoordinationStore.tryAcquireAvailablePartition();
+
+        if (ownedPartitions.isEmpty()) {
+            // Potential metric (AcquirePartitionMisses). May indicate that the number of nodes could be lowered or new partitions need to be added
```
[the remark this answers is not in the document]
graytaylor0: Yes metrics will be added at the end after holistically looking at all the source coordination code.

data-prepper-core/src/main/java/org/opensearch/dataprepper/sourcecoordination/LeaseBasedSourceCoordinator.java
graytaylor0: This will only be called by a single thread at a time.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/source/SourceCoordinationStore.java
```diff
@@ -23,4 +23,13 @@ boolean tryCreatePartitionItem(final String partitionKey,
...
                                    final String partitionProgressState);
+
+    Optional<SourcePartitionStoreItem> tryAcquireAvailablePartition();
+
+    /**
+     * This method attempts to update the partition item to the desired state
+     * @throws org.opensearch.dataprepper.model.source.coordinator.exceptions.PartitionUpdateException when the partition was not updated successfully
+     * @param updateItem - The item to update in the source coordination store
+     */
+    void tryUpdateSourcePartitionItem(final SourcePartitionStoreItem updateItem);
```
cmanning09, line 34: Take it or leave it:`try` is redundant now.

cmanning09 (approved):

engechas (approved):

PR #2470, opened 2023-04-10
Retry flushing of buffer on buffer TimeoutException

graytaylor0:
### Description
* This change retries the flushing of the buffer for the S3Source when it receives a TimeoutException from the buffer
* Refactors the `S3SelectWorkerObjectTest` 
 
### Check List

### DIFF
--- a/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
+++ b/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
@@ -59,7 +59,7 @@ public class S3SelectObjectWorker implements S3ObjectHandler {
     private static final Logger LOG = LoggerFactory.getLogger(S3SelectObjectWorker.class);
 
     static final long MAX_S3_OBJECT_CHUNK_SIZE = 64 * 1024 * 1024;
-    private static final int MAX_FLUSH_RETRIES_ON_IO_EXCEPTION = 5;
+    private static final int MAX_FLUSH_RETRIES_ON_IO_EXCEPTION = Integer.MAX_VALUE;
     private static final Duration INITIAL_FLUSH_RETRY_DELAY_ON_IO_EXCEPTION = Duration.ofSeconds(5);
     static final String S3_BUCKET_NAME = "bucket";
     static final String S3_OBJECT_KEY = "key";
@@ -73,7 +73,7 @@ public class S3SelectObjectWorker implements S3ObjectHandler {
     private final Buffer<Record<Event>> buffer;
     private final String expression;
     private final S3SelectSerializationFormatOption serializationFormatOption;
-    private final S3SelectResponseHandler s3SelectResponseHandler;
+    private final S3SelectResponseHandlerFactory s3SelectResponseHandlerFactory;
     private final S3ObjectPluginMetrics s3ObjectPluginMetrics;
     private final CompressionType compressionType;
     private final BiConsumer<Event, S3ObjectReference> eventConsumer;
@@ -89,7 +89,7 @@ public class S3SelectObjectWorker implements S3ObjectHandler {
         this.expression = s3ObjectRequest.getExpression();
         this.serializationFormatOption = s3ObjectRequest.getSerializationFormatOption();
         this.s3AsyncClient = s3ObjectRequest.getS3AsyncClient();
-        this.s3SelectResponseHandler = s3ObjectRequest.getS3SelectResponseHandler();
+        this.s3SelectResponseHandlerFactory = s3ObjectRequest.getS3SelectResponseHandlerFactory();
         this.s3ObjectPluginMetrics = s3ObjectRequest.getS3ObjectPluginMetrics();
         this.compressionType = s3ObjectRequest.getCompressionType();
         this.eventConsumer = s3ObjectRequest.getEventConsumer();
@@ -130,6 +130,7 @@ public class S3SelectObjectWorker implements S3ObjectHandler {
                     .build();
             final SelectObjectContentRequest selectObjectContentRequest = getS3SelectObjRequest(s3ObjectReference, inputSerialization, scanRange);
 
+            final S3SelectResponseHandler s3SelectResponseHandler = s3SelectResponseHandlerFactory.provideS3SelectResponseHandler();
             s3AsyncClient.selectObjectContent(selectObjectContentRequest, s3SelectResponseHandler).join();
             if (s3SelectResponseHandler.getException() != null)
                 throw new IOException(s3SelectResponseHandler.getException());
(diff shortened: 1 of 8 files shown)


data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
```diff
@@ -191,6 +234,44 @@ private void parseCompleteStreamFromResponseHeader(final S3ObjectReference s3Obj
...
+                }
+            } catch (ExecutionException e) {
+                LOG.warn("Retrying of flushing the buffer accumulator hit an exception: {}", e.getMessage());
+            } catch (InterruptedException e) {
+                LOG.warn("Retrying of flushing the buffer accumulator was interrupted: {}", e.getMessage());
+            }
+        }
+
+
+        LOG.warn("Flushing the bufferAccumulator failed after {} attempts", maxRetries);
```
engechas, line 270: We should back off and retry indefinitely here IMO. If the current docs in the accumulator cannot be written to the buffer due to a timeout, pulling in more data doesn't seem like the correct behavior.

If a sink is unhealthy and the buffer is never clearing, we essentially stop the pipeline via infinite backoff and retry. Once the sink is healed, the process would pick back up where it left off without losing data or going OOM from the Accumulator continuing to accept data.

If we do go with infinite retry, we may want to use a linear backoff for responsiveness.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
```diff
@@ -94,91 +106,122 @@ public S3SelectObjectWorker(final S3ObjectRequest s3ObjectRequest) {
...
+        long startRange = 0;
+        long endRange = Math.min(MAX_S3_OBJECT_CHUNK_SIZE, objectSize);
+        while (startRange < objectSize) {
+            final ScanRange scanRange = ScanRange.builder()
+                    .start(startRange)
+                    .end(endRange)
+                    .build();
+            final SelectObjectContentRequest selectObjectContentRequest = getS3SelectObjRequest(s3ObjectReference, inputSerialization, scanRange);
+
+            s3AsyncClient.selectObjectContent(selectObjectContentRequest, s3SelectResponseHandler).join();
```
engechas, line 133: #2447 introduced issues with timeouts on larger files. These timeouts went away when reverting the client overrides set here: https://github.com/opensearch-project/data-prepper/pull/2447/files#diff-49e2eeca5ae076b473ca54aed8fb03db3edc2b074180cb59876f50ac72dcd6ffR52-R57

Could you include that revert in this PR? We can un-revert later if needed.

@udaych20 could you elaborate on why the defaults were overridden in that PR? I want to be sure there isn't an underlying issue that I have not yet encountered that necessitated the default timeouts being overridden.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
```diff
@@ -4,22 +4,12 @@
  */
```
engechas, line 4: I was getting an NPE in some of the testing from a bug in the OpenSearch retry strategy. If the entry in the map does not exist, an NPE is thrown when trying to unbox `null` into the primitive `int`.

https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkRetryStrategy.java#L205

This is unrelated to this PR but does affect the overall S3 select functionality. I was able to solve it locally by adding this above line 205:
```
if (!retryCountMap.containsKey(bulkRequestForRetry)) {
    retryCountMap.put(bulkRequestForRetry, 1);
}
```

Could you include that fix as well? @kkondaka might have some input on whether or not the above snippet is the correct solution

engechas (changes requested):

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
```diff
@@ -94,91 +106,122 @@ public S3SelectObjectWorker(final S3ObjectRequest s3ObjectRequest) {
...
+        while (startRange < objectSize) {
+            final ScanRange scanRange = ScanRange.builder()
+                    .start(startRange)
+                    .end(endRange)
+                    .build();
+            final SelectObjectContentRequest selectObjectContentRequest = getS3SelectObjRequest(s3ObjectReference, inputSerialization, scanRange);
+
+            s3AsyncClient.selectObjectContent(selectObjectContentRequest, s3SelectResponseHandler).join();
+            if (s3SelectResponseHandler.getException() != null)
+                throw new IOException(s3SelectResponseHandler.getException());
```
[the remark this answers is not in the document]
graytaylor0: What do you think the proper fix here is? To add methods to the `S3SelectResponseHandler` to clear the exception and the SelectObjectContentEventStream list?

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
graytaylor0: Makes sense. Can I keep this easily changeable by setting the retryCount to INTEGER.MAX_VALUE?

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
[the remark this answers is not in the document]
engechas: We could modify the `S3SelectResponseHandler` to clear properly between requests, but I don't see benefit in reusing the same response handler object. I would do away with passing it into this class as a member variable and instead instantiate a new one inline here.

We could deal with un-plumbing the member variable later and just instantiate a new one in the loop here for now

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
engechas: That works for me. I think the OpenSearch sink does the same thing with `Integer.MAX_VALUE`

engechas (approved):

dlvenable (approved):

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3SelectObjectWorker.java
udaych20: @engechas sometimes we are facing below exception while doing testing. so we have overriden the changes. 

2023-03-31T17:46:33,364 [Thread-1] ERROR org.opensearch.dataprepper.plugins.source.ScanObjectWorker - Error while process the parseS3Object. 
java.io.IOException: java.util.concurrent.CompletionException: software.amazon.awssdk.core.exception.SdkClientException: Unable to execute HTTP request: Acquire operation took longer than the configured maximum time. This indicates that a request cannot get a connection from the pool within the specified maximum time. This can be due to high request rate.

Consider taking any of the following actions to mitigate the issue: increase max connections, increase acquire timeout, or slowing the request rate.

Increasing the max connections can increase client throughput (unless the network interface is already fully utilized), but can eventually start to hit operation system limitations on the number of file descriptors used by the process. If you already are fully utilizing your network interface or cannot further increase your connection count, increasing the acquire timeout gives extra time for requests to acquire a connection before timing out. If the connections doesn't free up, the subsequent requests will still timeout.

If the above mechanisms are not able to fix the issue, try smoothing out your requests so that large traffic bursts cannot overload the client, being more efficient with the number of times you need to call AWS, or by increasing the number of hosts sending requests.

PR #2652, opened 2023-05-08
Pipeline creation should succeed even when sink(s) are not ready

kkondaka:
### Description
This change modifies, DataPrepper to make the pipeline creation succeed even when one or more sinks are not ready.
The source is not started until all sinks are ready. So, the ingestion is not started until all sinks are ready.

### Issues Resolved
 
### Check List


dlvenable (changes requested):
I see some tagging code here. Can you rebase your branch and push again with just the new changes?

data-prepper-core/src/main/java/org/opensearch/dataprepper/DataPrepper.java
```diff
@@ -83,29 +80,6 @@ public DataPrepper(
...
     public boolean execute() {
         peerForwarderServer.start();
-        Set<String> waitingPipelineNames = transformationPipelines.keySet();
-        int numRetries = 0;
-        while (waitingPipelineNames.size() > 0 && numRetries++ < MAX_RETRIES) {
-            Set<String> uninitializedPipelines = new HashSet<String>();
-            Iterator pipelineIter = waitingPipelineNames.iterator();
-            while (pipelineIter.hasNext()) {
-                String pipelineName = (String)pipelineIter.next();
-                if (!transformationPipelines.get(pipelineName).isReady()) {
```
cmanning09, line 93: Now that this logic is removed, does this method need to be a public method?

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/Pipeline.java
```diff
@@ -211,7 +211,21 @@ public void execute() {
                 final SourceCoordinator sourceCoordinator = sourceCoordinatorFactory.provideSourceCoordinator(partionProgressModelClass, name);
                 ((UsesSourceCoordination) source).setSourceCoordinator(sourceCoordinator);
             }
-            source.start(buffer);
+
+            sinkExecutorService.submit(() -> {
+                while (!isReady()) {
```
cmanning09, line 216: What happens if `/shutdown` is called while we are waiting for sinks to be ready? Can we call stop on a source that is not started yet? Have we tested this?

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/PipelineConnector.java
```diff
@@ -51,6 +53,10 @@ public void enableAcknowledgements() {
         areAcknowledgementsEnabled = true;
     }
 
+    public void setReady() {
+        ready = true;
+    }
```
cmanning09, line 58: I think a call to the  `start()` method should switch this connector to ready instead of having some external caller determine when this connector is ready.

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/PipelineConnector.java
```diff
@@ -51,6 +53,10 @@ public void enableAcknowledgements() {
         areAcknowledgementsEnabled = true;
     }
 
+    public void setReady() {
```
cmanning09, line 56: Missing unit tests

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/Pipeline.java
```diff
@@ -211,7 +211,21 @@ public void execute() {
                 final SourceCoordinator sourceCoordinator = sourceCoordinatorFactory.provideSourceCoordinator(partionProgressModelClass, name);
                 ((UsesSourceCoordination) source).setSourceCoordinator(sourceCoordinator);
             }
-            source.start(buffer);
+
+            sinkExecutorService.submit(() -> {
```
cmanning09, line 215: Can we add unit tests?

cmanning09:
Both builds (11 & 17) failed for the same 3 integration tests.

cmanning09 (approved):

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/Pipeline.java
```diff
@@ -211,7 +212,21 @@ public void execute() {
...
+                long retryCount = 0;
+                while (!isReady() && !isStopRequested()) {
+                    if (++retryCount % 60 == 0) {
```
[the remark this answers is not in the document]
kkondaka: That's every one minute. What do you think is more appropriate time interval?

data-prepper-core/src/main/java/org/opensearch/dataprepper/DataPrepper.java
```diff
@@ -83,29 +80,6 @@ public DataPrepper(
      */
     public boolean execute() {
         peerForwarderServer.start();
-        Set<String> waitingPipelineNames = transformationPipelines.keySet();
```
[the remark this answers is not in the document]
kkondaka: Not really. This is the newly added code by me in PR https://github.com/opensearch-project/data-prepper/pull/2198 (on Feb 2023). There should not be any implications of it.

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/Pipeline.java
```diff
@@ -211,7 +212,21 @@ public void execute() {
...
+                        LOG.info("Pipeline {} Waiting for Sink to be ready", name);
```
dlvenable, line 220: A lot of our logs wrap pipeline names in brackets. Let's keep that pattern.

Also, I see a lot of hyphens after this.

So maybe this:

```
LOG.info("Pipeline [{}] - Waiting for Sink to be ready", name);
```

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/Pipeline.java
```diff
@@ -211,7 +212,21 @@ public void execute() {
...
+                    }
+                    try {
+                        Thread.sleep(1000);
+                    } catch (Exception e){}
+                }
+                LOG.info("Pipeline {} Sink is ready, starting source...", name);
+                source.start(buffer);
+            }, null);
+
             LOG.info("Pipeline [{}] - Submitting request to initiate the pipeline processing", name);
```
dlvenable, line 230: Let's include this in the async thread that initialize the source in. And we can keep the existing initialization order as was currently there.

It could also help to move this into an inner class rather than a lambda since there is quite a bit there.

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/Pipeline.java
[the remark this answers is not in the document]
dlvenable: You are right. This is ok.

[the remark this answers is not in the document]
dlvenable: If we make this `retryCount++` we will log this sooner for users (before the the first sleep). And it should be easier to get code coverage on the log line.

dlvenable (changes requested):

dlvenable (approved):

cmanning09 (approved):

PR #2663, opened 2023-05-09
Avro codecs

### LINKED ISSUE #1532: Support generic parsers/codecs

umayr-codes:
### Description
Support for nested and non-primitive data types along with test cases added.
 
### Issues Resolved
Resolves #1532  
 
### Check List


dlvenable (changes requested):
@umayr-codes , Thank you for submitting this PR. It appears that this PR already has a lot of the changes which were previously committed by #2527.

Please re-work this PR so that it has only the changes you are making from that point.

You might need to rebase your changes against `main`.

umayr-codes:
closed in favour of #2715.

PR #2707, opened 2023-05-17
Add support for adding metadata entries

### LINKED ISSUE #2687: Add support for adding metadata entries as part of add_entries processor
Currently `add_entries` processor adds new fields to an `Event`. We need the ability to add entries to event's metadata.

Suggest the following change to `add_entries` processor

```
processor
   - add_entries:                                                                                                                                      
         entries:                                                                                                                                        
         - metadata_key: "message_len"                                                                                                                                    
            value: 10
```

or

```
processor
   - add_entries:                                                                                                                                      
         entries:                                                                                                                                        
         - metadata_key: "message_len"                                                                                                                                    
            value_expression: `length(/message)`
```

**Additional context**

kkondaka:
### Description
Adds support for adding metadata entries to the Add Entries processor.
A new configuration option `metadata_key:` may be specified to set an attribute in the metadata. Only of the `metadata_key` or `key` can be specified. Value options are same as existing value options.
 Resolves #2687 

### Issues Resolved
Resolves #2687 
 
### Check List

### DIFF
--- a/data-prepper-plugins/mutate-event-processors/README.md
+++ b/data-prepper-plugins/mutate-event-processors/README.md
@@ -56,7 +56,8 @@ then when we run with the same input, the processor will parse the message into 
 
 ### Configuration
 * `entries` - (required) - A list of entries to add to an event
-  * `key` - (required) - The key of the new entry to be added
+  * `key` - (required) - The key of the new entry to be added. One of `key` or `metadata_key` is required.
+  * `metadata_key` - (required) - The key of the new metadata attribute to be added. One of `key` or `metadata_key` is required.
   * `value` - (optional) - The value of the new entry to be added. Strings, booleans, numbers, null, nested objects, and arrays containing the aforementioned data types are valid to use. Required if `format` is not specified.
   * `format` - (optional) - A format string to use as value of the new entry to be added. For example, `${key1}-${ke2}` where `key1` and `key2` are existing keys in the event. Required if `value` is not specified.
   * `overwrite_if_key_exists` - (optional) - When set to `true`, if `key` already exists in the event, then the existing value will be overwritten. The default is `false`. 
--- a/data-prepper-plugins/mutate-event-processors/README.md
+++ b/data-prepper-plugins/mutate-event-processors/README.md
@@ -57,7 +57,7 @@ then when we run with the same input, the processor will parse the message into 
 ### Configuration
 * `entries` - (required) - A list of entries to add to an event
   * `key` - (required) - The key of the new entry to be added. One of `key` or `metadata_key` is required.
-  * `metadata_key` - (required) - The key of the new metadata attribute to be added. One of `key` or `metadata_key` is required.
+  * `metadata_key` - (required) - The key of the new metadata attribute to be added. Argument must be a literal string key and not JsonPointer. One of `key` or `metadata_key` is required.
   * `value` - (optional) - The value of the new entry to be added. Strings, booleans, numbers, null, nested objects, and arrays containing the aforementioned data types are valid to use. Required if `format` is not specified.
   * `format` - (optional) - A format string to use as value of the new entry to be added. For example, `${key1}-${ke2}` where `key1` and `key2` are existing keys in the event. Required if `value` is not specified.
   * `overwrite_if_key_exists` - (optional) - When set to `true`, if `key` already exists in the event, then the existing value will be overwritten. The default is `false`. 
--- a/data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorConfig.java
+++ b/data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorConfig.java
@@ -16,9 +16,9 @@ import java.util.Objects;
 
 public class AddEntryProcessorConfig {
     public static class Entry {
-        @NotEmpty
-        @NotNull
         private String key;
+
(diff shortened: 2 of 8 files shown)


data-prepper-plugins/mutate-event-processors/README.md
```diff
@@ -56,7 +56,8 @@ then when we run with the same input, the processor will parse the message into
 
 ### Configuration
 * `entries` - (required) - A list of entries to add to an event
-  * `key` - (required) - The key of the new entry to be added
+  * `key` - (required) - The key of the new entry to be added. One of `key` or `metadata_key` is required.
+  * `metadata_key` - (required) - The key of the new metadata attribute to be added. One of `key` or `metadata_key` is required.
```
graytaylor0, line 60: This doesn't support json pointer right? Should we mention that in the documentation here?

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorConfig.java
```diff
@@ -54,12 +58,20 @@ public boolean hasValueOrFormat() {
...
 
         public Entry(final String key,
+                     final String metadataKey,
                      final Object value,
                      final String format,
                      final boolean overwriteIfKeyExists,
                      final String addWhen)
         {
+            if (key != null && metadataKey != null) {
+                throw new IllegalArgumentException("Only one of the two - key and metadatakey - should be specified");
```
graytaylor0, line 68: I don't feel too strongly about it but is there any reason that we can
t set both at once? Seems like there may be a case where a value should be set in both metadata and the Event

data-prepper-plugins/mutate-event-processors/README.md
cmanning09: There are mutually exclusive in the code and it appears `key` takes presedence over `metadata`.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorConfig.java
cmanning09: @graytaylor0 , do you foresee a scenario where a user wants the same data in the event and metadata. With the current proposal a user can do that with:

```
  processor:
    - add_entries:
        entries:
        - key: "newMessage"
           format: "new ${message}"
           overwrite_if_key_exists: true
        - metadata_key: "newMessage"
           format: "new ${message}"
           overwrite_if_key_exists: true
```
I like this because it's explicit and follows our exiting model of one entry equaling one mutation.

data-prepper-plugins/mutate-event-processors/src/test/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorTests.java
```diff
@@ -355,13 +356,41 @@ public void testKeyIsNotAdded_when_addWhen_condition_is_false() {
         assertThat(editedRecords.get(0).getData().containsKey("newMessage"), is(false));
     }
 
+    @Test
+    public void testMetadataKeySet() {
```
cmanning09, line 360: This test name should describe the test. Please be more explicit. 

Please add other tests which test `overwrite_if_key_exists`, `format`, `value` and any failure scenario that comes with using these options with the `metadata_key`. I know we are leveraging the some of the same code under the hood but our unit tests should harden the contract that `metadata_key` works with these other configuration options. This will help prevent future developers breaking support for these features unknowingly and bugs being released. These unit tests will catch any breaking changes.

data-prepper-plugins/mutate-event-processors/README.md
kkondaka: It does not support Json pointer. I have updated the documentation.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorConfig.java
kkondaka: May be we can add that if needed. For now, this is not supported.

graytaylor0: Seems fine either way

graytaylor0 (dismissed):

data-prepper-plugins/mutate-event-processors/src/test/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorTests.java
kkondaka: @cmanning09  I am thinking if we should do something like this for configuration to avoid duplicate config options
```
processor:
    - add_entries:
        entries:
        - key: "newMessage"
           format: "new ${message}"
           overwrite_if_key_exists: true
           add_to_metadata: true
```
This will keep the existing functionality/options as is but adds a new boolean flag that will add to metadata instead of the event. I can't think of a case where same thing needs to be added to both event and metadata

graytaylor0 (approved):

cmanning09 (approved):

PR #2723, opened 2023-05-20
Instrument metrics in LeaseBasedSourceCoordinator

graytaylor0:
### Description
Instruments a number of metrics to give insight into progress tracking and failures in the source coordinator
 
### Issues Resolved
Related to #2412 
 
### Check List

### DIFF
--- a/data-prepper-core/src/main/java/org/opensearch/dataprepper/sourcecoordination/LeaseBasedSourceCoordinator.java
+++ b/data-prepper-core/src/main/java/org/opensearch/dataprepper/sourcecoordination/LeaseBasedSourceCoordinator.java
@@ -6,6 +6,8 @@
 package org.opensearch.dataprepper.sourcecoordination;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.micrometer.core.instrument.Counter;
+import org.opensearch.dataprepper.metrics.PluginMetrics;
 import org.opensearch.dataprepper.model.source.SourceCoordinationStore;
 import org.opensearch.dataprepper.model.source.coordinator.PartitionIdentifier;
 import org.opensearch.dataprepper.model.source.coordinator.SourceCoordinator;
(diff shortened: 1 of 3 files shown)


kkondaka (approved):
Looks good.

data-prepper-core/src/main/java/org/opensearch/dataprepper/sourcecoordination/LeaseBasedSourceCoordinator.java
```diff
@@ -170,8 +226,23 @@ public void closePartition(final String partitionKey, final Duration reopenAfter
             itemToUpdate.setReOpenAt(Instant.now().plus(reopenAfter));
         }
 
+        try {
+            sourceCoordinationStore.tryUpdateSourcePartitionItem(itemToUpdate);
+        } catch (final PartitionUpdateException e) {
+            if (SourcePartitionStatus.COMPLETED.equals(itemToUpdate.getSourcePartitionStatus())) {
+                completePartitionUpdateErrorCounter.increment();
+            } else {
+                closePartitionUpdateErrorCounter.increment();
```
cmanning09, line 235: This logic won't extend well if we add a new status that is supported in this method. SaveProgressState does not have a metric for each partition status type. Do we need to differentiate between Closed and Completed? I would recommend creating a dimension that keyed off of status if we need to.

graytaylor0: The only updates are saveState, complete, and close. I chose to give granularity on which action failed to update by separating to three different metrics. I am fine with adding a single metric for `partitionUpdateErrors` with a dimension that shows the action, but how does this work with micrometer

cmanning09 (approved):
Don't worry about changing if you don't think its worth it. I am just advocating for code that can be easily extended beyond what we can plan and know of as possibilities right now. It's a simple adjustment that will make it easier for future engineers to inherently get metrics but its not required.

PR #2770, opened 2023-05-26
OpenSearch initialization fix to retry after any exception

### LINKED ISSUE #2769: [BUG] OpenSearch Sink should continue to try to initialize even after non-retryable exceptions
**Describe the bug**
Currently OpenSearch sink checks to see if the exception is retry-able or not to retry  the initialization. If the exception is considered non-retryable, initialization is not retried. This is not ideal because some of the non-retry-able errors can be fixed by fixing the permissions/OpenSearch config etc.

**To Reproduce**
If Trust relationship is not properly set with OS domain, then STS assume role errors occur which makes the OpenSearch Sink initialization non-retry-able.

**Expected behavior**
Let the OpenSearch sink init code to retry in all exception cases.

**Screenshots**

 - OS: [e.g. Ubuntu 20.04 LTS]

**Additional context**

kkondaka:
### Description
OpenSearch initialization fix to retry after any exception.

Resolves #2769 
 
### Issues Resolved
Resolves #2769 
 
### Check List


dlvenable (approved):

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -147,12 +147,11 @@ public void doInitialize() {
         LOG.error("Failed to initialize OpenSearch sink due to a configuration error.", e);
         this.shutdown();
```
engechas, line 148: There's a case here where boot can fail even though the issue is resolvable by the customer.

In the case of managed indices, if an index exists with the same name as the intended alias, the initialization will fail and shut down the pipeline. The action to fix this is to delete the index with same name as the intended alias. Since this is a configuration issue external to DataPrepper, I don't think we should shut down in this scenario.

What do you think?

cmanning09: @engechas, would that result in an `IllegalArgumentException`?

engechas: No, it manifested in a `InvalidPluginConfigurationException`:
```
2023-05-30T20:04:53.575 [raw-pipeline-sink-worker-6-thread-1] ERROR org.opensearch.dataprepper.plugins.sink.opensearch.OpenSearchSink - Failed to initialize OpenSearch sink due to a configuration error.
org.opensearch.dataprepper.model.plugin.InvalidPluginConfigurationException: An index exists with the same name as the reserved index alias name [otel-v1-apm-span], please delete or migrate the existing index
	at org.opensearch.dataprepper.plugins.sink.opensearch.index.AbstractIndexManager.checkAndCreateIndex(AbstractIndexManager.java:271) ~[opensearch-2.2.1.jar:?]
	at org.opensearch.dataprepper.plugins.sink.opensearch.index.AbstractIndexManager.setupIndex(AbstractIndexManager.java:204) ~[opensearch-2.2.1.jar:?]
	at org.opensearch.dataprepper.plugins.sink.opensearch.OpenSearchSink.doInitializeInternal(OpenSearchSink.java:175) ~[opensearch-2.2.1.jar:?]
	at org.opensearch.dataprepper.plugins.sink.opensearch.OpenSearchSink.doInitialize(OpenSearchSink.java:140) ~[opensearch-2.2.1.jar:?]
	at org.opensearch.dataprepper.model.sink.AbstractSink.initialize(AbstractSink.java:48) ~[data-prepper-api-2.2.1.jar:?]
	at org.opensearch.dataprepper.pipeline.Pipeline.isReady(Pipeline.java:195) ~[data-prepper-core-2.2.1.jar:?]
	at org.opensearch.dataprepper.pipeline.Pipeline.lambda$execute$2(Pipeline.java:243) ~[data-prepper-core-2.2.1.jar:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) ~[?:?]
	at java.lang.Thread.run(Thread.java:829) [?:?]
```

Quickly followed by a shutdown:
```
2023-05-30T20:04:53.587 [raw-pipeline-sink-worker-6-thread-1] INFO  org.opensearch.dataprepper.pipeline.Pipeline - Pipeline [raw-pipeline] - Received shutdown signal with processor shutdown timeout PT5M and sink shutdown timeout PT30S. Initiating the shutdown process
```

engechas: I am not sure what all cases result in `InvalidPluginConfigurationException`. I would argue that the above one should just be retried rather than shutdown, but there may be other paths to this exception that are truly non-retryable

engechas (approved):

PR #2821, opened 2023-06-03
Implement opensearch index partition creation supplier and PitWorker without processing indices

graytaylor0:
### Description
Implements the `OpenSearchIndexPartitionSupplier`, which will list all indices of the source cluster and filter them by the include and exclude regex patterns.

Also starts implementation for PitWorker to use the scheduling configuration with source coordinator without processing the object. When calling `getNextPartition` returns empty, the PitWorker will back off and retry in a fixed 30 seconds. The code added for PitWorker here would be essentially the same for ScrollWorker (the interactions with source coordinator will be very similar.

`OpenSearchService` now uses a scheduled executor service instead of spawning new threads directly. This allows scheduling to start at the `start_time`. The `stop` method will let the `SearchWorker` future know to not grab another partition, and will get 30 seconds to complete processing of the index before the executor service shuts it down completely. This means that if 30 seconds is not enough time to finish processing the current partition, duplicate data processing will occur (unless there is a state that can track where to pick up from, at which point the `SearchWorker` could saveState and give up the partition gracefully)

Tested and expected partitions are created and "processed" successfully

 
### Issues Resolved
Related to #1985 
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/OpenSearchService.java
+++ b/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/OpenSearchService.java
@@ -79,7 +79,7 @@ public class OpenSearchService {
         final long waitTimeBeforeStartMillis = startTime.toEpochMilli() - Instant.now().toEpochMilli() < 0 ? 0L :
                 startTime.toEpochMilli() - Instant.now().toEpochMilli();
 
-        LOG.info("The opensearch source will start processing data at {}. It is currently {}", startTime, Instant.now().toString());
+        LOG.info("The opensearch source will start processing data at {}. It is currently {}", startTime, Instant.now());
 
         searchWorkerFuture = scheduledExecutorService.schedule(() -> searchWorker.run(), waitTimeBeforeStartMillis, TimeUnit.MILLISECONDS);
     }
--- a/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/OpenSearchService.java
+++ b/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/OpenSearchService.java
@@ -8,44 +8,66 @@ import org.opensearch.dataprepper.model.buffer.Buffer;
 import org.opensearch.dataprepper.model.event.Event;
 import org.opensearch.dataprepper.model.record.Record;
 import org.opensearch.dataprepper.model.source.coordinator.SourceCoordinator;
+import org.opensearch.dataprepper.plugins.source.opensearch.worker.OpenSearchIndexPartitionCreationSupplier;
 import org.opensearch.dataprepper.plugins.source.opensearch.worker.PitWorker;
 import org.opensearch.dataprepper.plugins.source.opensearch.worker.ScrollWorker;
+import org.opensearch.dataprepper.plugins.source.opensearch.worker.SearchWorker;
+import org.opensearch.dataprepper.plugins.source.opensearch.worker.client.ClusterClientFactory;
 import org.opensearch.dataprepper.plugins.source.opensearch.worker.client.SearchAccessor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
 public class OpenSearchService {
+
+    private static final Logger LOG = LoggerFactory.getLogger(OpenSearchService.class);
+
+    static final Duration EXECUTOR_SERVICE_SHUTDOWN_TIMEOUT = Duration.ofSeconds(30);
 
     private final SearchAccessor searchAccessor;
     private final OpenSearchSourceConfiguration openSearchSourceConfiguration;
     private final SourceCoordinator<OpenSearchIndexProgressState> sourceCoordinator;
     private final Buffer<Record<Event>> buffer;
+    private final OpenSearchIndexPartitionCreationSupplier openSearchIndexPartitionCreationSupplier;
+    private final ScheduledExecutorService scheduledExecutorService;
 
-    private Thread searchWorkerThread;
+    private SearchWorker searchWorker;
+    private ScheduledFuture<?> searchWorkerFuture;
 
     public static OpenSearchService createOpenSearchService(final SearchAccessor searchAccessor,
                                                             final SourceCoordinator<OpenSearchIndexProgressState> sourceCoordinator,
                                                             final OpenSearchSourceConfiguration openSearchSourceConfiguration,
                                                             final Buffer<Record<Event>> buffer) {
-        return new OpenSearchService(searchAccessor, sourceCoordinator, openSearchSourceConfiguration, buffer);
+        return new OpenSearchService(searchAccessor, sourceCoordinator, openSearchSourceConfiguration, buffer, Executors.newSingleThreadScheduledExecutor());
     }
 
     private OpenSearchService(final SearchAccessor searchAccessor,
                              final SourceCoordinator<OpenSearchIndexProgressState> sourceCoordinator,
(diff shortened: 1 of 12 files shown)


data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/OpenSearchService.java
```diff
@@ -8,55 +8,96 @@
...
                         String.format("Search context type must be POINT_IN_TIME or SCROLL, type %s was given instead",
                                 searchAccessor.getSearchContextType()));
         }
 
-        searchWorkerThread.start();
+        final Instant startTime = openSearchSourceConfiguration.getSchedulingParameterConfiguration().getStartTime();
+        final long waitTimeBeforeStartMillis = startTime.toEpochMilli() - Instant.now().toEpochMilli() < 0 ? 0L :
+                startTime.toEpochMilli() - Instant.now().toEpochMilli();
+
+        LOG.info("The opensearch source will start processing data at {}. It is currently {}", startTime, Instant.now().toString());
```
cmanning09, line 82: Is the `.toString()` redundant?

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
```diff
@@ -0,0 +1,116 @@
...
+                .map(indexRecord -> PartitionIdentifier.builder().withPartitionKey(indexRecord.index()).build())
+                .collect(Collectors.toList());
+    }
+
+    private boolean isIndexIncludedInOneOfTheIncludePatternsAndNotExcludedInAnExcludePattern(final IndicesRecord indicesRecord) {
+        if (Objects.isNull(indicesRecord.index())) {
+            return false;
+        }
+
+        final IndexParametersConfiguration indexParametersConfiguration = openSearchSourceConfiguration.getIndexParametersConfiguration();
```
cmanning09, line 81: Take or leave it: This and other configurations values can be saved as a class variable instead of creating a new local variable for every index filter. These values are always the same.

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
```diff
@@ -0,0 +1,116 @@
...
+
+        if (Objects.isNull(openSearchSourceConfiguration.getIndexParametersConfiguration())) {
```
cmanning09, line 83: Can be simplified to: `if (Objects.isNull(indexParametersConfiguration)) {`

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
```diff
@@ -0,0 +1,116 @@
...
+            return Collections.emptyList();
+        }
+
+        return indicesResponse.valueBody().stream()
+                .filter(this::isIndexIncludedInOneOfTheIncludePatternsAndNotExcludedInAnExcludePattern)
+                .map(indexRecord -> PartitionIdentifier.builder().withPartitionKey(indexRecord.index()).build())
+                .collect(Collectors.toList());
+    }
+
+    private boolean isIndexIncludedInOneOfTheIncludePatternsAndNotExcludedInAnExcludePattern(final IndicesRecord indicesRecord) {
```
cmanning09, line 76: Take or leave it: A simpler name and one that would extend well if we added other filtering options could be: 

- `shouldIndexBeProcessed`
- `isIndexSelected`

The current method name is very long and a little too specific. I am open to other ideas as well.

cmanning09 (approved):
It looks fundamentally correct to me. I am offering a few suggestions to improve the code. Take them or leave them.

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
graytaylor0: I like `shouldIndexBeProcessed`

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
```diff
@@ -0,0 +1,116 @@
...
+        boolean matchesExcludePattern = false;
+
+        for (final OpenSearchIndex index : excludedIndices) {
+            final Matcher matcher = index.getIndexNamePattern().matcher(indicesRecord.index());
+
+            if (matcher.matches()) {
+                matchesExcludePattern = true;
+                break;
+            }
+        }
```
cmanning09, line 111: You can eliminate code duplication, variable re-assignment and breaks by pull this into a separate function:

```
    pubic boolean doesIndexMatchPattern(final List<OpenSearchIndex> indices, final IndicesRecord indicesRecord) {
        for (final OpenSearchIndex index : excludedIndices) {
            final Matcher matcher = index.getIndexNamePattern().matcher(indicesRecord.index());

            if (matcher.matches()) {
                return true;
            }
        }
        return false;
    }
```

Then you can do something simple like:

```
final boolean matchesIncludedPattern = includedIndices.isEmpty() ? true : doesIndexMatchPattern(includedIndices, indicesRecord);
final boolean matchesExcludePattern = doesIndexMatchPattern(excludedIndices, indicesRecord);
return matchesIncludedPattern && !matchesExcludePattern;
```

cmanning09 (approved):

data-prepper-plugins/opensearch-source/src/test/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/OpenSearchIndexPartitionCreationSupplierTest.java
```diff
@@ -0,0 +1,153 @@
...
+        indicesRecords.add(neitherIncludedOrExcluded);
+
+        when(indicesResponse.valueBody()).thenReturn(indicesRecords);
+
+        when(openSearchCatClient.indices()).thenReturn(indicesResponse);
+        when(openSearchClient.cat()).thenReturn(openSearchCatClient);
+
+        final List<PartitionIdentifier> partitionIdentifierList = createObjectUnderTest().apply(Collections.emptyMap());
+
+        assertThat(partitionIdentifierList, notNullValue());
```
asifsmohammed, line 146: nit: I think we should verify number of partitions here.

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/ElasticsearchAccessor.java
```diff
@@ -56,4 +56,9 @@ public SearchScrollResponse searchWithScroll(SearchScrollRequest searchScrollReq
     public void deleteScroll(DeleteScrollRequest deleteScrollRequest) {
         //todo: implement
     }
+
+    @Override
+    public Object getClient() {
```
asifsmohammed, line 61: Why do we implement `ClusterClientFactory` here if we are returning null?

asifsmohammed (approved):

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/ElasticsearchAccessor.java
graytaylor0: The ElasticSearch accessor is not implemented yet

PR #2839, opened 2023-06-06
Create and delete point in time for processing an index

graytaylor0:
### Description
Create and delete point in time for an index to process it

* Note: The deletePit request does not work when `enable_sigv4` is set to true. I have not been able to debug, but it works properly when `enable_sigv4` is false. I will create a separate issue to work on this bug in the future.
 
### Issues Resolved
related to #1985 
 
### Check List


kkondaka:
Please comment out the failing tests, if they are failing because of sigv4

graytaylor0:
> Please comment out the failing tests, if they are failing because of sigv4

A couple were failing from an unnecessary mock. I fixed those. No unit tests are failing from sigv4 issues. Manual testing was done to see the issue.

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/OpenSearchAccessor.java
```diff
@@ -5,20 +5,37 @@
...
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Objects;
 
 public class OpenSearchAccessor implements SearchAccessor, ClusterClientFactory {
 
+    private static final Logger LOG = LoggerFactory.getLogger(OpenSearchAccessor.class);
+
+    static final String PIT_RESOURCE_LIMIT_ERROR_TYPE = "rejected_execution_exception";
```
engechas, line 37: This error type seems more generic than the PIT logic. Are there other cases where we can get a RejectedExecutionException unrelated to PIT?

engechas (approved):

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/OpenSearchAccessor.java
graytaylor0: Probably but what is the alternative? Check the message contains something unique to the PIT message? Even for other exception we would log it and then give up the index and wait to start processing again, which is ok

oeyh (approved):

PR #2849, opened 2023-06-08
Os source buffer backoff retry

graytaylor0:
### Description
This change uses the `BufferAccumulator` class from the s3 source to write and flush documents to the buffer with backoff and retry.

The `records_to_accumulate` is currently the same as the `batch_size` (pagination size) for searching.
 
Waiting on #2847 before rebasing and pushing non-draft PR

### Issues Resolved
Related to #1985 
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
+++ b/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
@@ -10,6 +10,7 @@ import org.opensearch.dataprepper.model.record.Record;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.concurrent.NotThreadSafe;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -28,6 +29,8 @@ import java.util.concurrent.TimeoutException;
  *
  * @param <T> Type of record to accumulate
  */
+// todo: Move this class to its own @DataPrepperPlugin so it can be reused between the s3 and opensearch source, and other sources that may need it (https://github.com/opensearch-project/data-prepper/issues/2855)
+@NotThreadSafe
 public class BufferAccumulator<T extends Record<?>> {
     private static final Logger LOG = LoggerFactory.getLogger(BufferAccumulator.class);
 
--- a/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
+++ b/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.dataprepper.plugins.source.opensearch.worker;
+
+import org.opensearch.dataprepper.model.buffer.Buffer;
+import org.opensearch.dataprepper.model.record.Record;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Accumulates {@link Record} objects before placing them into a Data Prepper
+ * {@link Buffer}. This class is not thread-safe and should only be used by one
+ * thread at a time.
+ *
+ * @param <T> Type of record to accumulate
+ */
+public class BufferAccumulator<T extends Record<?>> {
+    private static final Logger LOG = LoggerFactory.getLogger(BufferAccumulator.class);
+
+    private static final int MAX_FLUSH_RETRIES_ON_IO_EXCEPTION = Integer.MAX_VALUE;
+    private static final Duration INITIAL_FLUSH_RETRY_DELAY_ON_IO_EXCEPTION = Duration.ofSeconds(5);
+
+    private final Buffer<T> buffer;
+    private final int numberOfRecordsToAccumulate;
+    private final int bufferTimeoutMillis;
+    private int totalWritten = 0;
+
+    private final Collection<T> recordsAccumulated;
+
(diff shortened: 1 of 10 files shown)


data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
```diff
@@ -0,0 +1,130 @@
+/*
```
engechas, line 1: Could we pull out the original BufferAccumulator into a place where multiple plugins can consume it? It'd be better to maintain the logic in a singular place

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/PitWorker.java
```diff
@@ -138,9 +134,14 @@ private void processIndex(final SourcePartition<OpenSearchIndexProgressState> op
...
                         .build());
-                buffer.writeAll(searchPointInTimeResults.getDocuments().stream().map(Record::new).collect(Collectors.toList()), BUFFER_TIMEOUT_MILLIS);
-            } catch (final TimeoutException e) {
-                // todo: implement backoff and retry, can reuse buffer accumulator code from the s3 source
+
+                searchPointInTimeResults.getDocuments().stream().map(Record::new).forEach(record -> {
+                    try {
+                        bufferAccumulator.add(record);
+                    } catch (Exception e) {
+                        LOG.error("Failed writing OpenSearch documents to buffer due to: {}", e.getMessage());
```
engechas, line 142: Should we emit a metric here and/or log the document that failed? The user won't be able to take action without knowing which documents are failing

graytaylor0: Agreed but can we do that in the future? I’ll create an issue for it

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
graytaylor0: Agreed but can we do that in the future? I’ll create an issue for it

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/PitWorker.java
graytaylor0: Will the message not contain the failed record info? Otherwise how do I get it. Log it in the `BufferAccumulator`?

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Accumulates {@link Record} objects before placing them into a Data Prepper
+ * {@link Buffer}. This class is not thread-safe and should only be used by one
+ * thread at a time.
+ *
+ * @param <T> Type of record to accumulate
+ */
+public class BufferAccumulator<T extends Record<?>> {
```
chenqi0805, line 31: You can use javax annotation @NotThreadSafe

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
graytaylor0: Here is the followup PR after this one is merged (https://github.com/opensearch-project/data-prepper/pull/2857)

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
```diff
@@ -0,0 +1,130 @@
...
+    private static final Duration INITIAL_FLUSH_RETRY_DELAY_ON_IO_EXCEPTION = Duration.ofSeconds(5);
+
+    private final Buffer<T> buffer;
+    private final int numberOfRecordsToAccumulate;
+    private final int bufferTimeoutMillis;
+    private int totalWritten = 0;
+
+    private final Collection<T> recordsAccumulated;
+
+    private BufferAccumulator(final Buffer<T> buffer, final int numberOfRecordsToAccumulate, final Duration bufferTimeout) {
```
kkondaka, line 47: There are other Buffer Accumulator implementations. Do you think they can all be combined to one implementation?

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
```diff
@@ -0,0 +1,130 @@
...
+            flushWithBackoff();
+        }
+    }
+
+    private boolean flushWithBackoff() throws Exception{
+        final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+        long nextDelay = INITIAL_FLUSH_RETRY_DELAY_ON_IO_EXCEPTION.toMillis();
+        boolean flushedSuccessfully;
+
+        for (int retryCount = 0; retryCount < MAX_FLUSH_RETRIES_ON_IO_EXCEPTION; retryCount++) {
```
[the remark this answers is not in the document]
graytaylor0: For now I don’t think there’s any value in it. But it’s always an option in the future

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/BufferAccumulator.java
graytaylor0: Yes I have a follow-up to full this out into a reusable class

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/PitWorker.java
engechas: Looking back at the code, I think my comment was wrong. BufferAccumulator::add adds to the local collection and flushes if needed. There might be a case where adding to the local collection fails, but this exception should almost always be because the flush failed. I thought we were silently dropping the record if this path occurred, but that's not the case.

Having the doc/index in the log is useful to denote progress nonetheless

engechas (approved):

kkondaka (approved):

PR #2854, opened 2023-06-10
(J)Ruby Processor — Skeleton + Run code from config implementation

finnroblin:
### Description
Adds a Ruby processor that can execute custom Ruby code. This PR adds the code skeleton, config options, and an implementation to run code from a config string. 

### Issues Resolved
Works towards addressing Issue #2388 

### Check List


data-prepper-plugins/ruby-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessorIT.java
```diff
@@ -0,0 +1,63 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.opensearch.dataprepper.test.helper.ReflectivelySetField.setField;
+
+public class RubyProcessorIT {
+    private static final int NUMBER_EVENTS_TO_TEST = 100;
+    private static final String PLUGIN_NAME = "ruby";
+    private static final String TEST_PIPELINE_NAME = "ruby_processor_test";
+    static String CODE_STRING = "require 'java'\n" +
```
asuresh8: Copied from slack:


Having to require the customer have the to_java  is a bad customer experience. Ideally the Ruby code block should just be pure ruby without any jruby type conversions in there, and the Ruby processor should handle the translation between Ruby and Java behind the scenes.

data-prepper-plugins/ruby-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessor.java
```diff
@@ -57,20 +62,52 @@ public Collection<Record<Event>> doExecute(final Collection<Record<Event>> recor
...
+            }
+            else {
+                if (!config.isIgnoreException()) {
+                    throw new RuntimeException(); // todo: should we tag instead of breaking pipeline?
+                }
             }
         }
     }
 
+    private String cleanEventAPICalls(String script) {
```
asuresh8, line 89: I like the idea of this method, but I think the implementation can be improved. Does the Ruby code the customer supplies even need to specify `<X>.class`? Could it instead be `event.get("key")` converts to `event.get("key", java.lang.Object.java_class)`?

data-prepper-plugins/ruby-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessorIT.java
```diff
@@ -131,4 +135,163 @@ void when_dotClassMethodCalledAnywhereBesidesEventGetCall_then_remainsUnchanged(
...
+//        CODE_STRING = "out_arr = event.get('message', Object.class)\n" +
+//                "event.put('sum', out_arr.at(0) + out_arr.at(1))\n";
+
+//        CODE_STRING = "out_arr = event.get('message')\n" +
+//                "event.put('sum', out_arr.at(0) + out_arr.at(1))\n";
+
+//        CODE_STRING = "out_arr = event.getList('message')\n" +
+//                "event.put('sum', out_arr.at(0) + out_arr.at(1))\n";
+
+        CODE_STRING = "out_arr = event.getList('message')\n" + // also works with get()
```
asuresh8, line 173: This example looks good to me!

data-prepper-plugins/ruby-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessorIT.java
```diff
@@ -294,4 +275,42 @@ void when_listCreatedInRuby_then_listRetrievableInJava() {
...
+        try {
+            setField(RubyProcessorConfig.class, RubyProcessorIT.this.rubyProcessorConfig, "initCode", initString);
+        } catch (NoSuchFieldException | IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
+
+        PluginMetrics pluginMetrics = PluginMetrics.fromNames(PLUGIN_NAME, TEST_PIPELINE_NAME);
+
+        rubyProcessor = new RubyProcessor( pluginMetrics, rubyProcessorConfig);
+
```
asuresh8, line 298: Nit: remove extra line

config/data-prepper-config.yaml
```diff
@@ -0,0 +1 @@
+ssl: false
```
kkondaka, line 1: Do we need this change?

data-prepper-plugins/ruby-processor/.gitignore
```diff
@@ -0,0 +1,42 @@
+.gradle
```
kkondaka, line 1: Do we need this file?

finnroblin: Nope, sorry about that — will remove on next commit.

data-prepper-plugins/ruby-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessor.java
```diff
@@ -23,6 +23,8 @@ public class RubyProcessor extends AbstractProcessor<Record<Event>, Record<Event
     private ScriptingContainer container;
     private String script;
 
+    private Boolean runningCodeFromFile = false;
```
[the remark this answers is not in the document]
finnroblin: Some plugins like [csv processor](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/csv-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/csv/CsvProcessorConfig.java) use the object. Others like [Otel source ](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/otel-logs-source/src/main/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSourceConfig.java) use primitive. I'm fine with changing to a primitive.

data-prepper-plugins/ruby-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessor.java
```diff
@@ -37,37 +54,104 @@ public RubyProcessor(final PluginMetrics pluginMetrics, final RubyProcessorConfi
...
+
+        if (this.shouldThrowExceptionSinceNoProcessFound) {
+            LOG.error("must implement interface described in README when using Ruby code from file. Exiting pipeline.");
+            throw new RuntimeException("Ruby Processor file bad implementation");
+        }
     }
 
-    private void runInitCodeFromFileAndDefineProcessMethod(final String codeFilePath) {
-        // check if init code exists
+    private boolean assertThatRubyProcessorInterfaceImplementedAndReturnInitMethodDefined(SeekableInputStream fileStreamForTestingRegex)
```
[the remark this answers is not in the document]
finnroblin: Sounds good, I renamed it to `assertThatInterfaceImplementedInRubyFile` to clarify that it refers to custom code in a Ruby file.

data-prepper-plugins/ruby-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessorE2ETest.java
```diff
@@ -0,0 +1,4 @@
+package org.opensearch.dataprepper.plugins.processor.ruby;
+
+public class RubyProcessorE2ETest {
```
finnroblin, line 3: I did not mean to make this a new file and thought I deleted it from the commit — I will remove this on next revision.

asuresh8 (commented):
Looks good!

data-prepper-plugins/ruby-processor/runBundlerToUpdateGems.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+jruby_jar_path=$1
+echo "Starting runBundlerToUpdateGems.sh...."
+# Check if JRuby jar path is provided
+if [[ -z "$jruby_jar_path" ]]; then
+echo "Please provide the JRuby jar path as an argument."
```
asuresh8, line 7: indent this block

asuresh8: Same for other if statements

asuresh8 (commented):
Generally looks good. Only a few small comments

data-prepper-plugins/ruby-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessor.java
```diff
@@ -0,0 +1,201 @@
...
+                throw new RuntimeException(format("Exception while processing Ruby code on Events: [%s]",events),
+                        exception
+                );
+            } else {
+                retryEventsOneByOneAndApplyTags(events);
+            }
+        }
+    }
+
+    private void retryEventsOneByOneAndApplyTags(List<Event> events) {
```
graytaylor0, line 171: Nit: `final` on these variables

data-prepper-plugins/ruby-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessor.java
```diff
@@ -0,0 +1,201 @@
...
+        script = "events.each { |event| \n"
+                + this.codeToInject +
+                "}";
+
+        try {
+            container.runScriptlet(script);
+        } catch (Exception exception) {
+            LOG.error("Exception while processing Ruby code on Events: [{}]",events, exception);
+            if (!config.isIgnoreException()) {
+                throw new RuntimeException(format("Exception while processing Ruby code on Events: [%s]",events),
```
graytaylor0, line 162: Do we need to log all the Events? Would the Events be dropped if this exception is thrown? I think the preferred behavior would be to log the error and then return the Events so that they are sent through the pipeline like they were before they entered the ruby processor

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/event/Event.java
```diff
@@ -53,6 +53,30 @@ public interface Event extends Serializable {
...
+     * Retrieves the given key, of generic type Object, from the Event as a List.
+     * Used by the Ruby processor to
+     * provide an Event API from within Ruby.
+     *
+     * @param key the value to retrieve from
+     * @param <T> The type
+     * @return {@literal List<T>} a list of Object elements
+     * @since 2.4
+     */
+    <T> List<T> getList(String key);
```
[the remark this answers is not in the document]
asuresh8: I suggested Finn go this route to avoid requiring the second parameter in the ruby code. The logstash ruby plugin doesn't require the second parameter, so this helps acheive parity with logstash.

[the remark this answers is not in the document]
finnroblin: Thanks for the feedback and coupling the Ruby processor and Event is also a concern of mine. 
Another issue with keeping the second parameter is that any Ruby code needs to do some complicated footwork to send the right class literal. So `get(key, Object.class)` would require something along the lines of `require 'java' \n get(key, java.lang.Object.class)` within Ruby. 
An alternative implementation that I spent some time thinking about was creating a specific RubyEvent for this purpose. Going down this route would require some type casting within the processor. If I recall it would also require some changes to Event like making some methods `protected` instead of `private`. 

What do you think about this solution, or keeping the original modification to `get` and `getList`?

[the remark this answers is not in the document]
finnroblin: A third alternative is to have users write `get(field)` in their code and expand these calls to `get(field, java_object_class_literal)` before they are sent into the JRuby environment. While normally these string manipulation are fragile, the fact that there is only one expansion makes me more comfortable suggesting this.

data-prepper-plugins/ruby-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/ruby/RubyProcessor.java
```diff
@@ -102,9 +185,11 @@ private void injectAndProcessEvents(List<Event> events) {
         try {
             container.runScriptlet(script);
         } catch (Exception exception) {
-            LOG.error("Exception processing Ruby code", exception);
+            LOG.error("Exception while processing Ruby code on Events: [{}]",events, exception);
```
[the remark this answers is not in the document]
asifsmohammed: We should not log the event data here. please remove it. Maybe you can use SENSITIVE tag in log here.

data-prepper-plugins/ruby-processor/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+def process(event) # code must implement this interface. See [[Ruby interface]] below.
+        event.put("message_translation", 
+            I18n.t(event.get("message"))
+        )
+end
+```
+## Configuration options
+
+code — Ruby code that will be run on each Event.
```
asifsmohammed, line 46: Nit: options can be in a code block so it's easier to read. 
`code` -

asifsmohammed (commented):
LGTM. Address the comments on logging.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/event/Event.java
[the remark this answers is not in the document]
dlvenable: You can create a `RubyEvent` in the Ruby Processor project. 


```
class RubyEvent implements Event {
  RubyEvent(Event delegate) {
    this.delegate = delegate;
  }

  <T> T get(String key) {
    // your implementation
  }

  // follow this pattern for any existing methods that don't need different behavior.
  <T> T get(String key, Class<T> clazz) {
    return delegate.get(key, clazz);
  }
```

data-prepper-plugins/ruby-processor/rubygems/rubygems.jar
dlvenable, line 1: This file should not be checked in.

We don't have a version information here, license information, etc. This is a binary file that we don't have any insight into.

It also adds additional size to the overall repo.

The best solution is to use this jar as Gradle dependency. An alternative is to download the file using a Gradle task. In this way, we can know the exact version used and the source.

dlvenable (changes requested):

PR #2866, opened 2023-06-13
Use queue visibility timeout setting when no visibility_timeout is set in the pipeline config, lookup visibility timeout for acknowledgments

### LINKED ISSUE #2484: S3 Source SQS queue visibility timeout is overridden when it is not configured
The s3 source visibility timeout defaults to 30 seconds in both SQS and Data Prepper. However, if the visibility timeout is changed on the SQS queue, and not configured in the S3 Source, the default of 30 seconds is still overridden in the call to pull the message from SQS. This is confusing and can lead to unexpected timeouts if processing larger messages from the queue.

If no `visibility_timeout` is set on the sqs queue in the data prepper pipeline configuration, the SQS queue's visibility timeout should be used, rather than the default visibility timeout set by data prepper.

**Additional context**

graytaylor0:
### Description
Removes the default `visibility_timeout` of 30 seconds from Data Prepper, and will pass null to the `ReceiveMessageRequest` for the visibility_timeout.

When `acknowledgments` are enabled and visibility_timeout is not set, a `GetQueueAttributes` API call is made to get the queue's visibility timeout. This value is then set in the `SqsOptions`, and will be passed to the `ReceiveMessageRequest` and used for acknowledgments timeout.  
 
### Issues Resolved
Resolves #2484 
 
### Check List

### DIFF
--- a/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/SqsWorker.java
+++ b/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/SqsWorker.java
@@ -10,13 +10,13 @@ import com.linecorp.armeria.client.retry.Backoff;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.Timer;
 import org.opensearch.dataprepper.metrics.PluginMetrics;
+import org.opensearch.dataprepper.model.acknowledgements.AcknowledgementSet;
+import org.opensearch.dataprepper.model.acknowledgements.AcknowledgementSetManager;
 import org.opensearch.dataprepper.plugins.source.configuration.OnErrorOption;
 import org.opensearch.dataprepper.plugins.source.configuration.SqsOptions;
 import org.opensearch.dataprepper.plugins.source.exception.SqsRetriesExhaustedException;
 import org.opensearch.dataprepper.plugins.source.filter.ObjectCreatedFilter;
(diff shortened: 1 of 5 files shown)


data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/SqsWorker.java
```diff
@@ -168,7 +169,7 @@ private ReceiveMessageRequest createReceiveMessageRequest() {
         return ReceiveMessageRequest.builder()
                 .queueUrl(sqsOptions.getSqsUrl())
                 .maxNumberOfMessages(sqsOptions.getMaximumMessages())
-                .visibilityTimeout((int) sqsOptions.getVisibilityTimeout().getSeconds())
+                .visibilityTimeout(Objects.nonNull(sqsOptions.getVisibilityTimeout()) ? (int) sqsOptions.getVisibilityTimeout().getSeconds() : null)
```
asifsmohammed, line 172: Looks good. Can we pass null here?

graytaylor0: I looked into the code for the `ReceiveMessageRequest` builder and it looks like it should be fine as null. But we still should do some manual testing without the visibility timeout to verify that this uses the queue visibility timeout when null is passed

PR #2873, opened 2023-06-14
Implement NoSearchContextWorker to search with search_after and not use pit or scroll, allow override with search_context_type parameter

graytaylor0:
### Description
* Implements `NoSearchContextWorker`, which will process indices using sort and search_after without creating search contexts via scroll or point in time
* Adds a new parameter to `search_options`, the `search_context_type`. 

The `search_context_type` can be set to `none`, `point_in_time`, or `scroll`. By default, point in time will be used for versions that can support it, and scroll will be used for versions that don't support point in time. 

However, there is a consideration to make the default just be `none`, and just use the version lookup to validate that point in time can be used if the `search_context_type` is set to `point_in_time`.
 
### Issues Resolved
Related to #1985 
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/configuration/SearchConfiguration.java
+++ b/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/configuration/SearchConfiguration.java
@@ -10,14 +10,20 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import jakarta.validation.constraints.AssertTrue;
+import org.opensearch.dataprepper.plugins.source.opensearch.worker.client.model.SearchContextType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import java.util.Map;
 
 public class SearchConfiguration {
 
     private static final ObjectMapper objectMapper = new ObjectMapper();
     private static final Logger LOG = LoggerFactory.getLogger(SearchConfiguration.class);
+
+    // TODO: Should we default this to NONE and remove the version lookup to determine scroll or point-in-time as the default behavior?
+    @JsonProperty("search_context_type")
+    private SearchContextType searchContextType;
 
     @JsonProperty("batch_size")
     private Integer batchSize = 1000;
@@ -28,6 +34,9 @@ public class SearchConfiguration {
     @JsonIgnore
     private Map<String, Object> queryMap;
 
+    public SearchContextType getSearchContextType() {
+        return searchContextType;
+    }
 
     public Integer getBatchSize() {
         return batchSize;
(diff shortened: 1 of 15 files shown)


data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/NoSearchContextWorker.java
```diff
@@ -0,0 +1,150 @@
...
+    private OpenSearchIndexProgressState initializeProgressState() {
+        return new OpenSearchIndexProgressState();
+    }
+
+    private List<String> getSearchAfter(final OpenSearchIndexProgressState openSearchIndexProgressState, final SearchWithSearchAfterResults searchWithSearchAfterResults) {
+        if (Objects.isNull(searchWithSearchAfterResults) && Objects.isNull(openSearchIndexProgressState.getSearchAfter())) {
+            return null;
+        }
+
+        if (Objects.isNull(searchWithSearchAfterResults) && Objects.nonNull(openSearchIndexProgressState.getSearchAfter())) {
```
kkondaka, line 144: You should combine these two if statements otherwise it is little confusing.

graytaylor0: I don't understand what is confusing about it.

graytaylor0: Oh I think that second if statement can be simplified to just `if (Objects.nonNull(openSearchIndexProgressState.getSearchAfter())`

kkondaka: It is not clear from the code that line at 148  `searchWithSearchAfterResults` is not null. better way of writing would be 
```
if (Objects.isNull(searchWithSearchAfterResults)) {
      if (Objects.nonNull(openSearchIndexProgressState.getSearchAfter())( {
         return openSearchIndexProgressState.getSearchAfter();
     } else {
         return null;
     }
} 
```
This way we know after the code after the "if" block is the "else" block.

kkondaka (approved):

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/configuration/SearchConfiguration.java
```diff
@@ -10,15 +10,21 @@
...
 import org.slf4j.LoggerFactory;
+
 import java.util.Map;
 
 public class SearchConfiguration {
 
     private static final ObjectMapper objectMapper = new ObjectMapper();
     private static final Logger LOG = LoggerFactory.getLogger(SearchConfiguration.class);
 
+    // TODO: Should we default this to NONE and remove the version lookup to determine scroll or point-in-time as the default behavior?
```
oeyh, line 24: My thought is that this looks like a similar situation to s3.compression - we can set NONE as default and have "AUTO" as an explicit option to do the version-dependent search.

oeyh (approved):

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/configuration/SearchConfiguration.java
graytaylor0: That's definitely a consideration. We can think about it more as this feature gets closer to release

PR #2896, opened 2023-06-16
Do not suppress logs when there are exception in s3 source.

asuresh8:
### Description
Do not suppress logs when there is an exception when reading from S3
 
### Check List

### DIFF
--- a/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3InputStream.java
+++ b/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3InputStream.java
@@ -13,11 +13,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.core.sync.ResponseTransformer;
 import software.amazon.awssdk.http.Abortable;
+import software.amazon.awssdk.http.HttpStatusCode;
 import software.amazon.awssdk.services.s3.S3Client;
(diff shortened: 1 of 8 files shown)


chenqi0805 (approved):

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3InputStream.java
```diff
@@ -598,4 +609,12 @@ static int readFullyDirectBuffer(InputStream f, ByteBuffer buf, byte[] temp) thr
 
         return totalBytesRead;
     }
+
+    private void recordS3Exception(final S3Exception ex) {
+        if (ex.statusCode() == HttpStatusCode.NOT_FOUND) {
+            s3ObjectPluginMetrics.getS3ObjectsFailedNotFoundCounter().increment();
+        } else if (ex.statusCode() == HttpStatusCode.FORBIDDEN) {
+            s3ObjectPluginMetrics.getS3ObjectsFailedAccessDeniedCounter().increment();
+        }
```
dinujoh, line 618: How about counter for other status code ?

asuresh8: This change just maintains the existing behavior. There probably should be additional status codes checked but that's outside the scope of this change.

dinujoh (approved):

PR #2908, opened 2023-06-19
OpenSearch Sink Optimizations

engechas:
### Description
This CR tackles 3 optimizations for the OpenSearch sink:
1. Estimate bulk request size based on the compressed size of the inner documents. The addition of a new transport client in #2385 silently introduced request compression when sending data to OpenSearch sinks. This breaks the currently bulk size estimation logic as it is estimating the uncompressed size. This PR compresses a sample of each bulk request and estimates based on that. There is room for improvement on accuracy here but this is a step in the right direction.
2. Filter out unnecessary data from the BulkResponse object. We are only concerned with the erroneous bulk operations in a BulkResponse. Filtering out the rest of the data reduces the response size by over half which reduces the network load on the underlying OpenSearch domain.
3. Continue building BulkRequests across event batches. The previous behavior was to flush the end of a batch in a BulkRequest even if the record count didn't match the intended bulk size. This PR changes the behavior to continue packing BulkRequests across batches until the bulk size is reached or a flush timeout expires. 
 
### Check List


data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/IndexConfiguration.java
```diff
@@ -55,6 +57,7 @@ public class IndexConfiguration {
     private final String documentIdField;
     private final String routingField;
     private final long bulkSize;
+    private final long flushTimeout;
```
graytaylor0, line 60: If we could use a Duration to be consistent between the other data prepper durations that would be ideal.

graytaylor0 (approved):

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/IndexConfiguration.java
engechas: Discussed offline - requires a non-trivial refactor since the OpenSearch sink setting logic doesn't use the common parser that can consume DataPrepperDurationSerializer. This is something we should look into for a followup

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/ProcessWorker.java
```diff
@@ -117,9 +117,8 @@ private void doRun() {
                 processAcknowledgements(inputEvents, records);
             }
         }
-        if (!records.isEmpty()) {
```
kkondaka, line 120: Why remove this check?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -181,8 +189,12 @@ private void doInitializeInternal() throws IOException {
 
     bulkRequestSupplier = () -> new JavaClientAccumulatingBulkRequest(new BulkRequest.Builder());
     final int maxRetries = openSearchSinkConfig.getRetryConfiguration().getMaxRetries();
+    final OpenSearchClient filteringOpenSearchClient = openSearchClient.withTransportOptions(
+            TransportOptions.builder()
```
kkondaka, line 193: If these options are always fixed, why not use them in `doInitializeInternal()`?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -201,11 +213,16 @@ public boolean isReady() {
 
   @Override
   public void doOutput(final Collection<Record<Event>> records) {
-    if (records.isEmpty()) {
-      return;
+    final long threadId = Thread.currentThread().getId();
+    if (!bulkRequestMap.containsKey(threadId)) {
+      bulkRequestMap.put(threadId, bulkRequestSupplier.get());
```
kkondaka, line 218: Not a good idea. We do `bulkRequestSupplier.get()` in BulkRetryStrategy.java which may overwrite this. In general, this kind of logic could result in NullPointerExceptions as we have seen before.

kkondaka (changes requested):

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/ProcessWorker.java
engechas: To ensure data does not get stuck in the OpenSearch sink when a flush time is specified and no further traffic comes in. We have to call sink::output to check the flush time which wouldn't happen with this `if` if no traffic ever comes back in

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
engechas: The OpenSearch client has contracts on the data returned for deserialization. We can't filter out the required fields for given calls. The OpenSearch client created in `doInitializeInternal` is used for more than just bulk requests so this filter path would break its contracts

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
engechas: > We do bulkRequestSupplier.get() in BulkRetryStrategy.java which may overwrite this
Looking at the code in `BulkRetryStrategy`, a new BulkRequest is retrieved when there are errors and we need to retry. The BulkRequest is populated from the failed entries in the original bulk request then retried until success or retry limit reached.

This does not overwrite the bulkRequestMap entries here and also does not affect the logic to pack across batches since the `BulkRetryStrategy` is only invoked once a bulk request is ready to be sent. The `BulkRetryStrategy` handles the current BulkRequest to a completed state as well, the logic is never passed back to this class.

Please let me know if I am missing something.

> In general, this kind of logic could result in NullPointerExceptions as we have seen before.
Could you elaborate on how this would lead to an NPE? The entries in the map are tied to thread IDs, there is no way for a thread to interfere with another thread's entries in this map. 

There's no path to a missing key causing an NPE since we check that before getting the entry for a thread ID. We aren't removing entries from this map at any point so once the key is created it exists until the pipeline is stopped. 

There's no path to a value in this map being null since we are making the put calls against the supplier or an already created AccumulatingBulkRequest.

I understand there were issues with NPEs in a different part of the code that had multiple threads operating on the same map entry. Using the thread ID as the key is intended to prevent this as there are never threads operating on the same map entry.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
kkondaka: Is there a path where `opensearchClient` without filtering is used?

kkondaka: Discussed on phone. Resolved.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
kkondaka: Also discussed on phone.

kkondaka (approved):
Looks good.

PR #2910, opened 2023-06-20
Elasticsearch client implementation with pit and no context search

graytaylor0:
### Description
This change adds support for Elasticsearch clusters by creating an Elasticsearch client when the opensearch client is unable to lookup cluster info. 

This change also implements processing indices with point in time or no search context. 

During testing, I found that es clusters with `build-flavor` of oss do not support point in time, while a `build-flavor` of `default` does. So the current requirements to use point in time is that the es cluster has to be greater than 7.10.0, and must have `build-flavor` that is not `oss`. 

This was tested on 2 different es clusters with version 7.10.2, one with `oss` and one with `default` build-flavor.

The code is very similar to that for opensearch, just with different classes from the elasticsearch client. 
 
### Issues Resolved
Related to #1985 
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
+++ b/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.dataprepper.plugins.source.opensearch.worker;
 
+import co.elastic.clients.elasticsearch.ElasticsearchClient;
+import co.elastic.clients.elasticsearch._types.ElasticsearchException;
 import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.client.opensearch._types.OpenSearchException;
 import org.opensearch.client.opensearch.cat.IndicesResponse;
@@ -32,7 +34,9 @@ public class OpenSearchIndexPartitionCreationSupplier implements Function<Map<St
 
     private final OpenSearchSourceConfiguration openSearchSourceConfiguration;
     private final IndexParametersConfiguration indexParametersConfiguration;
-    private final OpenSearchClient openSearchClient;
+    private OpenSearchClient openSearchClient;
+    private ElasticsearchClient elasticsearchClient;
+
 
     public OpenSearchIndexPartitionCreationSupplier(final OpenSearchSourceConfiguration openSearchSourceConfiguration,
                                                     final ClusterClientFactory clusterClientFactory) {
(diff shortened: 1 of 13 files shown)


data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
```diff
@@ -70,13 +78,37 @@ private List<PartitionIdentifier> applyForOpenSearchClient(final Map<String, Obj
...
+            return Collections.emptyList();
+        }
+
+        return indicesResponse.valueBody().stream()
+                .filter(esIndicesRecord -> shouldIndexBeProcessed(null, esIndicesRecord))
+                .map(indexRecord -> PartitionIdentifier.builder().withPartitionKey(indexRecord.index()).build())
+                .collect(Collectors.toList());
+    }
+
+    private boolean shouldIndexBeProcessed(final IndicesRecord openSearchIndicesRecord, final co.elastic.clients.elasticsearch.cat.indices.IndicesRecord elasticSearchIndicesRecord) {
```
engechas, line 101: We can just take in the index name here to simplify the logic.

The calls above to this method would then be `filter(record -> shouldIndexBeProcessed(record.index()))`

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/ElasticsearchAccessor.java
```diff
@@ -13,32 +33,100 @@
...
+import java.time.Instant;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import static org.opensearch.dataprepper.plugins.source.opensearch.worker.client.model.MetadataKeyAttributes.DOCUMENT_ID_METADATA_ATTRIBUTE_NAME;
+import static org.opensearch.dataprepper.plugins.source.opensearch.worker.client.model.MetadataKeyAttributes.INDEX_METADATA_ATTRIBUTE_NAME;
 
 public class ElasticsearchAccessor implements SearchAccessor, ClusterClientFactory {
```
engechas, line 52: I think there's room to consolidate common logic between this class and OpenSearchAccessor. We can handle that down the line though

data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/SearchAccessorStrategy.java
```diff
@@ -274,18 +300,107 @@ private SSLContext getTrustAllStrategy() {
...
+    }
+
+    private void validateDistribution(final String distribution) {
+        if (!distribution.equals(OPENSEARCH_DISTRIBUTION) && !distribution.startsWith(ELASTICSEARCH_DISTRIBUTION) && !distribution.equals(OPENDISTRO_DISTRIUBTION)) {
+            throw new IllegalArgumentException(String.format("Only %s, %s, or %s distributions are supported at this time. The cluster distribution being used is '%s'",
+                    OPENSEARCH_DISTRIBUTION, OPENDISTRO_DISTRIUBTION, ELASTICSEARCH_DISTRIBUTION, distribution));
+        }
+    }
+
+    private ElasticsearchTransport createElasticSearchTransport(final org.elasticsearch.client.RestClient restClient) {
```
engechas, line 349: The client construction logic here and below belongs in another class

graytaylor0: Will make a helper factory class and move the functions for client creation to that class for opensearch and elasticsearch

engechas (approved):

oeyh (approved):

PR #2931, opened 2023-06-24
Fix DLQ writer writing empty list

kkondaka:
### Description
Fix DLQ writer writing empty list.

DQL writer is writing empty list like this sometimes. Upon investigation, found that bulkOperation is being tried even the operations count is zero. This is resulting in writing empty DLQ List when the bulk operation fails.
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkRetryStrategy.java
+++ b/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkRetryStrategy.java
@@ -258,6 +258,10 @@ public final class BulkRetryStrategy {
 
     private BulkOperationRequestResponse handleRetry(final AccumulatingBulkRequest request, final BulkResponse response, int retryCount) throws InterruptedException {
         final AccumulatingBulkRequest<BulkOperationWrapper, BulkRequest> bulkRequestForRetry = createBulkRequestForRetry(request, response);
+        if (bulkRequestForRetry.getOperationsCount() == 0) {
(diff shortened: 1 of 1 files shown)


graytaylor0 (approved):
Should we have a unit test?

asifsmohammed (approved):

JonahCalvo (approved):

graytaylor0 (approved):

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkRetryStrategy.java
```diff
@@ -258,6 +258,10 @@ private void handleFailures(final AccumulatingBulkRequest<BulkOperationWrapper,
 
     private BulkOperationRequestResponse handleRetry(final AccumulatingBulkRequest request, final BulkResponse response, int retryCount) throws InterruptedException {
         final AccumulatingBulkRequest<BulkOperationWrapper, BulkRequest> bulkRequestForRetry = createBulkRequestForRetry(request, response);
+        if (bulkRequestForRetry.getOperationsCount() == 0) {
```
dinujoh, line 261: Can you give some insight what is this getOperationsCount ?

kkondaka: It is the number of operations in the bulk request. If this operation fails (for any reason), we will retry. But if the initial request has 0 operations, then there is no need to retry and no reason to write to DLQ when we fail to write.

PR #2992, opened 2023-07-07
Retry s3 reads on socket exceptions.

asuresh8:
S3 will reset the conenction on their end frequently. To not lose data, data prepper should retry all socket exceptions by attempting to re-open the stream.

### Description
Retries all socket exceptions
 
### Issues Resolved
S3 socket exceptions causing records to get dropped
 
### Check List

### DIFF
--- a/data-prepper-plugins/parquet-codecs/src/main/java/org/opensearch/dataprepper/plugins/codec/parquet/ParquetInputCodec.java
+++ b/data-prepper-plugins/parquet-codecs/src/main/java/org/opensearch/dataprepper/plugins/codec/parquet/ParquetInputCodec.java
@@ -91,8 +91,9 @@ public class ParquetInputCodec implements InputCodec {
 
                 eventConsumer.accept(new Record<>(event));
             }
-        } catch (Exception parquetException){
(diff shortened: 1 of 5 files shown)


engechas (commented):
Changes look good to me.

I think we also need to bubble up any fatal exceptions from this class [here](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/parquet-codecs/src/main/java/org/opensearch/dataprepper/plugins/codec/parquet/ParquetInputCodec.java#L94-L96) to prevent deleting the SQS message on read failure

asuresh8:
> Changes look good to me.
> 
> I think we also need to bubble up any fatal exceptions from this class [here](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/parquet-codecs/src/main/java/org/opensearch/dataprepper/plugins/codec/parquet/ParquetInputCodec.java#L94-L96) to prevent deleting the SQS message on read failure

Good call. I've added a commit to bubble up that exception.

engechas (approved):

asifsmohammed (approved):

PR #2994, opened 2023-07-10
Updated Kafka security configuration

kkondaka:
### Description
Updated Kafka security configuration.
1. Modified to allow authentication and encryption configuration to be separate so that it can extended easily later
2. Added region and sts role arn to aws config.
3. Modified to set MSK IAM properties in the kafka configuration when that auth mechanism is used.
 
### Issues Resolved
 
### Check List


data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/KafkaSourceConfig.java
```diff
@@ -19,6 +19,21 @@
  */
 
 public class KafkaSourceConfig {
+    public class SslConfig {
+        // TODO Add support for SSL Encryption by having
+        // path to certificates, etc
+    }
+
+    public class EncryptionConfig {
+        // Absence of ssl means no-encryption
```
hshardeesi, line 28: So this means PLAINTEXT? should we make it explicitly configurable. Customers may assume SSL by default if not specified.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/AuthConfig.java
```diff
@@ -5,33 +5,65 @@
...
-        this.plainTextAuthConfig = plainTextAuthConfig;
+    public SaslAuthConfig getSaslAuthConfig() {
+        return saslAuthConfig;
     }
 
-    public PlainTextAuthConfig getPlainTextAuthConfig() {
-        return plainTextAuthConfig;
+    @AssertTrue(message = "Only one of SSL or SASL auth config must be specified")
+    public boolean hasSaslOrSslConfig() {
+        return Stream.of(sslAuthConfig, saslAuthConfig).filter(n -> n!=null).count() == 1;
```
hshardeesi, line 66: should we have a similar check on sasl_mechanism as well? only one of the mechanisms should be specified.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/AwsConfig.java
```diff
@@ -10,10 +10,26 @@
...
-        return awsMskArn;
+    @Size(min = 1, message = "Region cannot be empty string")
+    @JsonProperty("region")
+    private String region;
+
+    @Size(min = 20, max = 2048, message = "sts_role_arn length should be between 20 and 2048 characters")
+    @JsonProperty("sts_role_arn")
+    private String stsRoleArn;
+
+    public String getMskArn() {
```
hshardeesi, line 24: We will most likely need to add an option to specify what type of bootstrap servers we want to read from MSK (public, single-vpc or multi-VPC) we can probably add that later also.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/KafkaSourceConfig.java
kkondaka: But SSL requires additional config like pointers to certificates, etc. Where as no-encryption does not require any config, right?

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/AwsConfig.java
kkondaka: Isn't that decided on the MSK server side? Does client have option to read public if the server is not public?

kkondaka: Will add this in the next PR.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/KafkaSourceConfig.java
kkondaka: Defaulted to SSL.

hshardeesi (approved):

Issue #2778, opened 2023-05-29
Support CloudWatch Logs as a Sink


MaGonzalMayedo:
Problem Satement:
* It would be good to support a large variety of sinks for data output, currently there is no support for shipping logs to CloudWatch logs.

Solution:
* As a solution, I want to propose the addition of a CloudWatch Logs Sink feature to support the pipelining of data transfer from Data Prepper to CloudWatch Logs.

PR #3023, opened 2023-07-12
GitHub-Issue#2778: Added CouldWatchLogsService, Tests and RetransmissionException

MaGonzalMayedo:
### Description
This change adds the CloudWatchLogsService which is the main module to be used by the sink to push logs to CloudWatchLogs Services. It includes a similar iterative style to checking Data-Prepper events for validity and proceeds to wrap them up in InputLogEvent structures before batch sending them to CloudWathcLogs. This change also introduces a new exception for denoting an error that was derived from this plugin (Runtime Exception).

### Issues Resolved
This PR will work towards resolving CWL-Sink for Data-Prepper [Issue https://github.com/https://github.com/https://github.com/opensearch-project/data-prepper/issues/2778]
Link: https://github.com/opensearch-project/data-prepper/issues/2778#issue-1731228391
 
### Check List


data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+
+                /*TODO: Can add DLQ logic here for sending these logs to a particular DLQ for error checking. (Explicitly for bad formatted logs).
+                    as currently the logs that are able to be published but rejected by CloudWatch Logs will simply be deleted if not deferred to
+                    a backup storage.
+                 */
+//TODO: Must also consider if the customer makes the logEvent size bigger than the send request size.
+//TODO: Can inject another class for the stopWatch functionality.
+
+public class CloudWatchLogsService {
+    public static final int LOG_EVENT_OVERHEAD_SIZE = 26; //Size of overhead for each log event message.
```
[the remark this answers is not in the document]
dlvenable: Good question. It may make sense to make this into a computed line.

For example (and I'm just making this calculation up):

```
public static final int LOG_EVENT_OVERHEAD_SIZE = "CWLSTART[".length() + "]CWLEND";
```

If it is an approximation, then include that in the field name.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+        stopWatchOn = false;
+    }
+
+    /**
+     * Function handles the packaging of events into log events before sending a bulk request to CloudWatchLogs.
+     * Implements simple conditional buffer. (Sends once batch size, request size in bytes, or time limit is reached)
+     * @param logs - Collection of Record events which hold log data.
+     */
+    public void output(final Collection<Record<Event>> logs) {
+        reentrantLock.lock();
```
dlvenable, line 94: A user can define multiple worker threads, and there is a sink thread per worker thread. Locking these will work fine, but it may impact the performance. Can the lock's scope be reduced at all?

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+            for (Record<Event> singleLog: logs) {
+                String logJsonString = singleLog.getData().toJsonString();
+                int logLength = logJsonString.length();
+
+                if (thresholdCheck.isGreaterThanMaxEventSize(logLength + LOG_EVENT_OVERHEAD_SIZE)) {
+                    LOG.warn("Event blocked due to Max Size restriction! {Event Size: " + (logLength + LOG_EVENT_OVERHEAD_SIZE) + " bytes}");
+                    continue;
+                }
+
+                int bufferSizeWithOverHead = (buffer.getBufferSize() + (buffer.getEventCount() * LOG_EVENT_OVERHEAD_SIZE));
```
dlvenable, line 110: nit: "overhead" is a single word - thus `bufferSizeWithOverhead`

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        for (EventHandle eventHandle : bufferedEventHandles) {
+            eventHandle.release(result);
+        }
+
+        bufferedEventHandles.clear();
+    }
+
+    private void startStopWatch() {
```
dlvenable, line 207: You could create a small class which encapsulates the stop watch logic you are looking for. Basically, all of these private methods would be the public methods on that class. 

This snippet from above could also be encapsulated into a public method. Perhaps `startIfNotRunning()`.

```
            if (!stopWatchOn) {
                startStopWatch();
            }
```

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+
+            runExitCheck();
+
+        } catch (InterruptedException e) {
+            LOG.error("Caught InterruptedException while attempting to publish logs!");
+            reentrantLock.unlock();
+        }
+    }
+
+    private void pushLogs() throws InterruptedException {
```
dlvenable, line 129: You could probably split this out into a new class. It could be some retryable log pushing class. This could make it much easier to test.

dlvenable (changes requested):

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+                /*TODO: Can add DLQ logic here for sending these logs to a particular DLQ for error checking. (Explicitly for bad formatted logs).
+                    as currently the logs that are able to be published but rejected by CloudWatch Logs will simply be deleted if not deferred to
+                    a backup storage.
+                 */
+//TODO: Must also consider if the customer makes the logEvent size bigger than the send request size.
+//TODO: Can inject another class for the stopWatch functionality.
+
+public class CloudWatchLogsService {
+    public static final int LOG_EVENT_OVERHEAD_SIZE = 26; //Size of overhead for each log event message.
+    public static final String NUMBER_OF_RECORDS_PUSHED_TO_CWL_SUCCESS = "cloudWatchLogsEventsSucceeded";
```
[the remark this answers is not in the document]
MaGonzalMayedo: Sorry about that, I might be getting the question confused, but I there are two strings that are used to declare counters for tracking successful logs and requests (one being "cloudWatchLogsRequestsSucceeded" and another "cloudWatchLogsEventsSucceeded"). They are very similar but mainly differ in their action. I could refactor these to a better name if they are confusing though.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+    private final String logGroup;
+    private final String logStream;
+    private final int retryCount;
+    private final long backOffTimeBase;
+    private final io.micrometer.core.instrument.Counter logEventSuccessCounter; //Counter to be used on the fly for counting successful transmissions. (Success per single event successfully published).
+    private final Counter requestSuccessCount;
+    private final io.micrometer.core.instrument.Counter logEventFailCounter;
+    private final io.micrometer.core.instrument.Counter requestFailCount; //Counter to be used on the fly during error handling.
+    private int failCounter = 0;
+    private boolean failedPost;
```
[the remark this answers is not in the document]
MaGonzalMayedo: It does get reassigned quite a bit as the variable is one of the main conditionals for logic flow in this class. But where would you recommend I place it? Would you recommend removing from that scope and placing inside the main method?

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+     */
+    public void output(final Collection<Record<Event>> logs) {
+        reentrantLock.lock();
+
+        try {
+            if (!stopWatchOn) {
+                startStopWatch();
+            }
+
+            for (Record<Event> singleLog: logs) {
```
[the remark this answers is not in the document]
MaGonzalMayedo: Got it thank you! I made the change to log.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+            InputLogEvent tempLogEvent = InputLogEvent.builder()
+                    .message(new String(data))
+                    .timestamp(System.currentTimeMillis())
+                    .build();
+            logEventList.add(tempLogEvent);
+        }
+
+        while (failedPost && (failCounter < retryCount)) {
+            try {
+                PutLogEventsRequest putLogEventsRequest = PutLogEventsRequest.builder()
```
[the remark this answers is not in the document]
MaGonzalMayedo: I have moved the builder out of the try catch for the next commit.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+                        .build();
+
+                cloudWatchLogsClient.putLogEvents(putLogEventsRequest);
+
+                requestSuccessCount.increment();
+                failedPost = false;
+
+                //TODO: When a log is rejected by the service, we cannot send it, can probably push to a DLQ here.
+
+            } catch (AwsServiceException | SdkClientException e) {
```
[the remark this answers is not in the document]
MaGonzalMayedo: These exceptions are all throwable by the PutLogEvents API call, but from the documentation I believe only a few of them are actually retryable  by their nature. In our case CloudWatchLogsException is a superclass of most of these exception types so I will refactor to have that one instead of AwsServiceException as AwsServiceException is generalized across services. I believe we don't have to specifically catch all of those unless we wanted to include unique retransmission handling procedures for each distinct call. It might be better to explore this in a future update to the sink as we are mainly focused on retrying when we get an exception we can recover from.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/threshold/ThresholdCheck.java
```diff
@@ -31,7 +31,7 @@ public ThresholdCheck(final int batchSize, final int maxEventSizeBytes, final in
      */
     public boolean isGreaterThanThresholdReached(final long currentTime, final int currentRequestSize, final int batchSize) {
         return ((isGreaterThanBatchSize(batchSize) || isGreaterEqualToLogSendInterval(currentTime)
-                || isGreaterThanMaxRequestSize(currentRequestSize)) && (batchSize > 0));
+                || isGreaterThanMaxRequestSize(currentRequestSize)));
```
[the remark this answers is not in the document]
MaGonzalMayedo: Thank you, I have removed the extra parenthesis.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+                if (singleLog.getData().getEventHandle() != null) {
+                    bufferedEventHandles.add(singleLog.getData().getEventHandle());
+                }
+                buffer.writeEvent(logJsonString.getBytes());
+            }
+
+            runExitCheck();
+
+        } catch (InterruptedException e) {
+            LOG.error("Caught InterruptedException while attempting to publish logs!");
```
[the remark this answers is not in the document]
MaGonzalMayedo: I was not thinking of placing dlq logic there, but it could benefit as a "last resort", I will leave a "TODO" to address the dlq procedure for a future update of the plugin. Thank you for pointing it out, I also added a retry method so we re-attempt to process these logs as an InterruptedException does not correlate with our services timing out. This will be up for the next PR commit.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
MaGonzalMayedo: I am a bit unclear on how the actual pipeline instance works. But if there is only one sink thread per worker, is there no risk of a race condition? If so can we just avoid the lock.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
MaGonzalMayedo: I have made the name change for the next PR commit.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
MaGonzalMayedo: I have made a SinkStopWatch class that encapsulates this logic.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
MaGonzalMayedo: I have extracted a class that deals with this method, it is a little awkward but it does facilitate its readability. It has been extracted into a LogPush class. This should also help if I need to add any more tests. Thank you.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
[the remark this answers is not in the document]
markkuhn: You typically place a variable up here if it is shared. `failedPost` doesn't seem to be shared so can be declared in the methods that need it.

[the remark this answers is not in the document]
MaGonzalMayedo: Understood, thank you. I have moved the variable within the method body now for proper scope use.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+            reentrantLock.unlock();
+        }
+    }
+
+    private void pushLogs() throws InterruptedException {
+        LOG.info("Attempting to push logs! {Batch size: " + buffer.getEventCount() + "}");
```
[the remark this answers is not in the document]
MaGonzalMayedo: I was thinking of using it to log information about the average batch sizes so the client can see on average its use, but Data Prepper seems to have a similar metric already being published, I will remove this for the next change.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private void runExitCheck() throws InterruptedException {
+        int bufferSizeWithOverHead = (buffer.getBufferSize() + (buffer.getEventCount() * LOG_EVENT_OVERHEAD_SIZE));
+        if ((thresholdCheck.isEqualToThresholdReached(bufferSizeWithOverHead, buffer.getEventCount()) && (buffer.getEventCount() > 0))) {
+            pushLogs();
+        }
+    }
+
+    private void releaseEventHandles(final boolean result) {
+        if (bufferedEventHandles.size() == 0) {
```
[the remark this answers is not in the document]
MaGonzalMayedo: I will have this change implemented for the next commit.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,180 @@
...
+/*
+   TODO: Can add DLQ logic here for sending these logs to a particular DLQ for error checking. (Explicitly for bad formatted logs).
+    as currently the logs that are able to be published but rejected by CloudWatch Logs will simply be deleted if not deferred to
+    a backup storage.
+*/
+//TODO: Must also consider if the customer makes the logEvent size bigger than the send request size.
+//TODO: Can inject another class for the stopWatch functionality.
+
+public class CloudWatchLogsService {
+    public static final int APPROXIMATE_LOG_EVENT_OVERHEAD_SIZE = 26; //Size of overhead for each log event message.
```
markkuhn, line 37: only used in `ThreshholdCheck` so might be better to declare it there

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/utils/SinkStopWatch.java
```diff
@@ -0,0 +1,40 @@
+package org.opensearch.dataprepper.plugins.sink.utils;
+
+import org.apache.commons.lang3.time.StopWatch;
+
+import java.util.concurrent.TimeUnit;
+
+public class SinkStopWatch {
+    private final StopWatch stopWatch;
+    private boolean stopWatchOn;
```
markkuhn, line 9: Let's use `stopWatch.isStarted()` instead

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/utils/SinkStopWatch.java
```diff
@@ -0,0 +1,40 @@
+package org.opensearch.dataprepper.plugins.sink.utils;
```
markkuhn, line 1: This class is not needed and stopwatch methods can be moved to the service class.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/utils/LogPusher.java
```diff
@@ -0,0 +1,88 @@
...
+    public boolean pushLogs(final Buffer buffer, final CloudWatchLogsClient cloudWatchLogsClient, final String logGroup, final String logStream) throws InterruptedException {
+        boolean failedPost = true;
+        int failCounter = 0;
+
+        ArrayList<InputLogEvent> logEventList = new ArrayList<>();
+
+        for (byte[] data: buffer.getBufferedData()) {
+            InputLogEvent tempLogEvent = InputLogEvent.builder()
+                    .message(new String(data))
+                    .timestamp(System.currentTimeMillis())
```
dlvenable, line 41: Is this value correct? This sets the timestamp to the current time - should it be the time of the event itself?

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
dlvenable: The user can configure some number of pipeline workers. This is the total number of threads running. Thus, you may have N threads calling `output` on this instance. So there is a need for thread synchronization.

I'm wondering if we can reduce the scope of the lock so that only part of the work needed here is locked.

One possible solution: Synchronize the calls to write to the buffer. Then have a blocking queue for the log pushing part. In this way, some threads could write to the new buffer while one thread is responsible for pushing to CloudWatch Logs.

This is something we can improve upon later though.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/utils/LogPusher.java
```diff
@@ -0,0 +1,88 @@
...
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClient;
+import software.amazon.awssdk.services.cloudwatchlogs.model.CloudWatchLogsException;
+import software.amazon.awssdk.services.cloudwatchlogs.model.InputLogEvent;
+import software.amazon.awssdk.services.cloudwatchlogs.model.PutLogEventsRequest;
+
+import java.util.ArrayList;
+
+public class LogPusher {
```
dlvenable, line 15: Please add some unit tests for this class.

dlvenable (changes requested):

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+                String logJsonString = singleLog.getData().toJsonString();
+                int logLength = logJsonString.length();
+
+                if (thresholdCheck.isGreaterThanMaxEventSize(logLength + LOG_EVENT_OVERHEAD_SIZE)) {
+                    LOG.warn("Event blocked due to Max Size restriction! {Event Size: " + (logLength + LOG_EVENT_OVERHEAD_SIZE) + " bytes}");
+                    continue;
+                }
+
+                int bufferSizeWithOverHead = (buffer.getBufferSize() + (buffer.getEventCount() * LOG_EVENT_OVERHEAD_SIZE));
+                if ((thresholdCheck.isGreaterThanThresholdReached(getStopWatchTime(),  bufferSizeWithOverHead + logLength + LOG_EVENT_OVERHEAD_SIZE, buffer.getEventCount() + 1) && (buffer.getEventCount() > 0))) {
```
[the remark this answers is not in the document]
MaGonzalMayedo: I have restructured the CloudWatchLimit to use the ThresholdConfig, this allows for a static utils class with less parameters to be called.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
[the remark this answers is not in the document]
MaGonzalMayedo: Sorry for the lack of explanation on this one, in the official AWS documentation, there is a header that gets appended to the message, this value is held at 26. Here is the link for reference: https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+            if (!stopWatchOn) {
+                startStopWatch();
+            }
+
+            for (Record<Event> singleLog: logs) {
+                String logJsonString = singleLog.getData().toJsonString();
+                int logLength = logJsonString.length();
+
+                if (thresholdCheck.isGreaterThanMaxEventSize(logLength + LOG_EVENT_OVERHEAD_SIZE)) {
+                    LOG.warn("Event blocked due to Max Size restriction! {Event Size: " + (logLength + LOG_EVENT_OVERHEAD_SIZE) + " bytes}");
```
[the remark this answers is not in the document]
MaGonzalMayedo: The most likely scenario is during the DLQ implementation, we can consider pushing the log event to it instead, otherwise the log event cannot meet restrictions and needs to be dropped. A warning metric could be added here as well to track the number of oversizedlogs.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+        try {
+            if (!stopWatchOn) {
+                startStopWatch();
+            }
+
+            for (Record<Event> singleLog: logs) {
+                String logJsonString = singleLog.getData().toJsonString();
+                int logLength = logJsonString.length();
+
+                if (thresholdCheck.isGreaterThanMaxEventSize(logLength + LOG_EVENT_OVERHEAD_SIZE)) {
```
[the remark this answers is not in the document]
MaGonzalMayedo: Yeah, I agree, I will reformat this into a static method class so we do not need to instantiate this.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+            reentrantLock.unlock();
+        }
+    }
+
+    private void pushLogs() throws InterruptedException {
+        LOG.info("Attempting to push logs! {Batch size: " + buffer.getEventCount() + "}");
+        stopAndResetStopWatch();
+        startStopWatch();
+
+        ArrayList<InputLogEvent> logEventList = new ArrayList<>();
```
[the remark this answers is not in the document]
MaGonzalMayedo: Agreed! Thank you, I have made the change for the next revision.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+                failedPost = false;
+
+                //TODO: When a log is rejected by the service, we cannot send it, can probably push to a DLQ here.
+
+            } catch (AwsServiceException | SdkClientException e) {
+                LOG.error("Failed to push logs with error: {}", e.getMessage());
+
+                Thread.sleep(calculateBackOffTime(backOffTimeBase));
+
+                LOG.warn("Trying to retransmit request... {Attempt: " + (++failCounter) + "}");
```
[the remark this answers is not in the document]
MaGonzalMayedo: I have made the changes for the next revision.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,221 @@
...
+    private final ThresholdCheck thresholdCheck;
+    private final List<EventHandle> bufferedEventHandles;
+    private final String logGroup;
+    private final String logStream;
+    private final int retryCount;
+    private final long backOffTimeBase;
+    private final io.micrometer.core.instrument.Counter logEventSuccessCounter; //Counter to be used on the fly for counting successful transmissions. (Success per single event successfully published).
+    private final Counter requestSuccessCount;
+    private final io.micrometer.core.instrument.Counter logEventFailCounter;
+    private final io.micrometer.core.instrument.Counter requestFailCount; //Counter to be used on the fly during error handling.
```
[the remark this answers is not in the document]
MaGonzalMayedo: Sorry for the confusing nomenclature, but I normally meant to differentiate them based on their action.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/utils/LogPusher.java
MaGonzalMayedo: This value is currently more of a default, if the event itself has it, then I can sub it in. I will try to add this feature in for the next commit.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/utils/LogPusher.java
MaGonzalMayedo: Thank you, I will have the unit tests for this class ready for the next commit.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsClientFactory.java
```diff
@@ -20,6 +20,9 @@
  * CloudWatchLogs services.
  */
 public final class CloudWatchLogsClientFactory {
+    private CloudWatchLogsClientFactory() {
+        throw new IllegalStateException("Static Factory Class!");
+    }
 
     /**
      * Generates a CloudWatchLogs Client based on STS role ARN system credentials.
```
markkuhn, line 28: In the `createOverrideConfiguration()` method you use nested builder. It is better practice to use lambdas instead (easier to read).
```
return ClientOverrideConfiguration.builder()
            .retryPolicy(r -> r.numRetries(AwsConfig.DEFAULT_CONNECTION_ATTEMPTS))
            .build();
```

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/buffer/InMemoryBuffer.java
```diff
@@ -38,8 +40,9 @@ public byte[] popEvent() {
     }
 
     @Override
-    public ArrayList<byte[]> getBufferedData() {
-        return eventsBuffered;
+    public List<byte[]> getBufferedData() {
+        Collections.unmodifiableList(eventsBuffered);
```
markkuhn, line 44: I think you forgot to remove this line.

markkuhn:
For your current revision let's move all util classes (CloudWatchLogsLimits, SinkStopWatch) under a utils folder instead of `push_condition` and `time`.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,118 @@
...
+
+        this.buffer = buffer;
+        this.dispatcher = dispatcher;
+        this.bufferedEventHandles = new ArrayList<>();
+        this.cloudWatchLogsLimits = cloudWatchLogsLimits;
+        this.taskQueue = blockingQueue;
+
+        bufferLock = new ReentrantLock();
+        sinkStopWatch = new SinkStopWatch();
+        sinkThreadManager = newCachedThreadPool();
```
markkuhn, line 50: nit: A better name for this could be `scheduler` or `executor`.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+import software.amazon.awssdk.services.cloudwatchlogs.model.PutLogEventsRequest;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+
+public class CloudWatchLogsDispatcher implements Runnable {
+    public static final long UPPER_RETRY_TIME_BOUND = 20000;
+    public static final float EXP_TIME_SCALER = 1.5F;
```
markkuhn, line 25: typo

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+        boolean failedPost = true;
+        int failCounter = 0;
+
+        PutLogEventsRequest putLogEventsRequest = PutLogEventsRequest.builder()
+                .logEvents(inputLogEvents)
+                .logGroupName(logGroup)
+                .logStreamName(logStream)
+                .build();
+
+        //TODO: Could also continue to retry even with InterruptedException instead of directly pushing to DLQ.
```
markkuhn, line 77: Your TODO comments are contradictory. Do you want to keep retrying even after an exception or do you want to push to DLQ?

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
```diff
@@ -0,0 +1,145 @@
...
+import software.amazon.awssdk.services.cloudwatchlogs.model.InputLogEvent;
+import software.amazon.awssdk.services.cloudwatchlogs.model.PutLogEventsRequest;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+
+public class CloudWatchLogsDispatcher implements Runnable {
+    public static final long UPPER_RETRY_TIME_BOUND = 20000;
```
markkuhn, line 24: 20s? That's seems too high to me.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
markkuhn, line 25: Why public?

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,118 @@
...
+            int logLength = log.getData().toJsonString().length();
+
+            if (cloudWatchLogsLimits.isGreaterThanMaxEventSize(logLength)) {
+                LOG.warn("Event blocked due to Max Size restriction! {Event Size: {} bytes}", (logLength + CloudWatchLogsLimits.APPROXIMATE_LOG_EVENT_OVERHEAD_SIZE));
+                continue;
+            }
+
+            long time = sinkStopWatch.getStopWatchTimeSeconds();
+
+            bufferLock.lock();
```
markkuhn, line 70: Instead of locking from here to line 87, what about unlocking on 75 and adding locking in `addToBuffer()`?

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            bufferLock.unlock();
+        }
+    }
+
+    private void stageLogEvents() {
+        sinkStopWatch.stopAndResetStopWatch();
+
+        ArrayList<byte[]> eventMessageCloneList = new ArrayList<>();
```
markkuhn, line 94: nit: No need to append list to name. The type tells us that already,

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,118 @@
...
+        cloneLists(buffer.getBufferedData(), eventMessageCloneList);
+
+        ThreadTaskEvents dataToPush = new ThreadTaskEvents(eventMessageCloneList, bufferedEventHandles);
+        taskQueue.add(dataToPush);
+
+        buffer.clearBuffer();
+        bufferedEventHandles = new ArrayList<>();
+
+        sinkThreadManager.execute(dispatcher);
```
markkuhn, line 103: Why are we dispatching the global dispatcher? Don't we want to create a new `CloudWatchLogsDispatcher` (could also be named `CloudWatchLogsUploader`) for each batch that we flush?

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsClientFactory.java
MaGonzalMayedo: Thank you! I have added the lambda to increase code readability.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
MaGonzalMayedo: I have refactored the name for clarity.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
MaGonzalMayedo: Apologies for the confusion, I have removed that TODO comment as we only want to push to DLQ when we hit an exception or timeout during transmission.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
MaGonzalMayedo: Agreed, I will change it to 5 seconds as the cap, this should prevent the customer from waiting too long.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
MaGonzalMayedo: I have changed these classes to private.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
MaGonzalMayedo: That is true, using a global object seems like it holds too many responsibilities. I will refactor this segment to create a brand new dispatcher for each task. Thank you!

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsClientFactory.java
```diff
@@ -20,6 +19,9 @@
  * CloudWatchLogs services.
  */
 public final class CloudWatchLogsClientFactory {
+    private CloudWatchLogsClientFactory() {
+        throw new IllegalStateException("Static Factory Class!");
```
dlvenable, line 23: You probably don't need this exception.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -23,158 +19,134 @@
...
-        boolean succeededTransmission = logPusher.pushLogs(buffer, cloudWatchLogsClient, logGroup, logStream);
-        releaseEventHandles(succeededTransmission);
-
-        if (!succeededTransmission) {
-            throw new RetransmissionLimitException("Error, timed out trying to push logs! (Max retry_count reached: {" + retryCount + "})");
+    private void addToBuffer(final Record<Event> log) {
+        if (log.getData().getEventHandle() != null) {
+            bufferedEventHandles.add(log.getData().getEventHandle());
         }
+        buffer.writeEvent(log.getData().toJsonString().getBytes());
```
dlvenable, line 144: We are calling `.toJsonString()` twice. It is a somewhat slow method as it has to convert the object to a JSON string.

You can call `.toJsonString()` once at the top. Then pass in the string in this method so only one call is needed.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -23,158 +19,134 @@
...
-                int logLength = logJsonString.length();
+    private void stageLogEvents() {
+        sinkStopWatch.stopAndResetStopWatch();
 
-                if (thresholdCheck.isGreaterThanMaxEventSize(logLength + APPROXIMATE_LOG_EVENT_OVERHEAD_SIZE)) {
-                    LOG.warn("Event blocked due to Max Size restriction! {Event Size: " + (logLength + APPROXIMATE_LOG_EVENT_OVERHEAD_SIZE) + " bytes}");
-                    continue;
-                }
+        ArrayList<byte[]> eventMessageClone = new ArrayList<>();
+        cloneLists(buffer.getBufferedData(), eventMessageClone);
```
dlvenable, line 119: You copy this list and then clear the buffer. Can we avoid the copy? Since we are in the lock, we could get the `List` and then instead of calling `clear()` on the list, create a new list in the buffer.

dlvenable (changes requested):

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsClientFactory.java
MaGonzalMayedo: I have removed the exception for the next revision.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
MaGonzalMayedo: I have added a resetBuffer() method that is meant to create a new list and subbed it in for the method call. I think this should do the trick.

dlvenable (dismissed):
This looks good to me. Thanks @MaGonzalMayedo !

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,152 @@
...
+                .taskQueue(taskQueue)
+                .cloudWatchLogsClient(cloudWatchLogsClient)
+                .cloudWatchLogsMetrics(cloudWatchLogsMetrics)
+                .logGroup(logGroup)
+                .logStream(logStream)
+                .retryCount(retryCount)
+                .backOffTimeBase(backOffTimeBase)
+                .build();
+
+        asyncExecutor.execute(newTaskDispatcher);
```
markkuhn, line 137: I dont think creating an entire `CloudWatchLogsDispatcher` is the way to go. This class is fairly big and includes a lot of stuff we can run on the same thread and stuff we don't need to re-create all the time.

MaGonzalMayedo: Agreed, I will try to minimize the the requirements for a log publisher. This way we can have more lightweight objects running on each thread.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsService.java
```diff
@@ -0,0 +1,152 @@
...
+                stageLogEvents();
+                addToBuffer(log, logString);
+            } else if (cloudWatchLogsLimits.isEqualToLimitReached(bufferSizeWithAddedEvent, bufferEventCountWithEvent)) {
+                addToBuffer(log, logString);
+                stageLogEvents();
+            } else {
+                addToBuffer(log, logString);
+            }
+
+            bufferLock.unlock();
```
paggynie, line 112: please put behind `finally` block
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantLock.html

MaGonzalMayedo: Thank you! I missed this while reading the documentation, I have made the change in the latest revision.

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Will read in a collection of log messages in byte form and transform them into a collection of InputLogEvents.
+     * @param eventMessageBytes Collection of byte arrays holding event messages.
+     * @return List of InputLogEvents holding the wrapped event messages.
+     */
+    public List<InputLogEvent> prepareInputLogEvents(final Collection<byte[]> eventMessageBytes) {
+        List<InputLogEvent> logEventList = new ArrayList<>();
+
+        /**
+         * Current implementation, timestamp is generated by system time during transmission.
```
markkuhn, line 60:
```suggestion
         * In the current implementation, the timestamp is generated during transmission.
```

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
```diff
@@ -0,0 +1,163 @@
...
+
+                    } catch (CloudWatchLogsException | SdkClientException e) {
+                        LOG.error("Failed to push logs with error: {}", e.getMessage());
+                        cloudWatchLogsMetrics.increaseRequestFailCounter(1);
+                        Thread.sleep(calculateBackOffTime(backOffTimeBase, failCount));
+                        failCount++;
+                    }
+                }
+            } catch (InterruptedException e) {
+                LOG.warn("Got interrupted while waiting!");
```
markkuhn, line 128: Better log needed here. This doesn't tell the customer anything. Interrupted by what? stacktrace? while waiting for what? in what class?

markkuhn (approved):

data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/client/CloudWatchLogsDispatcher.java
MaGonzalMayedo: Thank you, I have added a more appropriate message alongside the actual thrown error message.

dlvenable (approved):

markkuhn:
🚀

PR #3029, opened 2023-07-14
Kafka Source - Cleanup and Enhancements for MSK

kkondaka:
### Description
Kafka Source - Cleanup and Enhancements for MSK
1. Added functionality to get bootstrap servers from MSK cluster if MSK with IAM is used
2. Added validation to Kafka configurations
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/kafka-plugins/build.gradle
+++ b/data-prepper-plugins/kafka-plugins/build.gradle
@@ -21,9 +21,9 @@ dependencies {
     implementation 'io.confluent:kafka-schema-registry-client:7.3.3'
     implementation 'io.confluent:kafka-schema-registry:7.3.3:tests'
     implementation 'software.amazon.msk:aws-msk-iam-auth:1.1.6'
-    implementation 'com.amazonaws:aws-java-sdk-kafka:1.12.169'
-    implementation 'com.amazonaws:aws-java-sdk:1.0.12'
-    implementation 'com.amazonaws:aws-java-sdk-sts:1.12.505'
+    implementation 'software.amazon.awssdk:sts:2.20.103'
+    implementation 'software.amazon.awssdk:auth:2.20.103'
+    implementation 'software.amazon.awssdk:kafka:2.20.103'
     testImplementation 'org.mockito:mockito-inline:4.1.0'
     testImplementation 'org.yaml:snakeyaml:2.0'
     testImplementation testLibs.spring.test
--- a/data-prepper-plugins/kafka-plugins/build.gradle
+++ b/data-prepper-plugins/kafka-plugins/build.gradle
@@ -21,6 +21,9 @@ dependencies {
     implementation 'io.confluent:kafka-schema-registry-client:7.3.3'
     implementation 'io.confluent:kafka-schema-registry:7.3.3:tests'
     implementation 'software.amazon.msk:aws-msk-iam-auth:1.1.6'
+    implementation 'com.amazonaws:aws-java-sdk-kafka:1.12.169'
+    implementation 'com.amazonaws:aws-java-sdk:1.0.12'
+    implementation 'com.amazonaws:aws-java-sdk-sts:1.12.505'
     testImplementation 'org.mockito:mockito-inline:4.1.0'
     testImplementation 'org.yaml:snakeyaml:2.0'
     testImplementation testLibs.spring.test
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/AwsConfig.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/AwsConfig.java
@@ -18,7 +18,7 @@ public class AwsConfig {
         private String arn;
     
         @JsonProperty("broker_connection_type")
-        private MskBrokerConnectionType brokerConnectionType = MskBrokerConnectionType.PUBLIC;
+        private MskBrokerConnectionType brokerConnectionType = MskBrokerConnectionType.SINGLE_VPC;
 
         public String getArn() {
             return arn;
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/AwsConfig.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/AwsConfig.java
@@ -18,7 +18,7 @@ public class AwsConfig {
         private String arn;
     
         @JsonProperty("broker_connection_type")
-        private MskBrokerConnectionType brokerConnectionType;
+        private MskBrokerConnectionType brokerConnectionType = MskBrokerConnectionType.PUBLIC;
 
         public String getArn() {
             return arn;
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
@@ -25,6 +25,7 @@ import org.opensearch.dataprepper.model.source.Source;
 import org.opensearch.dataprepper.model.acknowledgements.AcknowledgementSetManager;
 import org.opensearch.dataprepper.plugins.kafka.configuration.KafkaSourceConfig;
 import org.opensearch.dataprepper.plugins.kafka.configuration.EncryptionType;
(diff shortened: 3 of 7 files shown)


data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
```diff
@@ -154,13 +166,60 @@ private long calculateLongestThreadWaitingTime() {
...
 
-    private Properties getConsumerProperties(TopicConfig topicConfig) {
+    public String getBootStrapServersForMsk(final AwsIamAuthConfig awsIamAuthConfig, final AwsConfig awsConfig) {
+        ClientConfiguration clientConfiguration = new ClientConfiguration();
+        AWSCredentialsProvider credentialProvider;
+        if (awsIamAuthConfig == AwsIamAuthConfig.DEFAULT) {
+            credentialProvider = new DefaultAWSCredentialsProviderChain();
+        } else if (awsIamAuthConfig == AwsIamAuthConfig.ROLE) {
+            String sessionName = "data-prepper-kafka-session"+UUID.randomUUID();
+            credentialProvider = new STSAssumeRoleSessionCredentialsProvider.Builder(awsConfig.getStsRoleArn(), sessionName).build();
```
hshardeesi, line 176: Do we need to wrap it in try/catch and log the exception? Also if it fails will we retry with backoff? like we retry from opensearch sink.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
```diff
@@ -154,13 +166,60 @@ private long calculateLongestThreadWaitingTime() {
...
+                .withClientConfiguration(clientConfiguration)
+                .withCredentials(credentialProvider)
+                .withRegion(awsConfig.getRegion())
+                .build();
+        GetBootstrapBrokersRequest request = new GetBootstrapBrokersRequest();
+        String clusterArn = awsConfig.getAwsMskConfig().getArn();
+        request.setClusterArn(clusterArn);
+        GetBootstrapBrokersResult result = kafkaClient.getBootstrapBrokers( request );
+        // TODO return this based on the broker_connection_mode
+        return result.getBootstrapBrokerStringSaslIam();
```
hshardeesi, line 190: This should be bootstrapBrokerStringVpcConnectivitySaslIam()
Also would be good to add try/catch and log exception

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
```diff
@@ -154,13 +166,60 @@ private long calculateLongestThreadWaitingTime() {
...
+    }
+
+    private Properties getConsumerProperties(final TopicConfig topicConfig) {
         Properties properties = new Properties();
+        AwsIamAuthConfig awsIamAuthConfig = null;
+        AwsConfig awsConfig = sourceConfig.getAwsConfig();
+        if (sourceConfig.getAuthConfig() != null) {
```
hshardeesi, line 197: What if user config aws: but not authConfig? it might be good to have separate validation class to check all the combinations.

hshardeesi, line 197: Can user configure awsAuth without sourceConfig.getAuthConfig()? or with non-SASL config. it may be better to create a separate validation class to cover all combinations.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
kkondaka: For some reason bootstrapBrokerStringVpcConnectivitySaslIam() is not working. I added "TODO" comment to investigate further and fix it later.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
kkondaka: Yes, they could configure awsConfig without authConfig but it won't be used. On the other hand, if they configured authConfig but not awsConfig, I will have to print error and shutdown the pipeline. Let me add that.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
kkondaka: I am getting the error `symbol not found` when using that API. I will investigate and update it in the next PR.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
kkondaka: This is just a builder. I don't think it can throw exception. The `kafkaClient.getBootstrapBrokers` below may result in exceptions and some of which are retry-able. I will add that logic.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
kkondaka: I may have to move to newer version of SDK. Will do it in the next PR.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
kkondaka: Check for AwsConfig being null when non-null config is expected is already there.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/AwsConfig.java
```diff
@@ -18,7 +18,7 @@ public static class AwsMskConfig {
         private String arn;
     
         @JsonProperty("broker_connection_type")
-        private MskBrokerConnectionType brokerConnectionType;
+        private MskBrokerConnectionType brokerConnectionType = MskBrokerConnectionType.PUBLIC;
```
hshardeesi, line 21: single VPC connection type should be default. but it can be addressed in next commit

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/source/KafkaSource.java
```diff
@@ -205,8 +220,11 @@ public String getBootStrapServersForMsk(final AwsIamAuthConfig awsIamAuthConfig,
             LOG.info("Failed to get bootstrap server information from MSK, using user configured bootstrap servers");
             return sourceConfig.getBootStrapServers();
         }
-        // TODO return this based on the broker_connection_mode
-        return result.getBootstrapBrokerStringSaslIam();
+        if (awsMskConfig.getBrokerConnectionType() == MskBrokerConnectionType.PUBLIC) {
+            return result.bootstrapBrokerStringPublicSaslIam();
+        } else {
+            return result.bootstrapBrokerStringVpcConnectivitySaslIam();
```
hshardeesi, line 226: There are 3 connection types. single-vpc, public and multi-Vpc

hshardeesi (approved):
left few minor comments

hshardeesi (approved):

hshardeesi (approved):

data-prepper-plugins/kafka-plugins/build.gradle
```diff
@@ -21,6 +21,9 @@ dependencies {
     implementation 'io.confluent:kafka-schema-registry-client:7.3.3'
     implementation 'io.confluent:kafka-schema-registry:7.3.3:tests'
     implementation 'software.amazon.msk:aws-msk-iam-auth:1.1.6'
+    implementation 'software.amazon.awssdk:sts:2.20.103'
```
asifsmohammed, line 24: minor: it's better to not set sdk version here but you can change this later

asifsmohammed (dismissed):

asifsmohammed (approved):

PR #3050, opened 2023-07-21
Translate Plugin - Added functionality to provide full path to the source field

vishalboin:
### Description
Removed the iterateOn option. Now the iterateOn functionality will be taken care by source option with the help of full path to the source.

### Configuring `source` option with path
* The source provided can be full path to the field in the log that requires translation.  
* If the source path is something like `field1/field2/source`, all the objects that satisfy the path will be translated based on the provided mappings and the targets fields will be placed in the object along the path `field1/field2/target`.  
* The fields (`field1`, `field2`) leading up to the source field in the path must be JSON arrays or JSON objects. It is important to note that other JSON data types cannot be iterated in this context.
* The `translate_when` expression if configured will only be applied for the JSON objects with reference to the path `field1/field2`.
* As `source` option supports multiple sources to be configured, it is required for all the sources in the array to have a common root path.

  ```yaml
  mappings:
    - source: ["/field1/field2/source1", "field1/field2/source2"]
      targets:
        - target: "targetField"
           map:
            foo: bar
            
  ```

  Let the pipeline configuration be:

  ```yaml
  processor:
  - translate:
      mappings:
        - source: collection/status
          targets:
            - target: result
              map:
                foo: bar
        - source: collection/status/http
          targets:
            - target: result
              map:
                120: "success"
                404: "failure"
  ```
  Let the contents of `logs_json.log` be:
  ```json
  {
    "collection": [
      {
        "status": "foo"
      },
      {
        "status": [
          {
            "http": 120
          }
        ]
      }
    ]
  }
  ```
  The translated log would look like this:
  ```json
  {
    "collection": [
      {
        "status": "foo",
        "result": "bar"
      },
      {
        "status": [
          {
            "http": 120,
            "result": "success"
          }
        ]
      }
    ]
  }

  ```
 
### Issues Resolved
https://github.com/opensearch-project/data-prepper/issues/1914
 
### Check List

### DIFF
--- a/data-prepper-plugins/translate-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/translate/MappingsParameterConfig.java
+++ b/data-prepper-plugins/translate-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/translate/MappingsParameterConfig.java
@@ -15,19 +15,12 @@ public class MappingsParameterConfig {
     @NotNull
     private Object source;
 
-    @JsonProperty("iterate_on")
-    private String iterateOn;
-
     @JsonProperty("targets")
     @Valid
     private List<TargetsParameterConfig> targetsParameterConfigs = new ArrayList<>();
 
     public Object getSource() {
         return source;
-    }
-
-    public String getIterateOn() {
-        return iterateOn;
     }
 
     public List<TargetsParameterConfig> getTargetsParameterConfigs() {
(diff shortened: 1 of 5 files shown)


kkondaka (approved):

data-prepper-plugins/translate-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/translate/MappingsParameterConfig.java
```diff
@@ -54,11 +47,29 @@ public boolean isSourceFieldValid() {
         }
         if (source instanceof List<?>) {
             List<?> sourceList = (List<?>) source;
-            return sourceList.stream().allMatch(sourceItem -> sourceItem instanceof String);
+            if(sourceList.isEmpty()){
+                return false;
+            }
+            return sourceList.stream().allMatch(sourceItem -> sourceItem instanceof String)
+                   && commonRootPath(sourceList);
```
graytaylor0, line 54: Is there a reason all the `source` keys have to have a common root path? If not we should allow different source keys. Can be in future PR

graytaylor0 (approved):

PR #3055, opened 2023-07-24
Exclude keys

### LINKED ISSUE #890: Potential KeyValueProcessor feature: exclude_keys
An array specifying the parsed keys which should not be added to the event. By default no keys will be excluded.

See: https://www.elastic.co/guide/en/logstash/current/plugins-filters-kv.html#plugins-filters-kv-exclude_keys

shenkw1:
### Description
Allow configuration for an array specifying the parsed keys which should not be added to the event. By default no keys will be excluded
 
### Issues Resolved
#890 
 
### Check List


data-prepper-plugins/key-value-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/keyvalue/KeyValueProcessor.java
```diff
@@ -100,6 +101,16 @@ public KeyValueProcessor(final PluginMetrics pluginMetrics, final KeyValueProces
         }
 
         if (keyValueProcessorConfig.getIncludeKeys() != null) {
```
chenqi0805, line 103: This validation block should be simplified:

1. both includeKeys and excludeKeys in the config have default values so the null check is unnecessary
2. What if there is overlap between includeKeys and excludeKeys? Shall we also throw Exception? consider using Set.retainAll to calculate intersection between the includeKeys and excludeKeys.

data-prepper-plugins/key-value-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/keyvalue/KeyValueProcessor.java
```diff
@@ -100,6 +101,16 @@ public KeyValueProcessor(final PluginMetrics pluginMetrics, final KeyValueProces
...
+            if (keyValueProcessorConfig.getExcludeKeys() != null) {
+                if (keyValueProcessorConfig.getIncludeKeys().equals(keyValueProcessorConfig.getExcludeKeys())) {
+                    if (!keyValueProcessorConfig.getIncludeKeys().isEmpty()) {
+                        throw new IllegalStateException("Include keys and exclude keys set cannot be the same", null);
```
chenqi0805, line 107: Nit: `IllegalArgumentException` might be more appropriate

data-prepper-plugins/key-value-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/keyvalue/KeyValueProcessorTests.java
```diff
@@ -276,6 +277,34 @@ void testIncludeKeysAsNullKeyValueProcessor() {
...
+
+        final Record<Event> record = getMessage("key1=value1&key2=value2");
+        final List<Record<Event>> editedRecords = (List<Record<Event>>) keyValueProcessor.doExecute(Collections.singletonList(record));
+        final LinkedHashMap<String, Object> parsed_message = getLinkedHashMap(editedRecords);
+
+        assertThat(parsed_message.size(), equalTo(2));
+        assertThatKeyEquals(parsed_message, "key1", "value1");
+        assertThatKeyEquals(parsed_message, "key2", "value2");
+    }
+
```
chenqi0805, line 307: Let us also add test case to cover the validation between `includeKeys` and `excludeKeys`.

data-prepper-plugins/key-value-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/keyvalue/KeyValueProcessor.java
```diff
@@ -100,18 +100,20 @@ public KeyValueProcessor(final PluginMetrics pluginMetrics, final KeyValueProces
...
-        if (keyValueProcessorConfig.getIncludeKeys() != null) {
-            if (keyValueProcessorConfig.getExcludeKeys() != null) {
-                if (keyValueProcessorConfig.getIncludeKeys().equals(keyValueProcessorConfig.getExcludeKeys())) {
-                    if (!keyValueProcessorConfig.getIncludeKeys().isEmpty()) {
-                        throw new IllegalStateException("Include keys and exclude keys set cannot be the same", null);
-                    }
-                }
-
-                excludeKeysSet.addAll(keyValueProcessorConfig.getExcludeKeys());
+        if (keyValueProcessorConfig.getIncludeKeys().equals(keyValueProcessorConfig.getExcludeKeys())) {
```
chenqi0805, line 103: I think you could absorb this into the validation logic that checks `Include keys and exclude keys set cannot have any overlap` since it is just a special case.

data-prepper-plugins/key-value-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/keyvalue/KeyValueProcessor.java
```diff
@@ -99,8 +100,14 @@ public KeyValueProcessor(final PluginMetrics pluginMetrics, final KeyValueProces
             throw new PatternSyntaxException("delete_value_regex is not a valid regex string", keyValueProcessorConfig.getDeleteValueRegex(), -1);
         }
 
-        if (keyValueProcessorConfig.getIncludeKeys() != null) {
-            includeKeysSet.addAll(keyValueProcessorConfig.getIncludeKeys());
+        includeKeysSet.addAll(keyValueProcessorConfig.getIncludeKeys());
+        excludeKeysSet.addAll(keyValueProcessorConfig.getExcludeKeys());
+
+        Set<String> intersectionSet = new HashSet<String>(includeKeysSet);
+        if (intersectionSet.retainAll(excludeKeysSet)) {
```
chenqi0805, line 107: Seems to me this will return true if this set changed as a result of the call, which is not relevant to our validation. i.e. we could have includeKeysSet unchanged as a result of this call but still have complete overlap with excludeKeySet

shenkw1: My idea was to create a copy of the includeKeysSet and compare that with excludeKeysSet, since retainAll() will modify the set. If the set is changed as a result of the call, that means that there is some overlap. I didn't want to completely overwrite includeKeysSet in case the original data was still needed somewhere

chenqi0805: I think the outer if condition can be removed
```suggestion
        intersectionSet.retainAll(excludeKeysSet)
        if (!intersectionSet.isEmpty()) {
```

data-prepper-plugins/key-value-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/keyvalue/KeyValueProcessor.java
```diff
@@ -99,8 +100,17 @@ public KeyValueProcessor(final PluginMetrics pluginMetrics, final KeyValueProces
             throw new PatternSyntaxException("delete_value_regex is not a valid regex string", keyValueProcessorConfig.getDeleteValueRegex(), -1);
         }
 
-        if (keyValueProcessorConfig.getIncludeKeys() != null) {
-            includeKeysSet.addAll(keyValueProcessorConfig.getIncludeKeys());
+        includeKeysSet.addAll(keyValueProcessorConfig.getIncludeKeys());
+        excludeKeysSet.addAll(keyValueProcessorConfig.getExcludeKeys());
+
+        if (includeKeysSet.equals(excludeKeysSet) && !includeKeysSet.isEmpty()) {
```
chenqi0805, line 106: I think this check is still unnecessary.

chenqi0805: Also as best practice, it is more readable to encapsulate L106-L114 into a separate private method `validateKeySets()` rather than expose the validation details in the constructor.

chenqi0805 (dismissed):
non-blocking comments.

chenqi0805 (approved):

data-prepper-plugins/key-value-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/keyvalue/KeyValueProcessor.java
```diff
@@ -153,6 +155,14 @@ private boolean validateRegex(final String pattern)
         return true;
     }
 
+    private void validateKeySets(final Set<String> includeSet, final Set<String> excludeSet) {
+        Set<String> intersectionSet = new HashSet<String>(includeSet);
```
JonahCalvo, line 159: nit: final

JonahCalvo (approved):
One non-blocking nit

PR #3121, opened 2023-08-08
Add ttl to all dynamo source coordination store items on creation, not just when they are COMPLETED

graytaylor0:
### Description
Previously, if TTL was enabled on the dynamo db store for source coordination, only COMPLETED partitions would get a TTL and expire after the duration specified in the config.

This change makes it so that the TTL is set on every item in the store on item creation and extended when the item is updated, not just the COMPLETED partitions.
 
### Check List


cmanning09 (dismissed):
The change looks good to me.

Just for clarification - the advantage to making this change is that all items in the ddb source coordination store will have a ttl for retrying. Am I reading into this correctly?

graytaylor0:
> The change looks good to me.
> 
> Just for clarification - the advantage to making this change is that all items in the ddb source coordination store will have a ttl for retrying. Am I reading into this correctly?

Yes all items will get the ttl if now if it configured on the store, instead of just the COMPLETED items.

dlvenable (approved):

cmanning09 (approved):

PR #3168, opened 2023-08-15
Only log that the scan is complete one time for s3 scan

graytaylor0:
### Description
Track when single scan is complete to only log once when the scan has already been completed
 
### Check List

### DIFF
--- a/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3ScanPartitionCreationSupplier.java
+++ b/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3ScanPartitionCreationSupplier.java
@@ -36,6 +36,7 @@ public class S3ScanPartitionCreationSupplier implements Function<Map<String, Obj
     private static final String BUCKET_OBJECT_PARTITION_KEY_FORMAT = "%s|%s";
     static final String SCAN_COUNT = "SCAN_COUNT";
     static final String LAST_SCAN_TIME = "LAST_SCAN_TIME";
+    static final String SINGLE_SCAN_COMPLETE = "SINGLE_SCAN_COMPLETE";
 
     private final S3Client s3Client;
     private final BucketOwnerProvider bucketOwnerProvider;
@@ -143,6 +144,7 @@ public class S3ScanPartitionCreationSupplier implements Function<Map<String, Obj
 
     private void initializeGlobalStateMap(final Map<String, Object> globalStateMap) {
         globalStateMap.put(SCAN_COUNT, 0);
+        globalStateMap.put(SINGLE_SCAN_COMPLETE, false);
     }
 
     private boolean isLastModifiedTimeAfterMostRecentScanForBucket(final String bucketName,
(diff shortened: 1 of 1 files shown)


data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3ScanPartitionCreationSupplier.java
```diff
@@ -174,14 +176,22 @@ private Instant getMostRecentLastModifiedTimestamp(final ListObjectsV2Response l
     }
 
     private boolean shouldScanBeSkipped(final Map<String, Object> globalStateMap) {
+
         if (Objects.isNull(schedulingOptions) && hasAlreadyBeenScanned(globalStateMap)) {
-            LOG.info("Skipping scan because the buckets have already been scanned once");
+
+            if (!(Boolean) globalStateMap.get(SINGLE_SCAN_COMPLETE)) {
+                LOG.info("Single S3 scan has already been completed");
+                globalStateMap.put(SINGLE_SCAN_COMPLETE, true);
```
asifsmohammed, line 184: What's the difference between `SINGLE_SCAN_COMPLETE` and `SCAN_COUNT`? can we just use SCAN_COUNT to determine if scan is complete?

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3ScanPartitionCreationSupplier.java
```diff
@@ -174,14 +176,22 @@ private Instant getMostRecentLastModifiedTimestamp(final ListObjectsV2Response l
     }
 
     private boolean shouldScanBeSkipped(final Map<String, Object> globalStateMap) {
+
         if (Objects.isNull(schedulingOptions) && hasAlreadyBeenScanned(globalStateMap)) {
-            LOG.info("Skipping scan because the buckets have already been scanned once");
+
+            if (!(Boolean) globalStateMap.get(SINGLE_SCAN_COMPLETE)) {
+                LOG.info("Single S3 scan has already been completed");
```
cmanning09, line 183: Why log only during the seconds attempt at scanning? Why not log once the first scan is complete?

graytaylor0: We could log after the first scan, but with source coordination this will only get run again after all the scanned objects have been processed, so as far as timing of the log it makes more since when trying to scan again

asifsmohammed (approved):

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/S3ScanPartitionCreationSupplier.java
graytaylor0: SCAN_COUNT could block scanning since the global state doesn't go away. so it's better to keep it separate for this log

cmanning09 (approved):

PR #3241, opened 2023-08-24
Catch when no object exists and mark as completed in s3 scan

graytaylor0:
### Description
If the order of operations for s3 scan is 

```
1. scan bucket for objects and create partitions
2. one of the objects is deleted in the bucket
3. this object is picked up from the store for processing
```

then a NoSuchKeyException is thrown, instead of this deleted object causing a cycle of retrying, which would essentially make that node useless, we log that the object could not be found, mark it as completed, and continue processing
 
### Issues Resolved
 
### Check List


dlvenable (approved):
Nice catch and fix!

asifsmohammed (approved):

PR #3374, opened 2023-09-21
Add _id as additional sort key for point-in-time and search_after

graytaylor0:
### Description
Searching with search_after solely based on `_doc` sort has a chance to skip documents, and the chance increases the smaller the `batch_size` is made. 

While Point-in-time is said to have built-in ways to handle this, it does not seem to add the `_shard_doc` parameter implicitly for PIT. In OS 2.4, `_shard_doc` was in documentation and 2.5+ it is not supported.

To workaround this, search_after and PIT will add a secondary sort key of `_id`, which is the document id. This guarantees a unique sort key that ensures no documents will be skipped with search_after. 

This is not the optimal setting performance wise, as the recommended approach is to have a unique field in the document itself. This could be an optimization that some users may want in the future. However, to prioritize user experience and data durability, this is the best default approach.

* Also added fix for Json Enum parsing that was broken by dependabot update to `jackson-datatype-2.15](com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.15.2)`
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/configuration/SearchConfiguration.java
+++ b/data-prepper-plugins/opensearch-source/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/configuration/SearchConfiguration.java
@@ -5,28 +5,40 @@
 
 package org.opensearch.dataprepper.plugins.source.opensearch.configuration;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.validation.constraints.AssertTrue;
 import org.opensearch.dataprepper.plugins.source.opensearch.worker.client.model.SearchContextType;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class SearchConfiguration {
 
-    private static final ObjectMapper objectMapper = new ObjectMapper();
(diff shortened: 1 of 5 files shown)


oeyh (approved):
Changes look good!

Can you elaborate on this: 
```
Searching with search_after solely based on _doc sort has a chance to skip documents, and the chance increases the smaller the batch_size is made.
``` 

What is the `_doc` field? Why sorting by _doc can result in skipped documents? is it not a unique field?

graytaylor0:
> Changes look good!
> 
> Can you elaborate on this:
> 
> > Searching with search_after solely based on _doc sort has a chance to skip documents, and the chance increases the smaller the batch_size is made.
> 
> What is the `_doc` field? Why sorting by _doc can result in skipped documents? is it not a unique field?

`_doc` field is just a default sort that groups documents, but it can have collisions. For example, if you have 10 documents, the sort value of `_doc` for these 10 may be `1, 1, 1, 1, 2, 2, 2, 2, 3, 3`. If batch_size is 2, for example, we would get the first 2 docs with sort of `1`, and then we would pass `search_after` as 1 on the following search call, and it would skip the 3rd and 4th documents with sort value of 1 and start at the fifth one with sort value 2. 

This is a known limitation with `search_after`, where not having a  unique sort key can result in skipping. The only way to guarantee no skipping without user input to provide a unique sort key in the documents is to use the `_id` field as a tie-breaker for sort so we don't skip these documents

travisbenedict:
LGTM

travisbenedict (approved):

PR #3461, opened 2023-10-09
Fix #3459 Unit tests fail on Windows machine

### LINKED ISSUE #3459: [BUG] Unit tests fail on Windows machine
**Describe the bug**
There are a lot of test failures on the Windows system.
1. SinkModelTest. It uses hardcode "\n" in the tests, but it is "\r\n" on Windows.
2. DataPrepperArgsTest. It uses a hardcode "/" path separator to compare the path, but on Windows which is "\\"
3. DateProcessorTests. The test uses the wrong time API to compare one local time in UTC and one in the local timezone.
4. InMemorySourceCoordinationStoreTest. There is a mistaken assumption that if Java code calls Instant.now() twice, the second call will return a later time than the first one. This is not correct, as on my Windows system, it returns the same time for both calls sometimes.
5. QueuedPartitionsItemTest. Similar issues as the above one.
6. RSSSourceTest. It uses a real-world RSS feed which requires an internet connection when running this test.
7. ParquetOutputCodecTest. The test doesn't close all outputStream objects which causes filelocked exception on Windows when deleting.
8. InMemoryBufferTest. Similar to the above Instant.now() issues.

**To Reproduce**
Just run `gradlew test` and meet a lot of errors.

**Expected behavior**
No unit tests error on main branch

**Screenshots**

 - OS: Windows 10, 21H2
 - Version: main branch

**Additional context**
I understand that Windows may not be the application runtime target. However, it seems there are several test code issues in the project.

topikachu:
1. SinkModelTest: Use system System.lineSeparator() instead of hardcode '\n'
2. DataPrepperArgsTest: Covert file path separators to the local system.
3. DateProcessorTests: Covert time to the same timezone before comparing.
4. InMemorySourceCoordinationStoreTest: Use greaterThanOrEqualTo to compare time since they may be the same.
5. QueuedPartitionsItemTest: Use sleep to get two different time instances.
6. RSSSourceTest: Use a mocker server to avoid an internet connection.
7. ParquetOutputCodecTest: Close all outputStream objects in the tests.
8. org.opensearch.dataprepper.plugins.sink.s3.accumulator.InMemoryBufferTest#getDuration_provides_duration_within_expected_range: No solution to fix. Disable it. Please see my comments in the test file.

### Description
Fix unstable or failed unit tests
 
### Issues Resolved
Resolves #3459 
 
### Check List

### DIFF
--- a/data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SinkModelTest.java
+++ b/data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SinkModelTest.java
@@ -87,7 +87,7 @@ class SinkModelTest {
 
         final String expectedJson = createStringFromInputStream(this.getClass().getResourceAsStream("sink_plugin.yaml"));
 
-        assertThat("---\n" + actualJson, equalTo(expectedJson));
+        assertThat("---" + System.lineSeparator() + actualJson, equalTo(expectedJson));
         assertThat(sinkModel.getTagsTargetKey(), equalTo(tagsTargetKey));
 
     }
@@ -140,7 +140,7 @@ class SinkModelTest {
 
         final String expectedJson = createStringFromInputStream(this.getClass().getResourceAsStream("/serialized_with_plugin_settings.yaml"));
 
-        assertThat("---\n" + actualJson, equalTo(expectedJson));
+        assertThat("---" + System.lineSeparator() + actualJson, equalTo(expectedJson));
     }
 
     @Test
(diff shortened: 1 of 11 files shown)


dlvenable (approved):
Nice, thank you @topikachu for this contribution!

One other thing that could be helpful would be to run the GitHub Actions on a Windows machine as well.

I think we could add a new matrix for windows in the following section.

https://github.com/opensearch-project/data-prepper/blob/774d71df8404a2f2a0310e95b654cc45865160d9/.github/workflows/gradle.yml#L13-L18

Would you be interested in creating a PR for that as well?

PR #3500, opened 2023-10-13
Miscellaneous improvements  for DynamoDB source

daixba:
### Miscellaneous improvements for DynamoDB source
Few changes made includes:
-  Add more informational logs for each processes
-  Some minor improvements of performance, such as reduce the interval of taking leases.
-  Fix one bug that the listing of shards is not working properly.
 
 
### Check List

### DIFF
--- a/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/DynamoDBService.java
+++ b/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/DynamoDBService.java
@@ -129,15 +129,15 @@ public class DynamoDBService {
      * Future optimization can be done to accept configuration changes
      */
     public void init() {
-        LOG.info("Start initialize DynamoDB service");
+        LOG.info("Try to initialize DynamoDB service");
 
         final Optional<EnhancedSourcePartition> initPartition = coordinator.acquireAvailablePartition(InitPartition.PARTITION_TYPE);
         if (initPartition.isEmpty()) {
             // Already initialized. Do nothing.
+            LOG.info("DynamoDB service is already initialized.");
             return;
         }
-
-        LOG.info("Start validating table configurations");
+        LOG.info("Start initialization process");
         List<TableInfo> tableInfos;
         try {
             tableInfos = tableConfigs.stream().map(this::getTableInfo).collect(Collectors.toList());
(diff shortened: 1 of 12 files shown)


data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/DynamoDBService.java
```diff
@@ -129,15 +129,15 @@ public void shutdown() {
      * Future optimization can be done to accept configuration changes
      */
     public void init() {
-        LOG.info("Start initialize DynamoDB service");
+        LOG.info("Try to initialize DynamoDB service");
```
graytaylor0, line 132: A lot of these logs really do belong in DEBUG level. Can you make them DEBUG in a future PR (once more thorough testing is done with the verbose logs)?

graytaylor0 (approved):

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/DataFileScheduler.java
```diff
@@ -31,8 +31,15 @@ public class DataFileScheduler implements Runnable {
...
-    private static final int DEFAULT_LEASE_INTERVAL_MILLIS = 30_000;
+    /**
+     * Maximum concurrent data loader per node
+     */
+    private static final int MAX_JOB_COUNT = 4;
+
+    /**
+     * Default interval to acquire a lease from coordination store
+     */
+    private static final int DEFAULT_LEASE_INTERVAL_MILLIS = 15_000;
```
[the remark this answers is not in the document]
daixba: Thanks Taylor, I do have those in my code at the very beginning but I removed it, like at first 15 seconds, then 30/60 seconds if no lease taken. Reset to 15 seconds if lease was taken.  Set it to 15 seconds minimum is so that other nodes can also take leases and the job can be evenly distributed.   The main backoff I put is the GetRecords call (which is to check the number of records/bytes returned), but it actually added complexity and increased latency.  So there is a trade off.  

But this is to say it still can be added in the future once we have done more testing.

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/DynamoDBService.java
daixba: Happy to do it.

dinujoh (approved):

PR #3554, opened 2023-10-27
Catch exceptions and backoff and retry ddb source threads instead of shutting down on exception

graytaylor0:
### Description
The three Scheduler threads in the ddb source, `ExportScheduler`, `StreamScheduler`, and `DataFileScheduler`, may shutdown silently if an exception is thrown during processing.

* Catch and log exceptions before backing off and retrying in these scheduler threads
* Fix `primary_key` attribute to divide partition and sort key by `|` to avoid conflicts (since `|` is an invalid ddb character for partition and sort keys, and is valid as OpenSearch document id

 
### Check List

### DIFF
--- a/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/DataFileScheduler.java
+++ b/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/DataFileScheduler.java
@@ -88,23 +88,32 @@ public class DataFileScheduler implements Runnable {
         LOG.debug("Start running Data File Scheduler");
 
         while (!Thread.currentThread().isInterrupted()) {
-            if (numOfWorkers.get() < MAX_JOB_COUNT) {
-                final Optional<EnhancedSourcePartition> sourcePartition = coordinator.acquireAvailablePartition(DataFilePartition.PARTITION_TYPE);
+            try {
+                if (numOfWorkers.get() < MAX_JOB_COUNT) {
+                    final Optional<EnhancedSourcePartition> sourcePartition = coordinator.acquireAvailablePartition(DataFilePartition.PARTITION_TYPE);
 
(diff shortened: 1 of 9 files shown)


dlvenable (approved):

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/DataFileScheduler.java
```diff
@@ -88,23 +88,32 @@ public void run() {
...
+                if (numOfWorkers.get() < MAX_JOB_COUNT) {
+                    final Optional<EnhancedSourcePartition> sourcePartition = coordinator.acquireAvailablePartition(DataFilePartition.PARTITION_TYPE);
+
+                    if (sourcePartition.isPresent()) {
+                        activeExportS3ObjectConsumersGauge.incrementAndGet();
+                        DataFilePartition dataFilePartition = (DataFilePartition) sourcePartition.get();
+                        processDataFilePartition(dataFilePartition);
+                        activeExportS3ObjectConsumersGauge.decrementAndGet();
+                    }
+                }
```
dinujoh, line 101: should we log when workers are not available ? Will this be too much logging ?

dinujoh (approved):

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/DataFileScheduler.java
graytaylor0: I think it will be too much logging yeah. These threads continue running forever, so you'd see lots of logging even after all the data files are processed

dinujoh: We can consider emitting metrics when threads are not available.

PR #3575, opened 2023-11-01
Add acknowledgments for the ddb source

### LINKED ISSUE #3538: End to end acknowledgments for Dynamo source
Currently, the s3 and opensearch source have an internal acknowledgment timeout of 1 or 2 hours. This is because these are mainly used for historical migrations of data,  so they are less time sensitive. However, as dynamodb streams is a streaming use case where order as well as any delay really matters, it is not as simple to provide a high default timeout like these other sources.

The simple approach would be to just make the timeout configurable for the user with an `acknowledgment_timeout` parameter that specifies the timeout, but this is not ideal either since this value can be hard to gauge for streams on their own (also with OpenSearch backpressure), and would still require a fairly large default value to make sure that timeouts don't occur too early right before the data gets to OpenSearch. We need a way to lower the `acknowledgment_timeout` to quickly recover for streams, but to do that we need to lower the amount of data, since a shard could take a long time to process, which means a longer time to indicate a failure and timeout (and save progress instead of restarting the entire shard when the last part of it has an acknowledgment timeout).

To handle these complications, I would propose that we chunk the `AcknowledgmentSets` for the dynamo source to a general amount of data (in bytes). We can do this by grouping by sequence numbers for a DDB stream shard. For example, we have a configurable 

```
acknowledgment_checkpoint_size: "10mb"
acknowledgment_timeout: "30s"
```
which will keep reading from the shard until it hits a sequence number that goes to (or past this point, we don't care if it's a little overestimated). Now when we receive an acknowledgment for one of these chunks (we should receive them in order), we can update the partition ownership timeout for that partition to be something small (potentially as low as 30 seconds, or however long the `acknowledgment_checkpoint_size` takes to travel through the pipeline (#3494 could be used to tune these values)but it is wise for us to allow this to be configurable as well since there is always sink backpressure. For a well scaled sink, low values for `acknowledgment_timeout` may be possible, which would allow nodes to pick up on crashes very quickly and continue processing the shard.

When an acknowledgment callback is received, not only do we update the ownership timeout with `acknowledgment_timeout` amount of time, we also update the partition progress state of the partition with the last sequence number that we received an ack callback for. This will allow us to pick up right where we left off in the case of failure instead of having to restart the processing of the entire shard.

**Additional context**

graytaylor0:
### Description
This change adds end to end acknowledgments to the dynamodb source. This includes configurable acknowledgment timeouts for both shards and data files. 

the new parameters are

```
Boolean to enable acknowledgments
acknowledgments:  true
```

```
 Duration to timeout before data file is reprocessed (from the time the last line of data file is written to the buffer to the time that it is reached by the sink (Data Prepper duration format)).

s3_data_file_acknowledgment_timeout: "30s" // Default value is 5 minutes (is this too high?)
```

```
 Duration to timeout before shard is reprocessed (from the time the last record in the shard is written to buffer till it is reached by the sink (Data Prepper duration format)

shard_acknowledgment_timeout: "30s" // Default value is 3 minutes (is this too high?)
```

Additionally, another parameter was added in this PR for the `s3_region` for the export. The default will be to use the same region for the bucket as the `aws.region` for sts, but this can be set to something different now with the `s3_region` parameter. Setting this region on the client fixes an issue where `GetObject` requests would hang infinitely in some scenarios (for unknown reasons)
 
### Issues Resolved
Resolves #3538 
 
### Check List

### DIFF
--- a/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/configuration/ExportConfig.java
+++ b/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/configuration/ExportConfig.java
@@ -7,14 +7,19 @@ package org.opensearch.dataprepper.plugins.source.dynamodb.configuration;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import jakarta.validation.constraints.NotBlank;
+import software.amazon.awssdk.regions.Region;
 
 public class ExportConfig {
 
     @JsonProperty("s3_bucket")
     @NotBlank(message = "Bucket Name is required for export")
     private String s3Bucket;
+
     @JsonProperty("s3_prefix")
     private String s3Prefix;
+
+    @JsonProperty("s3_region")
+    private String s3Region;
 
     public String getS3Bucket() {
         return s3Bucket;
@@ -23,4 +28,9 @@ public class ExportConfig {
     public String getS3Prefix() {
         return s3Prefix;
     }
+
+    public Region getAwsRegion() {
+        return s3Region != null ? Region.of(s3Region) : null;
+    }
+
 }
--- a/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/converter/RecordConverter.java
+++ b/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/converter/RecordConverter.java
@@ -6,6 +6,7 @@
 package org.opensearch.dataprepper.plugins.source.dynamodb.converter;
 
 import org.opensearch.dataprepper.buffer.common.BufferAccumulator;
+import org.opensearch.dataprepper.model.acknowledgements.AcknowledgementSet;
 import org.opensearch.dataprepper.model.event.Event;
 import org.opensearch.dataprepper.model.event.EventMetadata;
 import org.opensearch.dataprepper.model.event.JacksonEvent;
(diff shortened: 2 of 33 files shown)


data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/configuration/ExportConfig.java
```diff
@@ -7,20 +7,30 @@
...
 public class ExportConfig {
 
     @JsonProperty("s3_bucket")
     @NotBlank(message = "Bucket Name is required for export")
     private String s3Bucket;
+
     @JsonProperty("s3_prefix")
     private String s3Prefix;
 
+    @JsonProperty("s3_region")
```
daixba, line 21: Not really an issue, but just want to point out that I have never tested whether it can support s3 in a different region or not. Even it can, it will be strange since ddb and opensearch are in the same region, what is the point to write or read data files from another region.

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/converter/RecordConverter.java
```diff
@@ -91,12 +92,15 @@ public void addToBuffer(Map<String, Object> data, Map<String, Object> keys, long
             eventMetadata.setAttribute(PRIMARY_KEY_DOCUMENT_ID_METADATA_ATTRIBUTE, partitionKey);
         }
         bufferAccumulator.add(new Record<>(event));
+        if (acknowledgementSet != null) {
+            acknowledgementSet.add(event);
```
daixba, line 96: I didn't fully understand how this works, but will this have any memory issue? I only see there is a add of event, will that be a clear of event somewhere else.

daixba (approved):

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/configuration/ExportConfig.java
graytaylor0: Yeah I don't know but you never know. By default this can be ignored for same region though, so it can't really hurt to have

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/converter/RecordConverter.java
graytaylor0: This just adds an `eventHandle` to the acknowledgment set, and this takes up very small amount of memory. @kkondaka can explain more

kkondaka (approved):

PR #3600, opened 2023-11-08
Add splitExportTraceServiceRequest API to OTelProtoDecoder

kkondaka:
### Description
Add splitExportTraceServiceRequest API to OTelProtoDecoder.
This API splits ExportTraceServiceRequest to multiple ExportTraceServiceRequests, one per unique traceId. After the split, each ExportTraceServiceRequests will contain only spans belonging to same traceId.
 
### Issues Resolved
 
### Check List


chenqi0805 (dismissed):

dlvenable (changes requested):
I'd like to reconsider our approach here.

As I understand it, the goal of this is to route different traces in Kafka rather than have to forward it over HTTP in the peer-forwarder later.

As we are having to parse the binary and then recreate, I think we can improve this approach by just saving Events to the Kafka buffer. We can take any event or collection of events and then serialize those.

This approach would also solve the problem for other sources so that they can work with the `kafka` buffer without having to update to use `byte[]`.

dlvenable (dismissed):
We discussed offline and agreed to proceed with this approach for now. It is simple enough and solves the current problem.

graytaylor0 (approved):

PR #3724, opened 2023-11-29
Add aggregate metrics for ddb source export and stream

graytaylor0:
### Description
This change adds the following aggregate metrics (metrics without a `sub-pipeline` prefix) to dynamoDB source pipelines

* `stream5xxErrors` - The number of errors encountered when calling stream APIs (`getRecords`, `getShardIterator`)

* `streamApiInvocations` - The number of invocations of stream APIs

* `export5xxErrors` - The number of errors encountered when calling export APIs (`exportTableToPointInTime`, `describeExport`)

* `exportApiInvocations` - The number of invocations of export APIs

 
### Check List

### DIFF
--- a/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/ExportTaskManager.java
+++ b/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/ExportTaskManager.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.dataprepper.plugins.source.dynamodb.export;
 
+import org.opensearch.dataprepper.plugins.source.dynamodb.utils.DynamoDBSourceAggregateMetrics;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.core.exception.SdkException;
@@ -14,6 +15,7 @@ import software.amazon.awssdk.services.dynamodb.model.DescribeExportResponse;
 import software.amazon.awssdk.services.dynamodb.model.ExportFormat;
 import software.amazon.awssdk.services.dynamodb.model.ExportTableToPointInTimeRequest;
 import software.amazon.awssdk.services.dynamodb.model.ExportTableToPointInTimeResponse;
+import software.amazon.awssdk.services.dynamodb.model.InternalServerErrorException;
 import software.amazon.awssdk.services.dynamodb.model.S3SseAlgorithm;
 
 import java.time.Instant;
@@ -25,10 +27,12 @@ public class ExportTaskManager {
     private static final ExportFormat DEFAULT_EXPORT_FORMAT = ExportFormat.ION;
 
     private final DynamoDbClient dynamoDBClient;
+    private final DynamoDBSourceAggregateMetrics dynamoAggregateMetrics;
 
-
-    public ExportTaskManager(DynamoDbClient dynamoDBClient) {
+    public ExportTaskManager(final DynamoDbClient dynamoDBClient,
+                             final DynamoDBSourceAggregateMetrics dynamoAggregateMetrics) {
         this.dynamoDBClient = dynamoDBClient;
+        this.dynamoAggregateMetrics = dynamoAggregateMetrics;
     }
 
     public String submitExportJob(String tableArn, String bucket, String prefix, String kmsKeyId, Instant exportTime) {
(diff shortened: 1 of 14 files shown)


data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/ExportTaskManager.java
```diff
@@ -64,9 +73,13 @@ public String getExportManifest(String exportArn) {
 
         String manifestKey = null;
         try {
+            dynamoAggregateMetrics.getExportApiInvocations().increment();
             DescribeExportResponse resp = dynamoDBClient.describeExport(request);
             manifestKey = resp.exportDescription().exportManifest();
 
+        } catch (final InternalServerErrorException e) {
+            dynamoAggregateMetrics.getExport5xxErrors().increment();
+            LOG.error("Unable to get manifest file for export " + exportArn);
```
travisbenedict, line 82: Should we log the error here too? Same applies for the `catch` section below

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/leader/ShardManager.java
```diff
@@ -148,22 +153,30 @@ private List<Shard> listShards(String streamArn, String lastEvaluatedShardId) {
...
+                // at least one more page of shard IDs to retrieve
+                lastEvaluatedShardId = describeStreamResult.streamDescription().lastEvaluatedShardId();
 
-        } while (lastEvaluatedShardId != null);
+
+            } while (lastEvaluatedShardId != null);
+        } catch(final InternalServerErrorException e) {
+            LOG.error("Received an internal server exception from DynamoDB while listing shards: {}", e.getMessage());
+            dynamoDBSourceAggregateMetrics.getStream5xxErrors().increment();
+            return shards;
```
travisbenedict, line 178: Should we rethrow the error here instead of potentially returning partial shards?

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/ExportTaskManager.java
graytaylor0: Ah yeah we should log that

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/leader/ShardManager.java
graytaylor0: I don't think there's anything wrong with returning partial shards, it just means we will only check those shards for child shards. The next time we run `describeStream` we will still go through the same shards again and create child shards for them (even the ones we return here) until 24 hours is up (when the shard would expire).

travisbenedict (approved):

travisbenedict (approved):

chenqi0805 (approved):

data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/export/ExportTaskManager.java
```diff
@@ -46,12 +50,17 @@ public String submitExportJob(String tableArn, String bucket, String prefix, Str
...
         try {
+            dynamoAggregateMetrics.getExportApiInvocations().increment();
             ExportTableToPointInTimeResponse response = dynamoDBClient.exportTableToPointInTime(req);
 
             String exportArn = response.exportDescription().exportArn();
             String status = response.exportDescription().exportStatusAsString();
             LOG.debug("Export Job submitted with ARN {} and status {}", exportArn, status);
             return exportArn;
+        } catch (final InternalServerErrorException e) {
+            dynamoAggregateMetrics.getExport5xxErrors().increment();
```
dinujoh, line 61: should we also consider adding pipeline specific metric ?

graytaylor0: I don't think there's any value in that for these metrics at least. There may be other pipeline specific metrics we can add, but that can be a future PR

PR #4002, opened 2024-01-23
Added when condition and fixed building reader on each event

asifsmohammed:
### Description
- Adds when condition to geoip configuration
- Fixed DatabaseReader being built for each event which affects the performance and cache will not be used
- Adds default target as "geolocation"
- Moved `databasedownload` package to `extension`
 
### Check List


dlvenable (changes requested):

data-prepper-plugins/geoip-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/extension/databasedownload/DatabaseReaderCreateTest.java
```diff
@@ -0,0 +1,43 @@
...
+class DatabaseReaderCreateTest {
+    @Mock
+    private Path path;
+
+    @Test
+    void createLoaderTest() throws IOException {
+        final String testFileURL = "https://github.com/maxmind/MaxMind-DB/raw/main/test-data/GeoLite2-City-Test.mmdb";
+        final File file = File.createTempFile( "GeoIP2-City-Test", ".mmdb");
+
+        final BufferedInputStream in = new BufferedInputStream(new URL(testFileURL).openStream());
```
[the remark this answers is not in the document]
asifsmohammed: I have tried it with mock URL but looks like the DatabaseReader needs an actual mmdb file, it's validating some metadata. So I decided to use actual test mmdb file. Only way to avoid this is not to test the DatabaseReader build.

dlvenable (approved):

data-prepper-plugins/geoip-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/extension/GeoIPProcessorService.java
```diff
@@ -41,19 +37,19 @@ public class GeoIPProcessorService {
     private static final Logger LOG = LoggerFactory.getLogger(GeoIPProcessorService.class);
     public static final String DATABASE_1 = "first_database_path";
     public static final String DATABASE_2 = "second_database_path";
-    private static final String TEMP_PATH_FOLDER = "GeoIP";
-    private GeoIPProcessorConfig geoIPProcessorConfig;
-    private LicenseTypeOptions licenseType;
+    private static final String TEMP_PATH_FOLDER = "geoip";
```
oeyh, line 40: Should this instance of "GeoIP" be changed to "geoip" as well:
https://github.com/opensearch-project/data-prepper/blob/2be8166ac8ac1f80784201f2a6e33302145cb5a8/data-prepper-plugins/geoip-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/databasedownload/DBSource.java#L29
 Looks to me the two are referring to the same directory.

data-prepper-plugins/geoip-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/GeoIPProcessorConfig.java
```diff
@@ -44,4 +47,11 @@ public List<String> getTagsOnFailure() {
         return tagsOnFailure;
     }
 
+    /**
+     * Get when condition
+     * @return String When condition
+     */
+    public String getWhenCondition() {
```
oeyh, line 54: Does it make sense to add this when condition to the entry level? Just checking. I'm asking because processors like `add_entries` has it on the entry level: https://github.com/opensearch-project/data-prepper/blob/2be8166ac8ac1f80784201f2a6e33302145cb5a8/data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorConfig.java#L31

data-prepper-plugins/geoip-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/extension/GeoIPProcessorService.java
asifsmohammed: Yeah looks like it's hard cded here, I will remove it in next PR. This won't be used in upcoming change

data-prepper-plugins/geoip-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/GeoIPProcessorConfig.java
asifsmohammed: That's a good suggestion, I actually looked into it if we need it but I think it's worth making that change. I will move it next PR so it can be configured for each entry.

oeyh (approved):

PR #4108, opened 2024-02-12
Catch exception instead of shutting down in date processor

### LINKED ISSUE #4107: [BUG] Date Processor throws Exception and skips processing the remainder of a batch of Events
**Describe the bug**
The date processor does not catch any Exceptions in processing. This means that the pipeline will be shut down (in the current state of `data-prepper-core`), and will skip the processing of the remainder of Events in a batch once an Exception is hit. 

**To Reproduce**
1. Create a pipeline with date processor and a `date_when`
2. Send an Event that will trigger an exception evaluating the conditional statement against that Event
3. Observe the exception get thrown and not caught by the date processor.

**Expected behavior**

**Screenshots**

 - OS: [e.g. Ubuntu 20.04 LTS]

**Additional context**

graytaylor0:
### Description
This change makes it so the date processor hitting an exception does not shut down the pipeline, and instead catches the exception and continues on to the next Event, instead of skipping the remainder of the Events in a batch
 
### Issues Resolved
Resolves #4107 
 
### Check List

### DIFF
--- a/data-prepper-plugins/date-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/date/DateProcessor.java
+++ b/data-prepper-plugins/date-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/date/DateProcessor.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.dataprepper.plugins.processor.date;
 
+import io.micrometer.core.instrument.Counter;
+import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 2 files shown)


dlvenable (changes requested):
Please add some unit tests. One simple one would be to mock the ExpressionEvaluator to throw a RuntimeException on the first call and then return true on the second. Verify that the second event is processed.

dlvenable (approved):
Thanks!

PR #4124, opened 2024-02-13
Fix bug where s3 scan could skip when lastModifiedTimestamps are the same

### LINKED ISSUE #4123: [BUG] S3 Scan has potential to filter out objects with the same timestamp
**Describe the bug**
Scheduled S3 Scan filters out objects by checking a global state item that contains the timestamp of the most recent object that was processed during the previous scan. (https://github.com/opensearch-project/data-prepper/blob/503b7741fe63596ad994198e513404c1143f9e1d/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanPartitionCreationSupplier.java#L168)

This has the potential to skip objects that are missed in the scan, but created with the same timestamp.

**Expected behavior**
Do not filter out objects that have a timestamp equal to the last modified timestamp

**Additional context**

graytaylor0:
### Description
Provides some overlap of a second when filtering out objects based on lastmodified timestamp for subsequent scans. This filtering is done to lower the amount of lookups required to source coordination, and will not result in duplicate processing if there are any objects that are returned to source coordinator twice.
 
### Issues Resolved
Resolves #4123 
 
### Check List


data-prepper-plugins/dynamodb-source/README.md
asifsmohammed, line 1: Thanks for fixing this!

asifsmohammed (approved):

oeyh (approved):

PR #4185, opened 2024-02-26
Create file on shutdown to indicate that data prepper is shutdown

graytaylor0:
### Description
This change creates a file named `Is-Shutdown.log` when data prepper is shutdown. This can be used as an alternative method to check if a data prepper container has shutdown.

This was tested with a Docker volume to ensure this file was created successfully.
 
### Check List

### DIFF
--- a/data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/server/ShutdownHandler.java
+++ b/data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/server/ShutdownHandler.java
@@ -14,6 +14,9 @@ import org.slf4j.LoggerFactory;
 import javax.ws.rs.HttpMethod;
 import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 /**
  * HttpHandler to handle requests to shut down the data prepper instance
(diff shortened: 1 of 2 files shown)


data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/server/ShutdownHandler.java
```diff
@@ -22,6 +25,9 @@ public class ShutdownHandler implements HttpHandler {
     private final DataPrepper dataPrepper;
     private static final Logger LOG = LoggerFactory.getLogger(ShutdownHandler.class);
 
+    static final Path SHUTDOWN_FILE_PATH = Path.of("Is-Shutdown.log");
```
graytaylor0, line 28: What should the path here be? I noticed that when running in docker the data prepper directory structure had a `log/data-prepper/` directory which could make sense, but I am not sure how to make it consistent for running locally and with docker.

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/server/ShutdownHandler.java
```diff
@@ -44,6 +50,7 @@ public void handle(final HttpExchange exchange) throws IOException {
             exchange.sendResponseHeaders(HttpURLConnection.HTTP_INTERNAL_ERROR, 0);
         } finally {
             exchange.getResponseBody().close();
+            Files.write(SHUTDOWN_FILE_PATH, SHUTDOWN_MESSAGE.getBytes(StandardCharsets.UTF_8));
             dataPrepper.shutdownServers();
```
graytaylor0, line 54: I chose to create the file before the shutdownServer call because otherwise there is a chance for the File write to be interrupted (since we are shutting down the server which is where the ShutdownHandler exists.

kkondaka (approved):

asuresh8 (approved):

dlvenable (changes requested):
This is very specific behavior being added into the general shutdown code. And it could lead to confusion when Data Prepper is not run in a container.

I recommend taking one of two alternatives.
1. Create a shutdown hook that you can hook into in different code which writes the file.
2. Update the run script to create a PID file. You can then read that file to tell if the process is shutdown.

I think we should close this PR in favor of a more general approach.

graytaylor0:
Closing in favor of https://github.com/opensearch-project/data-prepper/pull/4189 based on @dlvenable's comments

PR #4190, opened 2024-02-27
Set default value of flattenAttributes to true in Otel metrics source

kkondaka:
### Description
Set default value of flattenAttributes to true in Otel metrics source.
Modify the OTEL metrics source to create events/records with flatten attributes set to true.
If OTEL processor config changes the flatten attributes to flase, then change it false in the processor.

OTEL metrics always have been implemented with flattening the attributes(ie fields inside "attributes" field are moved to the root). We have added option to not flatten some time back. But that's not used by anyone (that I know of). When the previous PR which moved the creation of OTEL metrics records to OTEL source (instead of otel processor) I picked the default to false and it is changed to true in the processor. But that was unnecessary also hit a bug due to valid characters in a key. This new approach will avoid it.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/otel-metrics-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otelmetrics/OTelMetricsRawProcessor.java
+++ b/data-prepper-plugins/otel-metrics-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otelmetrics/OTelMetricsRawProcessor.java
@@ -14,7 +14,6 @@ import static org.opensearch.dataprepper.model.metric.JacksonExponentialHistogra
 import static org.opensearch.dataprepper.model.metric.JacksonHistogram.BUCKETS_KEY;
 import org.opensearch.dataprepper.model.metric.JacksonMetric;
 import org.opensearch.dataprepper.model.metric.Metric;
-import static org.opensearch.dataprepper.model.metric.JacksonMetric.ATTRIBUTES_KEY;
 import org.opensearch.dataprepper.model.processor.AbstractProcessor;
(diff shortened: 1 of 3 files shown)


data-prepper-plugins/otel-metrics-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otelmetrics/OTelMetricsRawProcessor.java
```diff
@@ -85,7 +81,7 @@ public Collection<Record<? extends Metric>> doExecute(Collection<Record<?>> reco
         for (Record<?> rec : records) {
             if ((rec.getData() instanceof Event)) {
                 Record<? extends Metric> newRecord = (Record<? extends Metric>)rec;
-                if (otelMetricsRawProcessorConfig.getFlattenAttributesFlag() ||
+                if (!otelMetricsRawProcessorConfig.getFlattenAttributesFlag() ||
```
graytaylor0, line 84: Can you explain why this is being negated?

kkondaka: It's true by default. Only when it is set to false in the config, we call `JacksonMetric.setFlattenAttributes(false)`

graytaylor0 (approved):

dlvenable (commented):
Aren't these changes going to break existing pipelines? This makes the data fundamentally different.

dlvenable (approved):

PR #4199, opened 2024-02-28
Fix to set kafka source truststore path and password only when they are not null value

dinujoh:
Fix to set kafka source truststore path and password only when they are not null value

### Check List


kkondaka (approved):

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/util/KafkaSecurityConfigurer.java
```diff
@@ -129,7 +129,8 @@ private static void setPlainTextAuthProperties(final Properties properties, fina
     private static void setSecurityProtocolSSLProperties(final Properties properties, final EncryptionConfig encryptionConfig) {
         if (Objects.nonNull(encryptionConfig.getCertificateContent())) {
             setCustomSslProperties(properties, encryptionConfig.getCertificateContent());
-        } else {
+        } else if (Objects.nonNull(encryptionConfig.getTrustStoreFilePath()) &&
```
hshardeesi, line 132: is trusttore config valid if user does not provide certificate content?

hshardeesi (approved):

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/util/KafkaSecurityConfigurer.java
dinujoh: Yes, they are mutually exclusive.

PR #4346, opened 2024-03-27
Add creation and aggregation of dynamic S3 groups based on events

graytaylor0:
### Description
This change adds a concept of `S3Groups` to the S3 Sink. Groups will be created dynamically based on the Events that come into the sink, and will be sent to S3 as a single object when the thresholds are reached for that group.

This PR only adds support for a dynamic `path_prefix`. The support for exposing `object_name` as a dynamic expression will be added in a future PR

Future Planned PRs

* Refactor the buffer tracking of thresholds to the S3Group tracking this to reduce deduplication
* Support `aggregate_threshold` for flushing the largest groups when the threshold is reached
* Support injection of `getEpochNanos()` and `getRandomUUID()` into the dynamic expressions
 
### Issues Resolved
Related to #4345 

Ran the s3 sink integration tests successfully
 
### Check List

### DIFF
--- a/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluator.java
+++ b/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluator.java
@@ -11,6 +11,8 @@ import org.opensearch.dataprepper.model.event.JacksonEvent;
 
 import javax.inject.Inject;
 import javax.inject.Named;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Public class that {@link org.opensearch.dataprepper.model.processor.Processor},
@@ -73,5 +75,58 @@ class GenericExpressionEvaluator implements ExpressionEvaluator {
             fromIndex = endPosition + 1;
         }
         return true;
+    }
+
+    @Override
+    public List<String> extractDynamicKeysFromFormatExpression(final String format) {
+        final List<String> formatExpressionKeys = new ArrayList<>();
+
+        if (format == null) {
+            return formatExpressionKeys;
+        }
+
+        int fromIndex = 0;
+        int position = 0;
+        while ((position = format.indexOf("${", fromIndex)) != -1) {
+            int endPosition = format.indexOf("}", position + 1);
+            if (endPosition == -1) {
+                return formatExpressionKeys;
+            }
+            String name = format.substring(position + 2, endPosition);
+
+            if (JacksonEvent.isValidEventKey(name)) {
+                if (!name.startsWith("/")) {
+                    name = "/" + name;
+                }
+                formatExpressionKeys.add(name);
+            }
+            fromIndex = endPosition + 1;
+        }
+        return formatExpressionKeys;
+    }
+
+    @Override
+    public List<String> extractDynamicExpressionsFromFormatExpression(final String format) {
+        final List<String> dynamicExpressionStatements = new ArrayList<>();
+
+        if (format == null) {
+            return dynamicExpressionStatements;
+        }
+
+        int fromIndex = 0;
+        int position = 0;
+        while ((position = format.indexOf("${", fromIndex)) != -1) {
+            int endPosition = format.indexOf("}", position + 1);
+            if (endPosition == -1) {
+                return dynamicExpressionStatements;
+            }
+            String name = format.substring(position + 2, endPosition);
+
+            if (!JacksonEvent.isValidEventKey(name) && isValidExpressionStatement(name)) {
+                dynamicExpressionStatements.add(name);
+            }
+            fromIndex = endPosition + 1;
+        }
+        return dynamicExpressionStatements;
     }
 }
--- a/data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluatorTest.java
+++ b/data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluatorTest.java
@@ -20,13 +20,9 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.opensearch.dataprepper.model.event.Event;
 
-import java.util.Calendar;
 import java.util.List;
 import java.util.Random;
 import java.util.UUID;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import static org.hamcrest.CoreMatchers.equalTo;
@@ -178,7 +174,9 @@ class GenericExpressionEvaluatorTest {
             return Stream.of(
                     arguments("test-${foo}-${bar}", List.of("/foo", "/bar")),
                     arguments("test-${getMetadata(\"key\"}-${/test}", List.of("/test")),
-                    arguments("test-format", List.of())
+                    arguments("test-format", List.of()),
+                    arguments("test-${/test", List.of()),
+                    arguments(null, List.of())
             );
         }
     }
@@ -188,7 +186,9 @@ class GenericExpressionEvaluatorTest {
         public Stream<? extends Arguments> provideArguments(final ExtensionContext context) {
             return Stream.of(
                     arguments("test-${foo}-${bar}", List.of()),
-                    arguments("test-${getMetadata(\"key\")}-${/test}", List.of("getMetadata(\"key\")"))
+                    arguments("test-${getMetadata(\"key\")}-${/test}", List.of("getMetadata(\"key\")")),
+                    arguments("test-${/test", List.of()),
+                    arguments(null, List.of())
             );
         }
     }
--- a/data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluatorTest.java
+++ b/data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluatorTest.java
@@ -8,7 +8,11 @@ package org.opensearch.dataprepper.expression;
 import org.antlr.v4.runtime.tree.ParseTree;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.provider.ArgumentsSource;
 import org.junit.jupiter.params.provider.CsvSource;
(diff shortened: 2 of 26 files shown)


data-prepper-plugins/s3-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/s3/S3Sink.java
```diff
@@ -82,13 +86,27 @@ public S3Sink(final PluginSetting pluginSetting,
...
         S3OutputCodecContext s3OutputCodecContext = new S3OutputCodecContext(OutputCodecContext.fromSinkContext(sinkContext), compressionOption);
 
         codec.validateAgainstCodecContext(s3OutputCodecContext);
 
-        s3SinkService = new S3SinkService(s3SinkConfig, bufferFactory, codec, s3OutputCodecContext, s3Client, keyGenerator, Duration.ofSeconds(5), pluginMetrics);
+        final S3GroupIdentifierFactory s3GroupIdentifierFactory = new S3GroupIdentifierFactory(keyGenerator);
+        final S3GroupManager s3GroupManager = new S3GroupManager(s3SinkConfig, s3GroupIdentifierFactory, bufferFactory, s3Client);
+
+
+        s3SinkService = new S3SinkService(s3SinkConfig, codec, s3OutputCodecContext, s3Client, keyGenerator, Duration.ofSeconds(5), pluginMetrics, s3GroupManager);
```
dinujoh, line 109: nit: what is the duration ? Can you move that to constant ?

data-prepper-plugins/s3-sink/src/test/java/org/opensearch/dataprepper/plugins/sink/s3/grouping/S3GroupIdentifierTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+
+import org.junit.jupiter.api.Test;
+
+import java.util.UUID;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class S3GroupIdentifierTest {
```
dlvenable, line 16: We should test `hashCode()` as well.

data-prepper-plugins/s3-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/s3/S3SinkService.java
```diff
@@ -135,9 +138,20 @@ void output(Collection<Record<Event>> records) {
...
 
-                flushToS3IfNeeded();
+                final boolean flushed = flushToS3IfNeeded(currentBuffer);
+
+                if (flushed) {
+                    s3GroupManager.removeGroup(s3GroupEntry.getKey());
+                }
+            }
+
+            for (final Map.Entry<S3GroupIdentifier, S3Group> s3GroupEntry : s3GroupManager.getS3GroupEntries()) {
```
dinujoh, line 148: should this be flushed when global threshold is reached ?

dlvenable (changes requested):

data-prepper-plugins/s3-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/s3/S3SinkService.java
graytaylor0: The global threshold will be handled in a follow-up PR, and will be configured separately as `aggregate_threshold`

data-prepper-plugins/s3-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/s3/grouping/S3GroupIdentifierFactory.java
```diff
@@ -0,0 +1,27 @@
...
+    private final KeyGenerator keyGenerator;
+
+    public S3GroupIdentifierFactory(final KeyGenerator keyGenerator) {
+        this.keyGenerator = keyGenerator;
+     }
+
+
+    public S3GroupIdentifier getS3GroupIdentifierForEvent(final Event event) {
+
+        final String groupIdentificationHash = keyGenerator.generateKeyForEvent(event, false);
```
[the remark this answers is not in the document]
graytaylor0: I think this would have to be in `ExpressionEvaluator` since it would also need to run the expressions to create the values of the Map.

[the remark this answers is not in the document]
graytaylor0: Or are you thinking that expression evaluator would support a function to extract all of the `${}` sections out of a format expression, and then each of these individual expressions would need to be run separately to get the values? Or can we just have expression evaluator return the full Map result?

data-prepper-plugins/s3-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/s3/S3SinkService.java
dlvenable: I believe that @graytaylor0 will submit a follow-on to support the `aggregate_threshold` to do just that.

data-prepper-plugins/s3-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/s3/grouping/S3GroupIdentifierFactory.java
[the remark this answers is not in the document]
dlvenable: > Or are you thinking that expression evaluator would support a function to extract all of the ${} sections out of a format expression, and then each of these individual expressions would need to be run separately to get the values?

This is what I was thinking. And they should come out in a normalized manner. That is, all have the leading `/`. This way you always get a consistent value.

> Or can we just have expression evaluator return the full Map result?

This could be an option as well. But, it seems a little more specific to this use-case. Getting the list of keys may be more extensible.

data-prepper-plugins/s3-sink/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/s3/S3SinkIT.java
```diff
@@ -45,6 +45,7 @@
 import software.amazon.awssdk.services.s3.model.ListObjectsV2Request;
```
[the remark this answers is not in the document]
graytaylor0: Is there a way for me to create an actual `GenericExpressionEvaluator` for the integration tests? Or can I verify this while mocking `ExpressionEvaluator`?

data-prepper-plugins/s3-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/s3/grouping/S3GroupIdentifierFactory.java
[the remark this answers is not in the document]
graytaylor0: >  This is what I was thinking. And they should come out in a normalized manner. That is, all have the leading /. This way you always get a consistent value.

So in this case, the S3 sink would have to explicitly check for keys and other expressions, and would have to check the Event itself for the values and run the expression evaluation functions individually as well

data-prepper-plugins/s3-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/s3/S3Sink.java
graytaylor0: This was already here. Looks like the variable is called `retrySleepTime`. I can make it a constant

dlvenable (dismissed):

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluator.java
```diff
@@ -74,4 +76,57 @@ public Boolean isValidFormatExpression(final String format) {
...
+
+        if (format == null) {
+            return formatExpressionKeys;
+        }
+
+        int fromIndex = 0;
+        int position = 0;
+        while ((position = format.indexOf("${", fromIndex)) != -1) {
+            int endPosition = format.indexOf("}", position + 1);
+            if (endPosition == -1) {
```
dinujoh, line 92: This might be a user error in dynamic expression. Looks like we are ignoring any errors ?

data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluatorTest.java
```diff
@@ -142,5 +150,47 @@ void isValidFormatExpressionsReturnsFalseWhenIsValidKeyAndValidExpressionIsFalse
...
+    void extractDynamicKeysFromFormatExpression_returns_expected_result(final String formatExpression, final List<String> expectedDynamicKeys) {
+        final List<String> result = statementEvaluator.extractDynamicKeysFromFormatExpression(formatExpression);
+
+        assertThat(result, notNullValue());
+        assertThat(result.equals(expectedDynamicKeys), equalTo(true));
+    }
+
+    @ParameterizedTest
+    @ArgumentsSource(FormatExpressionsToExtractedDynamicExpressionsArgumentProvider.class)
+    void extractDynamicExpressionsFromFormatExpression_returns_expected_result(final String formatExpression, final List<String> expectedDynamicExpressions) {
```
dinujoh, line 164: can we add a negative test where invalid expression is passed ?

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluator.java
graytaylor0: This method is not responsible for catching invalid expressions, it will just parse out the valid ones. If you think it should throw, then it can here but it's job is not to validate, just to parse and return

data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluatorTest.java
graytaylor0: There is a negative case where invalid expression syntax is found, it will just not be added to the list. This method is just supposed to parse out valid existing expressions

dinujoh (approved):

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/GenericExpressionEvaluator.java
graytaylor0: Will add some unit tests for validating the expression on startup of the S3 sink in the next PR

dinujoh: We can add a test case to cover validation on startup for s3 sink constructor as discussed offline.

PR #4437, opened 2024-04-18
Make s3 partition size configurable and add unit test for S3 partition creator classes

dinujoh:
Make s3 partition size configurable and add unit test for S3 partition creator classes

 
### Check List

### DIFF
--- a/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/configuration/CollectionConfig.java
+++ b/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/configuration/CollectionConfig.java
@@ -9,14 +9,16 @@ import java.util.stream.Collectors;
 public class CollectionConfig {
     private static final String COLLECTION_SPLITTER = "\\.";
     private static final int DEFAULT_STREAM_BATCH_SIZE = 1000;
+    private static final int DEFAULT_PARTITION_COUNT = 100;
+    private static final int DEFAULT_EXPORT_PARTITION_SIZE = 4000;
     @JsonProperty("collection")
     private @NotNull String collection;
 
-    @JsonProperty("export_config")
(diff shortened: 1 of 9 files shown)


data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/configuration/CollectionConfig.java
```diff
@@ -9,15 +9,17 @@
...
     @JsonProperty("collection")
     private @NotNull String collection;
 
-    @JsonProperty("export_config")
-    private ExportConfig exportConfig;
-
     @JsonProperty("export")
     private boolean export;
 
+    @JsonProperty("export_partition_size")
```
graytaylor0, line 20: What does this mean exactly?

graytaylor0 (approved):

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/configuration/CollectionConfig.java
dinujoh: This is the number of records of chunk size the export partition worker will work on when it executes query to DocDB/Mongo. The export workers run in parallel and they work on independent chunk of data.

oeyh (commented):
Looks good. One comment about the parameter name above.

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/configuration/CollectionConfig.java
dinujoh: updated to `export_batch_size`

graytaylor0 (approved):

oeyh (approved):

PR #4455, opened 2024-04-24
Add folder-based partitioning for s3 scan source

graytaylor0:
### Description
This change adds support for the s3 scan source to partition based on folders instead of objects. This means that, unlike object partitions, folder partitions will never be marked as complete, and each node / worker that grabs a folder partition will list and process objects in that folder, then give it up. The use case for this is order guarantee within S3 folders, as well as aggregations within S3 folders, as the same worker / node will process objects from the same S3 folder. To enable this mode, one must set the following configuration. 

```
source:
  s3:
    scan:
      folder_partitions:
        # The depth of the folders to partition on. For example, given object keys "folder-1/folder-3/folder-4/"  and "folder-1/folder-2/folder-5/", a depth of 1 would result in one partition of "folder-1/", while a depth of 2 would result in 2 partitions, "folder-1/folder-3/" and "folder-1/folder-2/". If an object does not contain enough depth, it will not be included in the scan.
        depth: 2 # Defaults to 1
        
        # The max number of objects to process before giving up the folder partition. If the number of objects in the folder is less than this maximum, then all of the objects in the folder will be processed before giving up the partition
        max_objects_per_ownership: 10 # Defaults to 50
```
 
### Check List

### DIFF
--- a/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3SourceConfig.java
+++ b/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3SourceConfig.java
@@ -104,6 +104,16 @@ public class S3SourceConfig {
         return true;
     }
 
+    @AssertTrue(message = "acknowledgments and delete_s3_objects_on_read must both be set to true when using PREFIX partition mode")
+    boolean isPrefixPartitionModeValid() {
+        if (s3ScanScanOptions != null &&
+                s3ScanScanOptions.getPartitioningOptions() != null) {
+            return acknowledgments && deleteS3ObjectsOnRead;
+        }
+
+        return true;
+    }
+
     public NotificationTypeOption getNotificationType() {
         return notificationType;
     }
--- a/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3SourceConfig.java
+++ b/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3SourceConfig.java
@@ -104,7 +104,7 @@ public class S3SourceConfig {
         return true;
     }
 
-    @AssertTrue(message = "acknowledgments and delete_s3_objects_on_read must both be set to true when using PREFIX partition mode")
+    @AssertTrue(message = "acknowledgments and delete_s3_objects_on_read must both be set to true when using folder_partitions mode")
     boolean isPrefixPartitionModeValid() {
         if (s3ScanScanOptions != null &&
                 s3ScanScanOptions.getPartitioningOptions() != null) {
--- a/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/ScanObjectWorker.java
+++ b/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/ScanObjectWorker.java
@@ -45,16 +45,18 @@ import java.util.stream.Collectors;
  * Class responsible for processing the s3 scan objects with the help of <code>S3ObjectWorker</code>
  * or <code>S3SelectWorker</code>.
  */
-public class ScanObjectWorker implements Runnable{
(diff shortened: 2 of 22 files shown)


data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3SourceConfig.java
```diff
@@ -104,6 +104,16 @@ boolean isCodecProvidedWhenNeeded() {
         return true;
     }
 
+    @AssertTrue(message = "acknowledgments and delete_s3_objects_on_read must both be set to true when using PREFIX partition mode")
```
oeyh, line 107: Minor: we should probably be consistent on the naming. I see both prefix partition and folder partition being used.

oeyh (approved):

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3SourceConfig.java
graytaylor0: good catch this was an outdated message where I initially had a `mode` of PREFIX or FULL_OBJECT_KEY

oeyh (approved):

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/ScanObjectWorker.java
```diff
@@ -37,7 +48,9 @@
 public class ScanObjectWorker implements Runnable{
 
     private static final Logger LOG = LoggerFactory.getLogger(ScanObjectWorker.class);
+    private static final Integer MAX_OBJECTS_PER_ACKNOWLEDGMENT_SET = 1;
 
+    static final Duration NO_OBJECTS_FOUND_BEFORE_PARTITION_DELETION_DURATION = Duration.ofMinutes(3);
```
dinujoh, line 53: Is this needed ?

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/ScanObjectWorker.java
```diff
@@ -198,8 +231,141 @@ private Optional<DeleteObjectRequest> processS3Object(final S3ObjectReference s3
...
+            }
+
+            sourceCoordinator.giveUpPartition(folderPartition.getPartitionKey(), Instant.now());
+            return;
+        }
+
+        Optional<S3SourceProgressState> folderPartitionProgressState = folderPartition.getPartitionState();
+        if (folderPartitionProgressState.isEmpty()) {
+            folderPartitionProgressState = Optional.of(new S3SourceProgressState(Instant.now().toEpochMilli()));
+        } else {
```
dinujoh, line 260: This is duplicate. Can be moved to beginning to the method.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/ScanObjectWorker.java
```diff
@@ -198,8 +231,141 @@ private Optional<DeleteObjectRequest> processS3Object(final S3ObjectReference s3
...
+        // Update the last time objects were found to support deletion of the partition after no objects found for some time
+        sourceCoordinator.saveProgressStateForPartition(folderPartition.getPartitionKey(), folderPartitionProgressState.get());
+
+
+        int objectsProcessed = 0;
+        int objectIndex = 0;
+
+        AcknowledgementSet acknowledgementSet = null;
+        String activeAcknowledgmentSetId = null;
+        while (objectIndex < objectsToProcess.size() && objectsProcessed < folderPartitioningOptions.getMaxObjectsPerOwnership()) {
```
dinujoh, line 273: This method can be split into smaller methods for readability. We can move this while loop to different method.

dinujoh (approved):

PR #4477, opened 2024-04-30
Move MongoDB/DocDB stream checkpoint to separate thread

dinujoh:
### Description
Move MongoDB/DocDB stream checkpoint to separate thread

 
### Check List

### DIFF
--- a/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamWorker.java
+++ b/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamWorker.java
@@ -30,7 +30,6 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -235,14 +234,12 @@ public class StreamWorker {
     }
 
     private void checkpointStream() {
-        String globalCheckpoint = null;
         long lastCheckpointTime = System.currentTimeMillis();
         while (!Thread.currentThread().isInterrupted()) {
-            if (!Objects.equals(globalCheckpoint,lastLocalCheckpoint) && lastLocalRecordCount != null && !sourceConfig.isAcknowledgmentsEnabled() && (System.currentTimeMillis() - lastCheckpointTime >= checkPointIntervalInMs)) {
+            if (lastLocalRecordCount != null && (System.currentTimeMillis() - lastCheckpointTime >= checkPointIntervalInMs)) {
                 LOG.debug("Perform regular checkpoint for resume token {} at record count {}", lastLocalCheckpoint, lastLocalRecordCount);
                 partitionCheckpoint.checkpoint(lastLocalCheckpoint, lastLocalRecordCount);
                 lastCheckpointTime = System.currentTimeMillis();
-                globalCheckpoint = lastLocalCheckpoint;
             }
 
             try {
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamWorker.java
```diff
@@ -100,9 +107,13 @@ public StreamWorker(final RecordBufferWriter recordBufferWriter,
         this.successItemsCounter = pluginMetrics.counter(SUCCESS_ITEM_COUNTER_NAME);
         this.failureItemsCounter = pluginMetrics.counter(FAILURE_ITEM_COUNTER_NAME);
         this.bytesReceivedSummary = pluginMetrics.summary(BYTES_RECEIVED);
+        this.executorService = Executors.newSingleThreadExecutor(BackgroundThreadFactory.defaultExecutorThreadFactory("mongodb-stream-checkpoint"));
         if (sourceConfig.isAcknowledgmentsEnabled()) {
             // starts acknowledgement monitoring thread
             streamAcknowledgementManager.init((Void) -> stop());
+        } else {
+            // checkpoint in separate thread
+            this.executorService.submit(this::checkpointStream);
```
graytaylor0, line 116: Why are we only adding this thread with acknowledgments enabled? I know checkpointing is done in the callback, but we definitely can't rely on that because if no records are received, we will still just block on `cursor.next()` and not have acknowledgments getting called back to checkpoint, so you will lose ownership

dinujoh: When ack is enabled, it's handled in `streamAcknowledgementManager`.

> so you will lose ownership

I currently have this check `!Objects.equals(globalCheckpoint,lastLocalCheckpoint) `. If we remove this check, this checkpoints old value and extends the lease.

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamWorker.java
```diff
@@ -100,9 +106,13 @@ public StreamWorker(final RecordBufferWriter recordBufferWriter,
         this.successItemsCounter = pluginMetrics.counter(SUCCESS_ITEM_COUNTER_NAME);
         this.failureItemsCounter = pluginMetrics.counter(FAILURE_ITEM_COUNTER_NAME);
         this.bytesReceivedSummary = pluginMetrics.summary(BYTES_RECEIVED);
+        this.executorService = Executors.newSingleThreadExecutor(BackgroundThreadFactory.defaultExecutorThreadFactory("mongodb-stream-checkpoint"));
         if (sourceConfig.isAcknowledgmentsEnabled()) {
```
chenqi0805, line 110: should this be `!sourceConfig.isAcknowledgmentsEnabled()`?

dinujoh: when `sourceConfig.isAcknowledgmentsEnabled()` it uses `streamAcknowledgementManager`, otherwise checkpoint is done in the private method `checkpointStream`

graytaylor0 (approved):

chenqi0805 (approved):

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamWorker.java
chenqi0805: I see. I might have confused the checkpoint concept with that in e2e ack.

PR #4508, opened 2024-05-06
Update docdb metrics

oeyh:
### Description
- Removes abstract class RecordBufferWriter and renames ExportRecordBufferWriter to RecordBufferWriter. This is because stream and export use the same ExportRecordBufferWriter
- Renames a few metric names to be consistent with DDB source
- Adds missing update or fixes the updating logic for a few metrics
 
### Issues Resolved
N/A
 
### Check List


data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/export/ExportPartitionWorker.java
```diff
@@ -177,6 +186,9 @@ public void run() {
                             LOG.debug("Write to buffer for line " + (recordCount - DEFAULT_BATCH_SIZE) + " to " + recordCount);
                             recordBufferWriter.writeToBuffer(acknowledgementSet, records);
                             records.clear();
+                            recordBytes.clear();
+                            successItemsCounter.increment(DEFAULT_BATCH_SIZE);
```
dinujoh, line 190: If there are less DEFAULT_BATCH_SIZE records, the metrics emitted will not be correct. It should min(recordCount, DEFAULT_BATCH_SIZE)

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/export/ExportPartitionWorker.java
```diff
@@ -177,6 +186,9 @@ public void run() {
                             LOG.debug("Write to buffer for line " + (recordCount - DEFAULT_BATCH_SIZE) + " to " + recordCount);
```
dinujoh, line 186: I have seen `(recordCount - DEFAULT_BATCH_SIZE)` value reporting negative. Need to fix this.

oeyh: `recordCount` seems to be a duplicate of `totalRecords` and is always greater than `startLine` at this point. Is it possible that recordCount (an int) overflows, causing the negative number? `recordCount` should probably be of long type.

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/export/ExportPartitionWorker.java
oeyh: I will use `records.size()` since `records` represents the current batch.

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/export/ExportPartitionWorker.java
oeyh: The export partition size is an int, so recordCount shouldn't overflow. I can't see how `recordCount - DEFAULT_BATCH_SIZE` becomes negative.

chenqi0805 (approved):

dinujoh (approved):

PR #4563, opened 2024-05-22
Revert addition of default region function from AwsCredentialsSupplier

graytaylor0:
### Description
This change removes the function to return the default region from the `AwsCredentialsSupplier` class. Rather than passing this region all of the clients, we will rely on the default SDK behavior to choose a region when one is not supplied. The region for STS will continue to use the configured `default` region in the `data-prepper-config.yaml`

### Check List


graytaylor0:
Closing as this code has been used by some plugins since it was merged

PR #4645, opened 2024-06-19
DocumentDB Source improvements

dinujoh:
### Description
1. Extend the export partition ownership during query partition creation
2. Add support to shutdown task refresher that starts export, stream scheduler and worker
3. Add AcknowledgmentStatus enum and code refactor to fail negative ack right away
 
### Check List

### DIFF
--- a/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/documentdb/MongoTasksRefresher.java
+++ b/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/documentdb/MongoTasksRefresher.java
@@ -91,7 +91,7 @@ public class MongoTasksRefresher implements PluginConfigObserver<MongoDBSourceCo
     private void refreshJobs(MongoDBSourceConfig pluginConfig) {
         final List<Runnable> runnables = new ArrayList<>();
         if (pluginConfig.getCollections().stream().anyMatch(CollectionConfig::isExport)) {
-            currentMongoDBExportPartitionSupplier = new MongoDBExportPartitionSupplier(pluginConfig, documentDBAggregateMetrics);
+            currentMongoDBExportPartitionSupplier = new MongoDBExportPartitionSupplier(pluginConfig, sourceCoordinator, documentDBAggregateMetrics);
             runnables.add(new ExportScheduler(sourceCoordinator, currentMongoDBExportPartitionSupplier, pluginMetrics));
             runnables.add(new ExportWorker(
                     sourceCoordinator, buffer, pluginMetrics, acknowledgementSetManager, pluginConfig, s3PathPrefix, documentDBAggregateMetrics));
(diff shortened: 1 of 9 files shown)


data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/documentdb/MongoTasksRefresher.java
```diff
@@ -110,4 +110,15 @@ private boolean basicAuthChanged(final MongoDBSourceConfig.AuthenticationConfig
...
                 !Objects.equals(currentAuthConfig.getPassword(), newAuthConfig.getPassword());
     }
+
+    /**
+     * Interrupt the running of schedulers.
+     * Each scheduler must implement logic for gracefully shutdown.
+     */
+    public void shutdown() {
+        if (currentExecutor != null) {
+            LOG.info("shutdown down export worker and stream worker");
```
oeyh, line 120: Minor:
```suggestion
            LOG.info("Shut down export worker and stream worker");
```

chenqi0805 (approved):

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/export/MongoDBExportPartitionSupplier.java
```diff
@@ -111,11 +115,15 @@ private PartitionIdentifierBatch buildPartitions(final ExportPartition exportPar
...
                             .withPartitionKey(String.format(MONGODB_PARTITION_KEY_FORMAT, collectionDbName, gteValueString, lteValueString, gteClassName, lteClassName))
                             .build());
+                    documentDBAggregateMetrics.getExportPartitionQueryCount().increment();
 
                     if (isLastBatch) {
                         break;
                     }
 
+                    // extend the ownership of the partition
+                    enhancedSourceCoordinator.saveProgressStateForPartition(exportPartition, null);
```
graytaylor0, line 125: Will this spam the source coordinator or should there be some time difference or condition for us to increase ownership timeout?

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/utils/DocumentDBSourceAggregateMetrics.java
```diff
@@ -17,8 +17,7 @@ public class DocumentDBSourceAggregateMetrics {
     private static final String DOCUMENT_DB_EXPORT_5XX_ERRORS = "export5xxErrors";
     private static final String DOCUMENT_DB_EXPORT_4XX_ERRORS = "export4xxErrors";
     private static final String DOCUMENT_DB_EXPORT_API_INVOCATIONS = "exportApiInvocations";
-
-
+    private static final String DOCUMENT_DB_EXPORT_PARTITION_QUERY_COUNT = "exportPartitionQueryCount";
```
oeyh, line 20: What is this aggregated metric used for?

dinujoh: This is to track number of export partition query created. When the collection size if very large (10's to 100's of GB), the scan of collection takes in minutes to hours to create query partition. We want to provide visibility to customer of query partition getting created and pipeline is not stuck.

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/export/MongoDBExportPartitionSupplier.java
dinujoh: It will only make calls to source coordinator when query partitions are created. For very large collection it take long time to scan entire collection, so we need to keep ownership of the partition. The calls to source coordinator must be max in 10's per second. I ran the test locally and I see version bump to 150 in 4-5 mins.

oeyh (approved):

PR #4700, opened 2024-07-01
Add Lambda Synchronous processor support

### LINKED ISSUE #4699: AWS Lambda as Synchronous Processor
Is your feature request related to a problem? Please describe.

Pipeline users want to send events to AWS Lambda.

Describe the solution you'd like
Create a processor in Data Prepper which uses lambda as a remote processor . It should support

- Retries
- Buffering capabilities

Without Batching:
```
lambda-pipeline:
...
  processor:
    - lambda:
        aws:
            region: us-east-1
            sts_role_arn: <arn>
            sts_overrides:
        function_name: "uploadToS3Lambda"
        mode: synchronous
        max_retries: 3
        dlq:
            s3:
                bucket: test-bucket
                key_path_prefix: dlq/
```

With Batching:
```
lambda-pipeline:
...
  processor:
    - lambda:
        aws:
            region: us-east-1
            sts_role_arn: <arn>
            sts_overrides:
        function_name: "uploadToS3Lambda"
        mode: synchronous
        max_retries: 3
        batch:
            batch_key: "user_key"
            threshold:
                event_count: 3
                maximum_size: 6mb
                event_collect_timeout: 15s
...
```

**Additional context**

srikanthjg:
### Description
Adds AWS lambda as a remote processor for dataprepper.
Further details mentioned in #4699  

### Issues Resolved
Resolves #4699
 
### Check List


data-prepper-plugins/aws-lambda/build.gradle
```diff
@@ -26,7 +26,12 @@ dependencies {
     testAnnotationProcessor 'org.projectlombok:lombok:1.18.20'
     testImplementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
     testImplementation project(':data-prepper-test-common')
-    testImplementation project(':data-prepper-plugins:parse-json-processor')
+    testImplementation 'org.powermock:powermock-module-junit4:2.0.9'
+    testImplementation 'org.powermock:powermock-api-mockito2:2.0.9'
+    testImplementation 'junit:junit:4.13.2'
```
dlvenable, line 31: You shouldn't need any of these four lines. They are provided by the root project.

dlvenable (changes requested):

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaClientFactory.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.opensearch.dataprepper.aws.api.AwsCredentialsOptions;
+import org.opensearch.dataprepper.aws.api.AwsCredentialsSupplier;
+import org.opensearch.dataprepper.plugins.lambda.common.config.AwsAuthenticationOptions;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
+import software.amazon.awssdk.core.retry.RetryPolicy;
+import software.amazon.awssdk.services.lambda.LambdaClient;
+
+public final class LambdaClientFactory {
```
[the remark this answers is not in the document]
srikanthjg: i can create an interface to add a layer of abstraction but i dont see any other place where we could use it. I can add it if necessary.

kkondaka:
@srikanthjg white source check is failing. I am ready to approve this.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorConfig.java
```diff
@@ -0,0 +1,63 @@
...
+    @JsonProperty("function_name")
+    @NotEmpty
+    @NotNull
+    @Size(min = 3, max = 500, message = "function name length should be at least 3 characters")
+    private String functionName;
+
+    @JsonProperty("max_retries")
+    private int maxConnectionRetries = DEFAULT_CONNECTION_RETRIES;
+
+    @JsonProperty("mode")
```
dlvenable, line 35: The term "mode" is quite ambiguous. I think we can borrow the term "invocation_type" from AWS Lambda itself.

https://docs.aws.amazon.com/lambda/latest/api/API_Invoke.html#API_Invoke_RequestSyntax

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,274 @@
...
+        }
+        mode = lambdaProcessorConfig.getMode();
+        // TODO - Support for Async mode to be added.
+        if (mode != null && mode.equalsIgnoreCase(LambdaProcessorConfig.SYNCHRONOUS_MODE)) {
+            invocationType = SYNC_INVOCATION_TYPE;
+        } else {
+            throw new RuntimeException("Unsupported mode " + mode);
+        }
+
+        codec = new LambdaJsonCodec(batchKey);
```
dlvenable, line 121: This approach is very restrictive. It assumes that the body fits the format we ask. We should follow the same pattern used elsewhere in Data Prepper by allowing for a pluggable output codec.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorConfig.java
```diff
@@ -0,0 +1,63 @@
...
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotEmpty;
+import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.Size;
+import org.opensearch.dataprepper.plugins.lambda.common.config.AwsAuthenticationOptions;
+import org.opensearch.dataprepper.plugins.lambda.common.config.BatchOptions;
+
+public class LambdaProcessorConfig {
+
+    public static final String SYNCHRONOUS_MODE = "RequestResponse";
```
dlvenable, line 17: We should stick with Data Prepper naming conventions: `request_response`.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorConfig.java
```diff
@@ -0,0 +1,63 @@
...
+    public static final String ASYNCHRONOUS_MODE = "Event";
```
dlvenable, line 18: Change to `event`.

dlvenable (changes requested):

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
srikanthjg: LambdaJsonCodec is an implementation of OutputCodec - [link](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/codec/LambdaJsonCodec.java#L20) . I needed this specifically for batch processing when more than one event needs to be mapped to json. JsonOutputCodec is only event at a time. Maybe i can change the name to something more generic like BulkJsonOutputCodec? It will be used for all dial out processors.

dlvenable: All `OutputCodec`s are made for batches.

Customers can use either `JsonOutputCodec` or `NdjsonOutputCodec`. 

The `json` codec already supports a configurable key name. This would replace the `batch_key` which you don't need.

```
codec:
  json:
    key_name: myKey
```

dlvenable: So the customer can have two options:

1. JSON

```
codec:
  json:
    key_name: myKey
```

Yields:

```
{
  "myKey" : [
     { ...event1... },
     { ...event2... },
     { ...event3... }
  ]
}
```

2. The customer can use `ndjson`

```
codec:
  ndjson:
```

Yields:

```
{ ...event1... }
{ ...event2... }
{ ...event3... }
```

srikanthjg: Understood, there is also a difference when it comes to handling single event without batch. In this case, i still want to convert dataprepper event to json but i dont want to have a key, i want to pass on the user's data as it is to lambda as payload; but current output codec forces me to have a key. To address that, i either need to add new behaviour to json and ndjson writeEvent method, to convert event directly to json OR write a new codec(which is what i did).

[LambdaJsonCodec](https://github.com/srikanthjg/data-prepper/blob/8f33225a32bc4ba226f6f83eb88d3ffff5212ad6/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/codec/LambdaJsonCodec.java#L62) and [JsonOutputCodec](https://github.com/srikanthjg/data-prepper/blob/8f33225a32bc4ba226f6f83eb88d3ffff5212ad6/data-prepper-plugins/parse-json-processor/src/main/java/org/opensearch/dataprepper/plugins/codec/json/JsonOutputCodec.java#L70) difference in the way we handle writeEvent. 

If we want to go with the pluggable codec approach, i think we need to modify the behaviour of the existing outputCodec, is that what we want?

srikanthjg: regarding the configuration, i already have "invocation_type" as a configuration, this allows to set per event invocation or batch invocation(RequestResponse or Event).

If we are implementing this internally, i cannot use parse-json-processor as a plugin but will have to take a dependency on it. Is it ok for one processor to take a dependency on the other?

I wanted to avoid this, hence went with implementing a custom codec. But i think this codec can also be used by other dial-out processors eventually, i can make it generic.

dlvenable (changes requested):

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSinkConfig.java
```diff
@@ -18,10 +18,10 @@
 
 public class LambdaSinkConfig {
 
-    public static final String REQUEST_RESPONSE = "RequestResponse";
-    public static final String EVENT = "Event";
-    public static final String BATCH_EVENT = "batch_event";
-    public static final String SINGLE_EVENT = "single_event";
+    public static final String REQUEST_RESPONSE = "request-response";
```
dlvenable, line 21: Perhaps make a `CommonLambdaConfig` class that has these constants. We should avoid duplicating these or we may have future mismatches.

dlvenable (changes requested):

dlvenable (approved):
Thank you @srikanthjg for this contribution!

dlvenable (approved):

data-prepper-plugins/aws-lambda/README.md
```diff
@@ -1,4 +1,55 @@
...
+        payload_model: "batch_event"
+        batch:
+            key_name: "osi_key"
+            threshold:
+                event_count: 10
+                event_collect_timeout: 15s
+                maximum_size: 3mb
+```
+
+`invocation_type` as RequestResponse will be used when the response from aws lambda comes back to dataprepper.
```
dinujoh, line 27: nit:
`invocation_type` as RequestResponse is used when DataPrepper needs to process the response from AWS Lambda.

`invocation_type` as Event is used when the response from AWS Lambda is sent to an S3 bucket.

data-prepper-plugins/aws-lambda/README.md
```diff
@@ -1,4 +1,55 @@
...
+`invocation_type` as Event is used when the response from aws lambda goes to an s3 bucket.
+
+In batch options, an implicit batch threshold option is that if events size is 3mb, we flush it.
+`payload_model` this is used to define how the payload should be constructed from a dataprepper event.
+`payload_model` as batch_event is used when the output needs to be formed as a batch of multiple events,
```
dinujoh, line 32: are there other values for `paylod_model` ?

data-prepper-plugins/aws-lambda/src/integrationTest/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorServiceIT.java
```diff
@@ -0,0 +1,165 @@
...
+
+            Record<Event> eventRecord = new Record<>(JacksonEvent.builder().withData(eventData).withEventType("event").build());
+            recordList.add(eventRecord);
+        }
+        return recordList;
+    }
+
+    @ParameterizedTest
+    @ValueSource(ints = {1,3})
+    void verify_records_to_lambda_success(final int recordCount) throws Exception {
```
dinujoh, line 130: consider adding test for `InvocationType` Event

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/client/LambdaClientFactory.java
```diff
@@ -16,20 +16,21 @@
...
-         final AwsCredentialsOptions awsCredentialsOptions = convertToCredentialsOptions(lambdaSinkConfig.getAwsAuthenticationOptions());
+    public static LambdaClient createLambdaClient(final AwsAuthenticationOptions awsAuthenticationOptions,
+                                                  final int maxConnectionRetries,
+                                                  final AwsCredentialsSupplier awsCredentialsSupplier) {
+         final AwsCredentialsOptions awsCredentialsOptions = convertToCredentialsOptions(awsAuthenticationOptions);
          final AwsCredentialsProvider awsCredentialsProvider = awsCredentialsSupplier.getProvider(awsCredentialsOptions);
 
             return LambdaClient.builder()
-                .region(lambdaSinkConfig.getAwsAuthenticationOptions().getAwsRegion())
+                .region(awsAuthenticationOptions.getAwsRegion())
```
dinujoh, line 26: Consider enabling SDK metrics to track number of request, timeout, throttle etc.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/config/BatchOptions.java
```diff
@@ -7,21 +7,24 @@
...
 
 public class BatchOptions {
+    static final String DEFAULT_KEY_NAME = "events";
 
-    private static final String DEFAULT_BATCH_KEY = "events";
-
-    @JsonProperty("batch_key")
-    private String batchKey = DEFAULT_BATCH_KEY;
+    @JsonProperty("key_name")
+    @Size(min = 1, max = 2048)
```
dinujoh, line 17: Is max value on the `key_name` consistent across data prepper for other process ?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,296 @@
...
+            maxBytes = batchOptions.getThresholdOptions().getMaximumSize(); // remove
+            maxCollectionDuration = batchOptions.getThresholdOptions().getEventCollectTimeOut();
+            isBatchEnabled = true;
+            LOG.info("maxEvents:" + maxEvents + " maxbytes:" + maxBytes + " maxDuration:" + maxCollectionDuration);
+        } else if(payloadModel.equals(SINGLE_EVENT)) {
+            NdjsonOutputConfig ndjsonOutputCodecConfig = new NdjsonOutputConfig();
+            codec = new NdjsonOutputCodec(ndjsonOutputCodecConfig);
+            isBatchEnabled = false;
+        } else{
+            throw new RuntimeException("invalid payload_model option");
```
dinujoh, line 129: Can this validation be part of lambdaProcessorConfig ?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        if(!lambdaProcessorConfig.getInvocationType().equals(LambdaCommonConfig.EVENT) &&
+                !lambdaProcessorConfig.getInvocationType().equals(LambdaCommonConfig.REQUEST_RESPONSE)){
+            throw new RuntimeException("Unsupported invocation type " + lambdaProcessorConfig.getInvocationType());
```
dinujoh, line 134: same as above

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,296 @@
...
+
+            if (whenCondition != null && !expressionEvaluator.evaluateConditional(whenCondition, event)) {
+                continue;
+            }
+
+            try {
+                if (currentBuffer.getEventCount() == 0) {
+                    codec.start(currentBuffer.getOutputStream(), event, codecContext);
+                }
+                codec.writeEvent(event, currentBuffer.getOutputStream());
```
dinujoh, line 180: is `currentBuffer` thread safe ?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,296 @@
...
+    public boolean isReadyForShutdown() {
+        return false;
+    }
+
+    @Override
+    public void shutdown() {
+
+    }
+
+    void flushToLambdaIfNeeded(List<Record<Event>> resultRecords) throws InterruptedException, IOException {
```
dinujoh, line 215: this can be private function

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,296 @@
...
+
+        LOG.info("Flush to Lambda check: currentBuffer.size={}, currentBuffer.events={}, currentBuffer.duration={}", currentBuffer.getSize(), currentBuffer.getEventCount(), currentBuffer.getDuration());
```
dinujoh, line 217: Is there excessive logging in this method ?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,296 @@
...
+                //reset buffer after flush
+                currentBuffer = bufferFactory.getBuffer(lambdaClient, functionName, invocationType);
+            } else {
+                LOG.error("Failed to save to Lambda {}", functionName);
+                numberOfRecordsFailedCounter.increment(currentBuffer.getEventCount());
+            }
+        }
+    }
+
+    LambdaResult retryFlushToLambda(Buffer currentBuffer, final AtomicReference<String> errorMsgObj) throws InterruptedException {
```
dinujoh, line 245: this can be private function

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,296 @@
...
+                }
+                Thread.sleep(5000);
+            }
+        } while (!isUploadedToLambda);
+
+        LambdaResult lambdaResult = LambdaResult.builder().withIsUploadedToLambda(false).withLambdaResponse(null).build();
+        return lambdaResult;
+    }
+
+    Event convertLambdaResponseToEvent(InvokeResponse lambdaResponse) {
```
dinujoh, line 271: this can be private function

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSinkService.java
```diff
@@ -87,27 +93,34 @@ public LambdaSinkService(final LambdaClient lambdaClient,
...
-
-        codec = new LambdaJsonCodec(batchKey);
         bufferedEventHandles = new LinkedList<>();
         events = new ArrayList();
 
-        invocationType = ASYNC_INVOCATION_TYPE;
+        //Translate dataprepper invocation type to lambda invocation type
+        Map<String, String> invocationTypeMap = Map.of(
+                LambdaCommonConfig.EVENT, EVENT_LAMBDA
+        );
```
dinujoh, line 122: Move this to static constant

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,296 @@
...
+        bufferedEventHandles = new LinkedList<>();
+        events = new ArrayList();
+
+        lambdaClient = LambdaClientFactory.createLambdaClient(lambdaProcessorConfig.getAwsAuthenticationOptions(),
+                lambdaProcessorConfig.getMaxConnectionRetries()
+                , awsCredentialsSupplier);
+
+        this.bufferFactory = new InMemoryBufferFactory();
+        try {
+            currentBuffer = this.bufferFactory.getBuffer(lambdaClient, functionName, invocationType);
```
dinujoh, line 153: The buffer is overloaded in DataPrepper. Looks like this is not just buffer but tightly coupled with lambda. We should consider renaming this class and interface to be clear.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,296 @@
...
+
+            try {
+                InvokeResponse resp = currentBuffer.flushToLambdaSync();
+                isUploadedToLambda = Boolean.TRUE;
+                LambdaResult lambdaResult = LambdaResult.builder().withIsUploadedToLambda(isUploadedToLambda).withLambdaResponse(resp).build();
+                return lambdaResult;
+            } catch (AwsServiceException | SdkClientException e) {
+                errorMsgObj.set(e.getMessage());
+                LOG.error("Exception occurred while uploading records to lambda. Retry countdown  : {} | exception:", retryCount, e);
+                --retryCount;
```
dinujoh, line 258: is this retry on top of lambda client retry ? Any reason we need this ?

data-prepper-plugins/aws-lambda/src/integrationTest/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorServiceIT.java
srikanthjg: invocation type event will be disabled for now, will be releasing event type with asynchronous support that requires additional infra changes. I have disabled it in the verification for now, ll fix the readme.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/config/BatchOptions.java
srikanthjg: yes

data-prepper-plugins/aws-lambda/README.md
srikanthjg: fixed

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
srikanthjg: i am handling the same way we would do it in the sink. i can address refactor in another pr.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
srikanthjg: yes this is running in the context of a processor.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
srikanthjg: sure will reduce them.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorConfig.java
```diff
@@ -0,0 +1,69 @@
...
+    private String functionName;
+
+    @JsonProperty("max_retries")
+    private int maxConnectionRetries = DEFAULT_CONNECTION_RETRIES;
+
+    @JsonProperty("invocation_type")
+    private String invocationType = REQUEST_RESPONSE;
+
+    @JsonProperty("payload_model")
+    private String payloadModel = BATCH_EVENT;
```
dlvenable, line 38: We should make a Java `enum` for this.

Here is an example:

https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/configuration/NotificationTypeOption.java

I'm ok with doing this in a follow-on PR.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorConfig.java
```diff
@@ -0,0 +1,69 @@
...
+    @JsonProperty("function_name")
+    @NotEmpty
+    @Size(min = 3, max = 500, message = "function name length should be at least 3 characters")
+    private String functionName;
+
+    @JsonProperty("max_retries")
+    private int maxConnectionRetries = DEFAULT_CONNECTION_RETRIES;
+
+    @JsonProperty("invocation_type")
+    private String invocationType = REQUEST_RESPONSE;
```
dlvenable, line 35: We should make this an enum as well.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSinkConfig.java
```diff
@@ -38,12 +38,21 @@ public class LambdaSinkConfig {
     @JsonProperty("max_retries")
     private int maxConnectionRetries = DEFAULT_CONNECTION_RETRIES;
 
+    @JsonProperty("invocation_type")
+    private String invocationType = EVENT;
```
dlvenable, line 42: Should be an enum as well.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSinkConfig.java
```diff
@@ -38,12 +38,21 @@ public class LambdaSinkConfig {
...
+
+    @JsonProperty("payload_model")
+    private String payloadModel = BATCH_EVENT;
```
dlvenable, line 45: Should be an enum as well.

dinujoh:
Did you look into moving some of the null check validations from the plugin into Config ? Other than this the changes look good to me

dinujoh (approved):

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -0,0 +1,265 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+
+@DataPrepperPlugin(name = "aws_lambda", pluginType = Processor.class, pluginConfigurationType = LambdaProcessorConfig.class)
+public class LambdaProcessor extends AbstractProcessor<Record<Event>, Record<Event>> {
+
+    public static final String NUMBER_OF_RECORDS_FLUSHED_TO_LAMBDA_SUCCESS = "lambdaProcessorObjectsEventsSucceeded";
+    public static final String NUMBER_OF_RECORDS_FLUSHED_TO_LAMBDA_FAILED = "lambdaProcessorObjectsEventsFailed";
+    public static final String LAMBDA_LATENCY_METRIC = "lambdaProcessorLatency";
```
dlvenable, line 64: We can rename this to simply `requestLatency`. Or you could call it `lambdaRequestLatency`, but that seems unnecessary.

As I read the code, this is the time to make the request to Lambda regardless of it being request/response or event.

dlvenable (approved):
@srikanthjg , Thank you for this contribution. I have a few other changes we should try to get in to improve it. But, let's follow-on in another PR.

PR #4814, opened 2024-08-08
Fix null document in DLQ object

kkondaka:
### Description
When failures happen in opensearch sink and the documents are sent to DLQ, in some cases, the document is null. This change fixes that issue. There is a bug in the code where the document is set to null when the operation is UPDATE/UPSERT/DELETE
 
### Check List


oeyh (approved):
Thanks for fixing this! 
Were you able to verify the fix or should we add a test case for this?

graytaylor0 (approved):

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -448,9 +448,9 @@ public void doOutput(final Collection<Record<Event>> records) {
...
 
       SerializedJson serializedJsonNode = null;
-      if (StringUtils.equals(action, OpenSearchBulkActions.UPDATE.toString()) ||
-          StringUtils.equals(action, OpenSearchBulkActions.UPSERT.toString()) ||
-          StringUtils.equals(action, OpenSearchBulkActions.DELETE.toString())) {
+      if (StringUtils.equals(eventAction, OpenSearchBulkActions.UPDATE.toString()) ||
+          StringUtils.equals(eventAction, OpenSearchBulkActions.UPSERT.toString()) ||
+          StringUtils.equals(eventAction, OpenSearchBulkActions.DELETE.toString())) {
             serializedJsonNode = SerializedJson.fromJsonNode(event.getJsonNode(), document);
       }
```
dinujoh, line 455: We should handle the else condition so that the null document issue doesn't occur again.

kkondaka: Do we really need else? The variable is initialized to `null` because that's what we want in case of else. This `serializedJsonNode` is used in `BulkOperationWrapper` as follows
```
public Object getDocument() {                                                                                                         
        if (bulkOperation.isUpdate() || bulkOperation.isDelete()) {                                                                       
            return jsonNode;                                                                                                              
        }                                                                                                                                 
        return getValueFromConverter(BULK_OPERATION_TO_DOCUMENT_CONVERTERS);                                                              
    }      

```

PR #4943, opened 2024-09-12
Reproduce issue where a join expression doesn't work over array creat…

miguel-vila:
…ed in the middle of add_entries

### Description

So I'm not sure if this is something that should be supported: applying a `join` over a field that has been created by previous entries. I'm using this for some types of records where it works as I expect it, but I've noticed that for others it doesn't (I haven't looked deeply into the difference, might update the PR later if I find something).

In the test: the `elements` array is being built as expected, but the `join` expression is not being applied to build a new `text` field.


miguel-vila:
issue seems to be on our side 😅 , closing this

PR #4955, opened 2024-09-18
Shorten S3 prefix to meet the requirement of RDS export API

oeyh:
### Description
RDS startExportTask API requires S3 prefix fewer than 60 characters. The S3 prefix for export defined by pipeline is `user-defined_prefix/source_identifier/rds-export`. The source identifier along often exceed the limit. 

This PR shortens source identifier to 15 chars and change `rds-export` to just `rds`. This leaves about 40 chars for user defined prefix.
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/utils/IdentifierShortener.java
+++ b/data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/utils/IdentifierShortener.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.dataprepper.plugins.source.rds.utils;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
+
+public class IdentifierShortener {
+
+    public static String shortenIdentifier(final String identifier, final int maxLength) {
+        if (identifier.length() <= maxLength) {
+            return identifier;
+        }
+
+        try {
+            // Create SHA-256 hash
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            byte[] encodedhash = digest.digest(identifier.getBytes(StandardCharsets.UTF_8));
+
+            // Convert byte array to Base64 string
+            String base64Hash = Base64.getUrlEncoder().withoutPadding().encodeToString(encodedhash);
+
+            // Return the first maxLength characters
+            return base64Hash.substring(0, Math.min(base64Hash.length(), maxLength));
+        } catch (final NoSuchAlgorithmException e) {
+            return identifier.substring(0, maxLength);
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


data-prepper-plugins/rds-source/src/test/java/org/opensearch/dataprepper/plugins/source/rds/utils/IdentifierShortenerTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    void shortenIdentifier_when_input_is_longer_than_max_then_return_shortened_result() {
+        final int maxLength = 5;
+        final String testInput = UUID.randomUUID().toString();
+        assertThat(testInput.length(), greaterThan(maxLength));
+
+        final String result = IdentifierShortener.shortenIdentifier(testInput, maxLength);
+
+        assertThat(result.length(), lessThanOrEqualTo(maxLength));
```
dinujoh, line 41: should this be `assertThat(result.length(), equalTo(testInput));`

oeyh: `testInput` is longer than maxLength (5) here. This is to verify that after shortening, it's equal or less than maxLength.

dinujoh: I had typo.. `equalTo(maxLength))`

dinujoh: Verify if there are no collisions when shortening the identifier.

dinujoh (approved):

data-prepper-plugins/rds-source/src/test/java/org/opensearch/dataprepper/plugins/source/rds/utils/IdentifierShortenerTest.java
oeyh: Did some smoke testing with long and similar inputs shortened to 10 chars, no collisions. With 15 chars used in rds source, the chance of collision should be very slim.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/utils/IdentifierShortener.java
```diff
@@ -0,0 +1,34 @@
...
+        try {
+            // Create SHA-256 hash
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            byte[] encodedhash = digest.digest(identifier.getBytes(StandardCharsets.UTF_8));
+
+            // Convert byte array to Base64 string
+            String base64Hash = Base64.getUrlEncoder().withoutPadding().encodeToString(encodedhash);
+
+            // Return the first maxLength characters
+            return base64Hash.substring(0, Math.min(base64Hash.length(), maxLength));
```
chenqi0805, line 29: QUES: what is the impact if there is a collision in the encoded result?

oeyh: There's a slim chance of collision between two pipelines configured with the same `s3_bucket` and `s3_prefix` in RDS source. In that case, the two pipelines will share the same s3 buffer, so one pipeline can process data that is supposed to be processed by the other pipeline and vice versa.

oeyh: When the bucket is used by a single user, the collision only happens within a single user's ownership. But if the bucket is multi-tenant, we will need to be more careful to avoid collisions.

PR #4980, opened 2024-09-26
Adding a new API to get the current transformed pipelines as a JSON

sb2k16:
### Description
This PR Is to add a new API to get the body of the transformed pipelines running in the current data prepper instance as an entire JSON string. The API supports both GET and POST methods.

The main purpose is to have a mechanism to get the transformed body of the entire set of pipelines running in the current data prepper instance which should enable faster debugging and troubleshooting purpose.
 
### Check List

### DIFF
--- a/data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/server/DataPrepperServer.java
+++ b/data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/server/DataPrepperServer.java
@@ -30,7 +30,7 @@ public class DataPrepperServer {
     private static final Logger LOG = LoggerFactory.getLogger(DataPrepperServer.class);
     private final HttpServerProvider serverProvider;
     private final ListPipelinesHandler listPipelinesHandler;
-    private final GetTransformedPipelinesBodyHandler getTransformedPipelinesBodyHandler;
+    private final GetPipelinesHandler getPipelinesHandler;
     private final ShutdownHandler shutdownHandler;
     private final PrometheusMeterRegistry prometheusMeterRegistry;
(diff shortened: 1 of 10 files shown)


data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -30,6 +30,7 @@ public class DataPrepperServer {
     private static final Logger LOG = LoggerFactory.getLogger(DataPrepperServer.class);
     private final HttpServerProvider serverProvider;
     private final ListPipelinesHandler listPipelinesHandler;
+    private final GetTransformedPipelinesBodyHandler getTransformedPipelinesBodyHandler;
```
kkondaka, line 33: Should this be `getTransformedPipelinesConfigHander`?

data-prepper-core/src/main/java/org/opensearch/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -67,6 +70,7 @@ private HttpServer createServer() {
 
         createContext(server, listPipelinesHandler, authenticator, "/list");
         createContext(server, shutdownHandler, authenticator, "/shutdown");
+        createContext(server, getTransformedPipelinesBodyHandler, authenticator, "/getPipelineBody");
```
kkondaka, line 73: Same as above. Should this be `/getPipelineTransformedConfig`. Who knows we may need `/getPipelineConfig` in future to get the config before transformation.

dlvenable (approved):

PR #4988, opened 2024-09-30
Make acknowledgment_timeout configurable for s3 scan source, configure timeout to 10 minutes for MongoDB L2 transform

graytaylor0:
### Description
If there is some reason that causes partition for s3 scan to not be given up, or if the acknowledgment is never received for a partition, s3 scan defaults to a 2 hour acknowledgment wait before reprocessing. For mongodb pipelines, this can cause high delay for some records, and these pipelines are meant to be low latency.

This change adds an `acknowledgment_timeout` parameter to S3 scan, and sets it to 10 minutes in the DocumentDB L2 transform.
 
### Check List

### DIFF
--- a/data-prepper-plugins/mongodb/src/main/resources/org/opensearch/dataprepper/transforms/templates/documentdb-template.yaml
+++ b/data-prepper-plugins/mongodb/src/main/resources/org/opensearch/dataprepper/transforms/templates/documentdb-template.yaml
@@ -66,6 +66,7 @@
       delete_s3_objects_on_read: true
       disable_s3_metadata_in_event: true
       scan:
+        acknowledgment_timeout: "PT10M"
         folder_partitions:
           depth: "<<FUNCTION_NAME:calculateDepth,PARAMETER:$.<<pipeline-name>>.source.documentdb.s3_prefix>>"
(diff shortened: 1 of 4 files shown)


data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3SourceConfig.java
```diff
@@ -27,6 +27,7 @@
 
 public class S3SourceConfig {
     static final Duration DEFAULT_BUFFER_TIMEOUT = Duration.ofSeconds(10);
+    static final Duration DEFAULT_ACKNOWLEDGMENT_TIMEOUT = Duration.ofSeconds(10);
```
dinujoh, line 30: Is this constant used ?

graytaylor0: It is not used. Will remove

dinujoh (dismissed):

oeyh (approved):

PR #5103, opened 2024-10-24
Add jsonValue check on Enum deserializer, default to checking name() function of enum

graytaylor0:
### Description
The new EnumDeserializer defaults to checking for the lower case enum value when there is no `@JsonCreator` on the enum class, and this broke basic enums like

```
enum MyEnum {
  ENUM_ONE,
  ENUM_TWO
}
```

Since we don't have consistency in enums as lower case we have to default to checking both for lowercase and uppercase values.

New behavior for deserializing enums

1. Check for JsonCreator and compare using that
2. Check for JsonValue and compare using that
3. Compare to the enum.name() 
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-pipeline-parser/src/test/java/org/opensearch/dataprepper/pipeline/parser/EnumDeserializerTest.java
+++ b/data-prepper-pipeline-parser/src/test/java/org/opensearch/dataprepper/pipeline/parser/EnumDeserializerTest.java
@@ -67,6 +67,21 @@ public class EnumDeserializerTest {
     }
 
     @ParameterizedTest
+    @EnumSource(TestEnumOnlyUppercase.class)
+    void enum_class_with_just_enum_values_returns_expected_enum_constant(final TestEnumOnlyUppercase testEnumOption) throws IOException {
+        final EnumDeserializer objectUnderTest = createObjectUnderTest(TestEnumOnlyUppercase.class);
+        final JsonParser jsonParser = mock(JsonParser.class);
+        final DeserializationContext deserializationContext = mock(DeserializationContext.class);
+        when(jsonParser.getCodec()).thenReturn(objectMapper);
+
+        when(objectMapper.readTree(jsonParser)).thenReturn(new TextNode(testEnumOption.toString()));
+
+        Enum<?> result = objectUnderTest.deserialize(jsonParser, deserializationContext);
+
+        assertThat(result, equalTo(testEnumOption));
+    }
+
+    @ParameterizedTest
     @EnumSource(TestEnumWithoutJsonCreator.class)
     void enum_class_without_json_creator_annotation_returns_expected_enum_constant(final TestEnumWithoutJsonCreator enumWithoutJsonCreator) throws IOException {
         final EnumDeserializer objectUnderTest = createObjectUnderTest(TestEnumWithoutJsonCreator.class);
@@ -170,5 +185,10 @@ public class EnumDeserializerTest {
         static TestEnumWithoutJsonCreator fromOptionValue(final String option) {
             return NAMES_MAP.get(option);
         }
+    }
+
+    private enum TestEnumOnlyUppercase {
+        VALUE_ONE,
+        VALUE_TWO;
     }
 }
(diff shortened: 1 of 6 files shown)


dlvenable (dismissed):
Which enums are supporting upper case presently? Those should be updated with `@JsonCreator` and we should move away from having these inconsistencies.

Supporting this will allow the inconsistencies to grow, whereas we should be working to remove them.

sb2k16 (dismissed):

chenqi0805 (dismissed):

data-prepper-pipeline-parser/src/test/java/org/opensearch/dataprepper/pipeline/parser/EnumDeserializerTest.java
```diff
@@ -66,15 +66,45 @@ void enum_class_with_json_creator_annotation_returns_expected_enum_constant(fina
...
+
+    @ParameterizedTest
+    @EnumSource(TestEnumOnlyUppercase.class)
+    void enum_class_with_just_enum_values_returns_expected_enum_constant(final TestEnumOnlyUppercase testEnumOption) throws IOException {
+        final EnumDeserializer objectUnderTest = createObjectUnderTest(TestEnumOnlyUppercase.class);
+        final JsonParser jsonParser = mock(JsonParser.class);
+        final DeserializationContext deserializationContext = mock(DeserializationContext.class);
+        when(jsonParser.getCodec()).thenReturn(objectMapper);
+
+        when(objectMapper.readTree(jsonParser)).thenReturn(new TextNode(testEnumOption.toString()));
```
dlvenable, line 92: I think this should return using `.name()` instead of `.toString()`.

graytaylor0: No because in this case the value that users have to pass to the config is lowercase

data-prepper-pipeline-parser/src/test/java/org/opensearch/dataprepper/pipeline/parser/EnumDeserializerTest.java
```diff
@@ -66,15 +66,45 @@ void enum_class_with_json_creator_annotation_returns_expected_enum_constant(fina
...
+
+        Enum<?> result = objectUnderTest.deserialize(jsonParser, deserializationContext);
+
+        assertThat(result, equalTo(testEnumOption));
+    }
```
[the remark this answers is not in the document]
graytaylor0: Isn't that already covered with this test `void enum_class_without_json_creator_or_json_value_annotation_returns_expected_enum_constant`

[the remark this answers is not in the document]
dlvenable: For that test, `toString()` returns the name, so you can't be sure if the code is using name() or toString().

Can you change that enum's `toString()` to return a random UUID?

data-prepper-pipeline-parser/src/test/java/org/opensearch/dataprepper/pipeline/parser/EnumDeserializerTest.java
dlvenable: This testing is using the `TestEnumOnlyUppercase` enum. This does not have a `@JsonCreator`.

graytaylor0: Ah I see

data-prepper-pipeline-parser/src/test/java/org/opensearch/dataprepper/pipeline/parser/EnumDeserializerTest.java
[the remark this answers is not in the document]
graytaylor0: Sure will make that change

dlvenable (approved):

PR #5186, opened 2024-11-12
Handle primary key change events in RDS source

oeyh:
### Description
It's possible that value of a primary key is changed in MySQL update statement. When this happens, we need to generate two events - one for indexing the updated record (with the new primary key value), and another one for deleting the old record. This PR add changes to handle it.

#### Testing
Tested against an Aurora MySQL table and verified that the two expected events were generated.
 
### Issues Resolved
Contributes to #4561 
 
### Check List

### DIFF
--- a/data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/stream/BinlogEventListener.java
+++ b/data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/stream/BinlogEventListener.java
@@ -319,7 +319,8 @@ public class BinlogEventListener implements BinaryLogClient.EventListener {
         handleRowChangeEvent(event, data.getTableId(), data.getRows(), Collections.nCopies(data.getRows().size(), OpenSearchBulkActions.DELETE));
     }
 
-    private boolean isValidTableId(long tableId) {
+    // Visible For Testing
+    boolean isValidTableId(long tableId) {
         if (!tableMetadataMap.containsKey(tableId)) {
             LOG.debug("Cannot find table metadata, the event is likely not from a table of interest or the table metadata was not read");
             return false;
@@ -333,10 +334,11 @@ public class BinlogEventListener implements BinaryLogClient.EventListener {
         return true;
     }
 
-    private void handleRowChangeEvent(com.github.shyiko.mysql.binlog.event.Event event,
-                                      long tableId,
-                                      List<Serializable[]> rows,
-                                      List<OpenSearchBulkActions> bulkActions) {
+    // Visible For Testing
+    void handleRowChangeEvent(com.github.shyiko.mysql.binlog.event.Event event,
+                              long tableId,
+                              List<Serializable[]> rows,
+                              List<OpenSearchBulkActions> bulkActions) {
 
         // Update binlog coordinate after it's first assigned in rotate event handler
         if (currentBinlogCoordinate != null) {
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/stream/BinlogEventListener.java
```diff
@@ -343,31 +354,36 @@ private void handleRowChangeEvent(com.github.shyiko.mysql.binlog.event.Event eve
...
-                    tableMetadata.getTableName(),
-                    bulkAction,
-                    primaryKeys,
-                    eventTimestampMillis,
-                    eventTimestampMillis,
-                    event.getHeader().getEventType());
-            pipelineEvents.add(pipelineEvent);
+        for (Map.Entry<OpenSearchBulkActions, List<Serializable[]>> entry : bulkActionToRowDataMap.entrySet()) {
+            final OpenSearchBulkActions bulkAction = entry.getKey();
+            final List<Serializable[]> rows = entry.getValue();
```
dinujoh, line 360: We should process all the events in order. Here looks like we process INDEX and DELETE operations separately. We might have a race condition where the same key might exist in INDEX and DELETE operation.

```
k1-> k2; k3->k1

Index keys: k2, k1
Delete keys: k1, k3
```
k1 is in both index and delete.

dinujoh: Instead of `Map<OpenSearchBulkActions, List<Serializable[]>>`, we should have List<Object> that contains OpenSearchBulkAction as property to process the events in order.

oeyh: Good point. I took your suggestions and arrange the row change in order and delete before index. So in the above case, the generated events will be: delete k1, index k2, delete k3, index k1.

dinujoh: Changes look good to me. Can you add unit test ?

dinujoh (approved):

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/stream/BinlogEventListener.java
dinujoh: Thanks for adding the unit test!

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/stream/BinlogEventListener.java
```diff
@@ -281,12 +280,29 @@ void handleUpdateEvent(com.github.shyiko.mysql.binlog.event.Event event) {
         // Check if a cascade action is involved
         cascadeActionDetector.detectCascadingUpdates(event, parentTableMap, tableMetadataMap.get(data.getTableId()));
 
-        // updatedRow contains data before update as key and data after update as value
-        final List<Serializable[]> rows = data.getRows().stream()
-                .map(Map.Entry::getValue)
-                .collect(Collectors.toList());
+        final TableMetadata tableMetadata = tableMetadataMap.get(data.getTableId());
+        final List<OpenSearchBulkActions> bulkActions = new ArrayList<>();
+        final List<Serializable[]> rows = new ArrayList<>();
```
chenqi0805, line 285: Nit: It seems to me more readable variable name to change rows -> rowUpdates. Same applies to row -> rowUpdate.

chenqi0805 (approved):

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/stream/BinlogEventListener.java
oeyh: Thanks! Will make this change in coming PRs for RDS source.

PR #5203, opened 2024-11-18
some fixes to jira source

Galactus22625:
### Description
Commit: Prevent Polling when there is already a leader scheduler
Stop race condition which causes jira pipeline to infinitely write to open search on loop because lastpolltime doesnt get updated properly

Commit: verify internet address everytime we invoke restapi
addresses appsec finding in osc-02 where we have a (TOCATOU) time of checkout time of use vulnerability

Commit: re-parallelize crawler call
re-parallelizing initial pagination and child partition so that we don’t have high latency at the start where we wait to crawl all pages before we start writing documents

 
### Issues Resolved
 
### Check List


data-prepper-plugins/saas-source-plugins/jira-source/src/main/java/org/opensearch/dataprepper/plugins/source/jira/JiraIterator.java
```diff
@@ -40,12 +45,41 @@ public JiraIterator(final JiraService service,
...
+            }
+        }
+        return isRunning;
+    }
+
+    private void startCrawlerThreads() {
+        futureList.add(crawlerTaskExecutor.submit(() ->
+            service.getJiraEntities(sourceConfig, lastPollTime, itemInfoQueue), false));
+    }
+
```
kkondaka, line 82: These changes look good but I do not see corresponding changes to the JavaIteratorTest.java file. When using futures testing is very important as we realized with Lambda code. Please make sure all these changes have corresponding tests

kkondaka (approved):

PR #5256, opened 2024-12-12
Add metrics to stream acknowledgement manager

sb2k16:
### Description
There appears to be an issue with the DocumentDB source plugin in DataPrepper where, if the host is restarted, the pipeline is unable to continue consuming stream data. Upon further investigation, this occurs because the stream lease in the source coordination store is removed, preventing the pipeline from ingesting the affected stream partition.

To address this, we need metrics to monitor the cadence of checkpoints, acknowledgments, and lease extensions, providing better visibility into stream acknowledgment behavior.

This PR introduces additional metrics to the Stream Acknowledgment Manager, enabling improved visibility into checkpointing cadence, tracking both positive and negative acknowledgments, and monitoring when the lease for the current DataPrepper instance is extended.
 
### Check List

### DIFF
--- a/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamAcknowledgementManager.java
+++ b/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamAcknowledgementManager.java
@@ -18,6 +18,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.function.Consumer;
 
+import static org.opensearch.dataprepper.logging.DataPrepperMarkers.NOISY;
+
 
 public class StreamAcknowledgementManager {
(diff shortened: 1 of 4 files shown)


data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamAcknowledgementManager.java
```diff
@@ -79,18 +99,22 @@ private void monitorAcknowledgment(final ExecutorService executorService, final
...
                     }
                 } else {
                     if (System.currentTimeMillis() - lastCheckpointTime >= checkPointIntervalInMs) {
                         LOG.debug("No records processed. Extend the lease of the partition worker.");
                         partitionCheckpoint.extendLease();
+                        this.noDataExtendLeaseCount.increment();
                         lastCheckpointTime = System.currentTimeMillis();
+                    } else {
+
                     }
```
dinujoh, line 118: Remove the empty else condition.

PR #5268, opened 2024-12-17
bug(opensearch-sink) Setting 'insecure' should override 'cert'

### LINKED ISSUE #5267: [BUG][opensearch exporter] Config option 'insecure' not honored when 'cert' is configured
**Describe the bug**
The `insecure` option should override `cert`, not the other way around.

**To Reproduce**
1. Setup an Opensearch with demo certificates
2. Generate a TLS certificate pem file for the configuration, which does not match the host name of opensearch
3. Configure prepper with opensearch exporter having both `cert` and `insecure` in config:
    ```yaml
    sink:
      opensearch:
        hosts: ["https://localhost:9200"]
        cert: path/to/wrong/hostname/cert.pem
        insecure: true
    ```
4. Observe that prepper cannot connect to Opensearch, it complains about hostname validation, even if `insecure` is set.
5. Remove the `cert` line and try again
6. Now prepper can connect to your opensearch

**Expected behavior**
The documentation says the following for the two settings:

> **cert (Optional)** : CA certificate that is pem encoded. Accepts both .pem or .crt. This enables the client to trust the CA that has signed the certificate that the OpenSearch cluster is using. Default is null.
> **insecure (Optional)**: A boolean flag to turn off SSL certificate verification. If set to true, CA certificate verification will be turned off and insecure HTTP requests will be sent. Default to false.

These two are by definition mutual exclusive. But the documentation does not talk about their relationship or the fact that `cert` will override `insecure`. The principle of least surprise would be that `insecure: true` overrides the presence of `cert`, not the other way around. You'd also expect a warning in the log whenever this is the case.

**Screenshots**
N/A

- Kubernetes
- Opensearch helm chart
- Data prepper helm chart
- Opensearch demo certificates

**Additional context**
The k8s service has a different name than the CN in auto generated demo certificates. Since the Data prepper is configured to talk to the servicename, there is a host name verification error when using the `pem` cert from opensearch.

janhoy:
### Description
This PR flips the precedence between `cert` and `insecure` config options so that `insecure: true` overrides the presence of `cert`.

### Issues Resolved
Resolves #5267
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch/README.md
+++ b/data-prepper-plugins/opensearch/README.md
@@ -105,7 +105,7 @@ pipeline:
 
 - `hosts`: A list of IP addresses of OpenSearch nodes.
 
-- `cert`(optional): CA certificate that is pem encoded. Accepts both .pem or .crt. This enables the client to trust the CA that has signed the certificate that the OpenSearch cluster is using.
+- `cert`(optional): CA certificate that is pem encoded. Accepts both `.pem` or `.crt`. This enables the client to trust the CA that has signed the certificate that the OpenSearch cluster is using. This setting has no effect if `insecure` is set to `true`.
 Default is null.
 
 - `aws_sigv4`: A boolean flag to sign the HTTP request with AWS credentials. Only applies to Amazon OpenSearch Service. See [security](security.md) for details. Default to `false`.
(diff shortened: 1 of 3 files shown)


data-prepper-plugins/opensearch/README.md
```diff
@@ -118,7 +118,7 @@ Default is null.
 
 - `aws_sts_header_overrides`: An optional map of header overrides to make when assuming the IAM role for the sink plugin.
 
-- `insecure`: A boolean flag to turn off SSL certificate verification. If set to true, CA certificate verification will be turned off and insecure HTTP requests will be sent. Default to `false`.
+- `insecure`: A boolean flag to turn off SSL certificate verification. If set to true, CA certificate verification will be turned off and insecure HTTP requests will be sent. Setting this will override any `cert` configured. Default to `false`.
```
janhoy, line 121: I added the same documentation clarification on the *source* side as on the *sink* side. Peeking in the configuration code for source, it appears that `insecure: true` already overrides `cert`, so this is just a clarification.

dlvenable (approved):
Thank you @janhoy for this contribution to the project!

san81 (approved):

PR #5320, opened 2025-01-09
lambda processor should retry for certain class of exceptions

srikanthjg:
### Description
Categorize exceptions as retryable and non-retryable for lambda plugin.
It should retry with backoff for certain number of times.

### Issue
#5340 
 
### Check List


data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/util/LambdaRetryStrategy.java
```diff
@@ -0,0 +1,190 @@
...
+                    429  // Too Many Requests (often used as "throttling" for Lambda)
+            )
+    );
+
+
+    public static boolean isRetryable(final InvokeResponse response) {
+        if(response == null) return false;
+        int statusCode = response.statusCode();
+        // Example logic: 429 (Too Many Requests) or 5xx => retry
+        return statusCode == 429 || (statusCode >= 500 && statusCode < 600);
```
kkondaka, line 119: why have 429 in multiple places?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/util/LambdaRetryStrategy.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    public static InvokeResponse retryOrFail(
+            final LambdaAsyncClient lambdaAsyncClient,
+            final Buffer buffer,
+            final LambdaCommonConfig config,
+            final InvokeResponse previousResponse,
+            final Logger LOG
+    ) {
+        int maxRetries = config.getClientOptions().getMaxConnectionRetries();
```
kkondaka, line 150: Is max connection retries same as mac lambda retries? Because it is possible for a connection to be established immediately but failed due to server errors? Or are we trying to use the same config parameter for both?

srikanthjg: i guess it was a bad naming for that field. The corresponding field is "max_retries", i am re-using that. I hope that is ok? I think there was another school of that we should indefinitely retry, but because this is a processor i thought it is better to retry for only certain number of times. We could also keep a different config for this and have a big but finite number assigned to this as default.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
```diff
@@ -98,12 +108,6 @@ public InvokeRequest getRequestPayload(String functionName, String invocationTyp
       return null;
     }
 
-    try {
```
san81, line 101: Any advantage in moving this to another public method? One disadvantage I see is that the caller needs to know and must follow that he should call `completeCodec` first before he call `getRequestPayload` method - right?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/util/LambdaRetryStrategy.java
```diff
@@ -0,0 +1,121 @@
...
+ */
+public final class LambdaRetryStrategy {
+
+    private LambdaRetryStrategy() {
+    }
+
+    /**
+     * Possibly a set of “bad request” style errors which might fall
+     */
+    private static final Set<Integer> BAD_REQUEST_ERRORS = new HashSet<>(
```
san81, line 33: better to use `Set.of()` method instead of HashSet with Arrays.asList within it.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/util/LambdaRetryStrategy.java
```diff
@@ -0,0 +1,121 @@
...
+     * This can be considered retryable if you want to handle the throttling scenario.
+     */
+    private static final Set<Integer> TIMEOUT_ERRORS = new HashSet<>(
+            Arrays.asList(
+                    408, // Request Timeout
+                    429  // Too Many Requests (often used as "throttling" for Lambda)
+            )
+    );
+
+    public static boolean isRetryable(final int statusCode) {
```
san81, line 77: I am not seeing any references to this method. is this unused (other than tests)? If it is unused, we can remove it. If we have to keep it, then I would recommend changing the name to `isRetryableStatuscode` to be inline with the other method `isRetryableException`

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/common/client/LambdaClientFactoryTest.java
```diff
@@ -8,21 +8,40 @@
...
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.spy;
 
 @ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
```
san81, line 44: I would recommend to avoid this if possible. It forces to avoid unwanted stubbing. If things in setup method are not needed then we can repeat them in corresponding test and avoid stubbing in the setup method.

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/utils/CountingHttpClient.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public CompletableFuture<Void> execute(AsyncExecuteRequest request) {
+        requestCount.incrementAndGet();
+        return delegate.execute(request);
+    }
+
+    @Override
+    public void close() {
+        delegate.close();
```
san81, line 26: close should decrement the counter?

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/utils/CountingHttpClient.java
```diff
@@ -0,0 +1,33 @@
+package org.opensearch.dataprepper.plugins.lambda.utils;
+
+import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
+import software.amazon.awssdk.http.async.AsyncExecuteRequest;
+import software.amazon.awssdk.utils.CompletableFutureUtils;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class CountingHttpClient implements SdkAsyncHttpClient {
```
san81, line 10: This class is only for testcases? I don't see any references to this class

srikanthjg: yes this is only for test. The reason is that, the sdk retry is handled internally within sdk. I had no way of verify the call. This creates a client where i can check the calls explicitly. If there is any other way please let me know, i can try that out.

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/common/client/LambdaClientFactoryTest.java
srikanthjg: ok sure.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/util/LambdaRetryStrategy.java
srikanthjg: yes, i am not currently using it. It was part of my previous implementation where i had a manual retry logic. I can remove it now.

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/utils/CountingHttpClient.java
srikanthjg: using this class only for testing. i will add it but it wont be required/used.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
srikanthjg: i dont think we are allowed to have private methods in interface(Buffer interface). That is the reason for using public here.

san81: why do we want to have it as a separate method instead of keeping this code at where it is?

srikanthjg: i felt that complete codec should not be associated with `getRequestPayload` . completeCodec should be independent of request. completeCodec should be done as soon as the buffer is done being added. This is just my thoughts. Please let me know what you think.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/util/LambdaRetryStrategy.java
srikanthjg: addressed this.

san81 (approved):

kkondaka (approved):

PR #5354, opened 2025-01-24
Add stateful buffer for lambda sink

### LINKED ISSUE #5353: [Enhancement] Add Stateful Buffering to LambdaSink
## Summary
Currently, the LambdaSink sends all incoming records immediately to AWS Lambda, causing multiple small invocations if thresholds are not met. We want to keep a stateful (persistent) buffer in the sink so it only flushes full batches immediately (when thresholds are exceeded) and persists any partial (incomplete) batch until more events arrive or until the sink shuts down. This ensures fewer, larger Lambda invocations and avoids prematurely flushing partial data.

Persisting the last incomplete batch should meet the following conditions
 (a) becomes full or 
 (b) the sink shuts down. 

### Persistent Buffer
A single buffer (persistentBuffer) accumulates events across multiple doOutput() calls.
Only when size or event-count thresholds are reached do we treat that batch as “full” and flush it to Lambda.
The partial batch remains in memory until it either becomes full or the sink shuts down.

N-1 “full” buffers get flushed immediately,
The Nth (partial) buffer remains in memory until the next doOutput() call or shutdown().

### Ensure Thread-Safety

When using persistent buffer, make sure we dont hit race conditions when multiple threads might write to buffer.

srikanthjg:
### Description
Add Stateful Buffering to LambdaSink
 
### Issues Resolved
Resolves #5353
 
### Check List

### DIFF
--- a/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
+++ b/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
@@ -38,7 +38,6 @@ import software.amazon.awssdk.services.lambda.LambdaAsyncClient;
 import software.amazon.awssdk.services.lambda.model.InvokeResponse;
 
 import java.net.HttpURLConnection;
-import java.nio.channels.AsynchronousFileChannel;
 import java.time.Duration;
 import java.util.Collection;
 import java.util.ArrayList;
--- a/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
+++ b/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
@@ -248,8 +248,6 @@ public class LambdaSink extends AbstractSink<Record<Event>> {
       releaseEventHandles(failedRecords, false);
     }
   }
-
-
     /*
      * Release events per batch
      */
@@ -266,11 +264,7 @@ public class LambdaSink extends AbstractSink<Record<Event>> {
     }
 
     private void flushBuffers(final List<Buffer> buffersToFlush) {
-        // Combine all their records for a single call to sendRecords
-        List<Record<Event>> combinedRecords = new ArrayList<>();
-        for (Buffer buf : buffersToFlush) {
-            combinedRecords.addAll(buf.getRecords());
-        }
+
 
         Map<Buffer, CompletableFuture<InvokeResponse>> bufferToFutureMap;
         try {
@@ -281,6 +275,12 @@ public class LambdaSink extends AbstractSink<Record<Event>> {
             );
         } catch (Exception e) {
             LOG.error(NOISY, "Error sending buffers to Lambda", e);
+            // Note: if there are N buffers and only N-1 buffers are full,
+            // we only need to propagate N-1 buffer records as failure
+            List<Record<Event>> combinedRecords = new ArrayList<>();
+            for (Buffer buf : buffersToFlush) {
+                combinedRecords.addAll(buf.getRecords());
+            }
             handleFailure(combinedRecords, e, HttpURLConnection.HTTP_INTERNAL_ERROR);
             return;
         }
--- a/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
+++ b/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
@@ -38,6 +38,7 @@ import software.amazon.awssdk.services.lambda.LambdaAsyncClient;
 import software.amazon.awssdk.services.lambda.model.InvokeResponse;
 
 import java.net.HttpURLConnection;
+import java.nio.channels.AsynchronousFileChannel;
 import java.time.Duration;
 import java.util.Collection;
 import java.util.ArrayList;
@@ -46,6 +47,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static org.opensearch.dataprepper.logging.DataPrepperMarkers.NOISY;
 import static org.opensearch.dataprepper.plugins.lambda.common.LambdaCommonHandler.isSuccess;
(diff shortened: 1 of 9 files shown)


data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
```diff
@@ -75,6 +78,13 @@ public class LambdaSink extends AbstractSink<Record<Event>> {
     private final OutputCodecContext outputCodecContext;
     private volatile boolean sinkInitialized;
     private DlqPushHandler dlqPushHandler = null;
+    final int maxEvents;
+    final long maxBytes;
+    final Duration maxCollectTime;
+
+    // The partial buffer that may not yet have reached threshold.
+    // Access must be synchronized
+    private Buffer statefulBuffer;
```
san81, line 87: This should handle concurrent options

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
```diff
@@ -249,4 +264,65 @@ private void releaseEventHandles(Collection<Record<Event>> records, boolean succ
             }
         }
     }
+
+    private synchronized void flushBuffers(final List<Buffer> buffersToFlush) {
+        // Combine all their records for a single call to sendRecords
+        List<Record<Event>> combinedRecords = new ArrayList<>();
+        for (Buffer buf : buffersToFlush) {
+            combinedRecords.addAll(buf.getRecords());
+        }
```
san81, line 273: I may be not clear on the overall approach here but Why are we combining the buffers here?

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSinkTest.java
```diff
@@ -250,43 +375,14 @@ public void testHandleFailure_WithoutDlq() throws Exception {
...
-            lambdaCommonHandler.when(() ->
-                    LambdaCommonHandler.sendRecords(anyList(), any(LambdaSinkConfig.class), any(LambdaAsyncClient.class), any(OutputCodecContext.class))).thenReturn(Map.of(currentBufferPerBatch, future));
-
-            lambdaCommonHandler.when(() ->
-                    LambdaCommonHandler.isSuccess(any(InvokeResponse.class))).thenReturn(true);
-            lambdaSink.doOutput(records);
-
-            verify(numberOfRecordsFailedCounter, times(1)).increment(1);
+    // Utility to set private fields
+    private static void setPrivateField(Object target, String fieldName, Object value) {
```
san81, line 379: There is a `data-prepper-test-common` module that has these kind of common reusable methods used for test cases. We can reuse these utilities from there and avoid adding these methods in every plugin.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
srikanthjg: yes, all i have used "synchronized" on the methods where this is used. In s3 sink, they use reentrant lock to handle something similar.

san81 (approved):
Looks good to me

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
srikanthjg: i have addressed the comment, this might not be needed for flushing to lambda but needed for handlingFailures. Lets say the "full buffer" records have some failures, only those have to be tagged and forwarded(failure behaviour), the remaning records should persist for the next doOutput should not go through the failure handling mechanism. I have refactored the lambdaCommonHandler to take in Buffer and moved this logic within the catch.

san81 (approved):

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
```diff
@@ -249,4 +280,66 @@ private void releaseEventHandles(Collection<Record<Event>> records, boolean succ
...
+                    lambdaSinkConfig,
+                    lambdaAsyncClient,
+                    buffersToFlush
+            );
+        } catch (Exception e) {
+            LOG.error(NOISY, "Error sending buffers to Lambda", e);
+            // Note: if there are N buffers and only N-1 buffers are full,
+            // we only need to propagate N-1 buffer records as failure
+            List<Record<Event>> combinedRecords = new ArrayList<>();
+            for (Buffer buf : buffersToFlush) {
```
kkondaka, line 299: it is possible that not every buffer in buffersToFlush has failed, right?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
```diff
@@ -138,56 +155,72 @@ public void doInitialize() {
...
     @Override
     public void doOutput(final Collection<Record<Event>> records) {
-        if (records.isEmpty()) {
+        if (!sinkInitialized) {
+            LOG.warn("LambdaSink doOutput called before initialization");
             return;
         }
-
-        Map<Buffer, CompletableFuture<InvokeResponse>> bufferToFutureMap = new HashMap<>();
+        reentrantLock.lock();
```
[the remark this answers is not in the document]
srikanthjg: i initially tried it without the lock but hit concurrency issues. After adding the lock it was solved. Looks like an additional lock is needed. If there are any other suggested ways, i can give that a shot.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/sink/LambdaSink.java
srikanthjg: yes but thats how we handle failures in lambda processor as well. This can come in as an enhancement in the next PR? 
[processor-link](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java#L164)

kkondaka (approved):

kkondaka (approved):

san81 (approved):

PR #5359, opened 2025-01-27
Upgrade OTEL version to 1.2

kkondaka:
### Description
Update OTEL version to 1.2
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoCodec.java
+++ b/data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoCodec.java
@@ -12,7 +12,6 @@ import io.opentelemetry.proto.collector.metrics.v1.ExportMetricsServiceRequest;
 import io.opentelemetry.proto.collector.logs.v1.ExportLogsServiceRequest;
 import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest;
 import io.opentelemetry.proto.common.v1.AnyValue;
-import io.opentelemetry.proto.common.v1.InstrumentationLibrary;
 import io.opentelemetry.proto.common.v1.InstrumentationScope;
 import io.opentelemetry.proto.common.v1.KeyValue;
 import io.opentelemetry.proto.logs.v1.LogRecord;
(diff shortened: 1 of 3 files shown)


KarstenSchnitter:
This changeset is also part of #5322. @kkondaka what is your opinion on backwards compatibility to OTel data containing instrumentation library fields?

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoCodec.java
```diff
@@ -92,6 +89,7 @@ public class OTelProtoCodec {
     static final String INSTRUMENTATION_LIBRARY_VERSION = "instrumentationLibrary.version";
     static final String STATUS_CODE = "status.code";
     static final String STATUS_MESSAGE = "status.message";
+    static final String ATTRIBUTES_KEY = "attributes";
```
KarstenSchnitter, line 92: Shouldn't there be something like `instrumentationScope.attributes`? These attributes should be handled similar to resource, signal, or event attributes.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoCodec.java
```diff
@@ -1202,6 +1103,9 @@ public static Map<String, Object> getInstrumentationScopeAttributes(final Instru
         if (!instrumentationScope.getVersion().isEmpty()) {
             instrumentationScopeAttr.put(INSTRUMENTATION_SCOPE_VERSION, instrumentationScope.getVersion());
         }
+        if (!instrumentationScope.getAttributesList().isEmpty()) {
+            instrumentationScopeAttr.put(ATTRIBUTES_KEY, OTelProtoCodec.unpackKeyValueListLog(instrumentationScope.getAttributesList()));
```
KarstenSchnitter, line 1107: I think, the attributes should be put under `instumentationsScope.attributes` and filtered similar to all other attributes: Prefix + Dedotting. Won't this proposed handling add them in a very confusing location and just named `attributes`?

KarstenSchnitter (changes requested):
Thanks for providing that change. I would like to see some integration test using the old protobuf data format. Also, I have questions on the handling of instrumentation scope attributes. See the comments inline for the latter.

kkondaka:
@KarstenSchnitter @dlvenable the tests are not changed (and failing ) because I wanted to get your opinion on this. Forgot to make it a draft PR.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoCodec.java
kkondaka: I am comparing with the output generated by OTEL and it looks like they are putting it under "scope" like this 
```
"scope": {
        "name": "my.library",
        "attributes": {
          "my.scope.attribute": "some scope attribute"
        },
        "version": "1.0.0"
      },
```
So, it matches with it.

PR #5361, opened 2025-01-27
Implementation of sqs-common plugin and refactored sqs and s3 source

jmsusanto:
### Description

This PR integrates the sqs-common plugin. The sqs-source has been completely refactored to leverage sqs-common, while only the SqsService has been updated for the s3-source. A full refactor of the s3-source will be addressed in a future PR, but the scope has been intentionally limited for now.

### Issue Resolved
https://github.com/opensearch-project/data-prepper/issues/1049
 
### Check List


data-prepper-plugins/s3-source/src/integrationTest/java/org/opensearch/dataprepper/plugins/source/s3/SqsServiceIT.java
```diff
@@ -154,8 +155,7 @@ public void test_sqsService(int numWorkers) throws IOException {
     }
 
     private void clearSqsQueue() {
-        Backoff backoff = Backoff.exponential(SqsService.INITIAL_DELAY, SqsService.MAXIMUM_DELAY).withJitter(SqsService.JITTER_RATE)
-                .withMaxAttempts(Integer.MAX_VALUE);
+        Backoff backoff = SqsBackoff.createExponentialBackoff();
```
san81, line 158: Is there a possibility that this Backoff strategy could be different for S3 source vs sqs source?

jmsusanto: The Backoff strategy is currently the same for both S3 and SQS sources, as both handle SQS polling. If needed, we can extend SqsBackoff in the future to support options like linear backoff.

data-prepper-plugins/sqs-source/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/SqsWorker.java
```diff
@@ -78,24 +55,16 @@ public SqsWorker(final Buffer<Record<Event>> buffer,
                      final PluginMetrics pluginMetrics,
                      final SqsEventProcessor sqsEventProcessor,
                      final Backoff backoff) {
-
-        this.sqsClient = sqsClient;
+        this.sqsWorkerCommon = new SqsWorkerCommon(sqsClient, backoff, pluginMetrics, acknowledgementSetManager);
```
san81, line 58: I thought, `SqsWorkerCommon` is a shared instance between all the workers. It looks like you are instantiating `SqsWorkerCommon` per worker. Is that the intention? I am not sure you want per worker an instance of `SqsWorkerCommon` like for example, if you take `isStopped` flag, should be shared between all the workers. You don't want each worker to have their own `isStopped` state - right?  I thought, that is why you made `isStopped` volatile

jmsusanto: Good catch, you're absolutely right. In a future PR, I'm planning to support multiple SQS clients for cases where we have queues from different regions. In most scenarios (when the configuration only includes queues from a single region), only one SqsWorkerCommon instance will be created. Generally, we'll have one SqsWorkerCommon per region specified in the configuration.

jmsusanto: SqsWorkerCommon currently handles polling from an SqsClient, which is tied to a specific AWS region. If customers need to poll from SQS queues in different regions, we’ll need multiple SqsClient instances, and therefore, multiple SqsWorkerCommon instances.

Alternatively, we could remove the SqsClient from SqsWorkerCommon and pass it dynamically to the polling methods there. This makes SqsWorkerCommon region-agnostic.

I am planning to support multiple regions in an upcoming PR, so I’m thinking it’s better to address this there instead

san81: I am Ok, as long as it is addressed in the next PR

san81 (approved):
Thank you for reducing the redundant code

sb2k16 (approved):

PR #5409, opened 2025-02-03
SQS Source: Add on_error Config, Multi-Region Support, and sqsMessageDelayTimer Metric for Auto-Scaling

jmsusanto:
### Description
Added the following:
- Introduced an on_error configuration parameter that lets users choose whether to delete or retain messages upon processing failure.

- Enabled support for SQS queues across multiple regions within a single configuration by instantiating one client per region.

- Added the sqsMessageDelayTimer metric to assist with auto-scaling in OSI.

- modified README
 
### Issues Resolved
[Resolves #[Issue number to be closed when this PR is merged]
](https://github.com/opensearch-project/data-prepper/issues/1049) 

### Check List

### DIFF
--- a/data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/OnErrorOption.java
+++ b/data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/OnErrorOption.java
@@ -1,6 +1,11 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
  */
 
 package org.opensearch.dataprepper.plugins.source.sqs.common;
--- a/data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/OnErrorOption.java
+++ b/data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/OnErrorOption.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.dataprepper.plugins.source.sqs.common;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public enum OnErrorOption {
+    DELETE_MESSAGES("delete_messages"),
+    RETAIN_MESSAGES("retain_messages");
+
+    private static final Map<String, OnErrorOption> OPTIONS_MAP = Arrays.stream(OnErrorOption.values())
+            .collect(Collectors.toMap(
+                    value -> value.option,
+                    value -> value
+            ));
+
+    private final String option;
+
+    OnErrorOption(final String option) {
+        this.option = option;
+    }
+
+    @JsonCreator
+    static OnErrorOption fromOptionValue(final String option) {
+        return OPTIONS_MAP.get(option);
+    }
+}
--- a/data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/SqsWorkerCommon.java
+++ b/data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/SqsWorkerCommon.java
@@ -42,7 +42,6 @@ public class SqsWorkerCommon {
     public static final String SQS_VISIBILITY_TIMEOUT_CHANGED_COUNT_METRIC_NAME = "sqsVisibilityTimeoutChangedCount";
     public static final String SQS_VISIBILITY_TIMEOUT_CHANGE_FAILED_COUNT_METRIC_NAME = "sqsVisibilityTimeoutChangeFailedCount";
 
-    private final SqsClient sqsClient;
     private final Backoff standardBackoff;
     private final PluginMetrics pluginMetrics;
     private final AcknowledgementSetManager acknowledgementSetManager;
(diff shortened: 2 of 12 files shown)


data-prepper-plugins/sqs-source/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/SqsWorker.java
```diff
@@ -172,9 +172,13 @@ private List<DeleteMessageBatchRequestEntry> processSqsEvents(final List<Message
         }
 
         for (Message message : messages) {
-            Duration duration = Duration.between(Instant.ofEpochMilli(Long.parseLong(message.attributes().get(MessageSystemAttributeName.SENT_TIMESTAMP))), Instant.now());
-            sqsMessageDelayTimer.record(duration.isNegative() ? Duration.ZERO : duration); // Negative durations can occur if messages are processed immediately
-
+            final int receiveCount = Integer.parseInt(message.attributes().get(MessageSystemAttributeName.APPROXIMATE_RECEIVE_COUNT));
```
graytaylor0, line 175: If this attribute doesn't exist for some reason, will this default to 0, or will Integer.parseInt(null) throw an exception?

data-prepper-plugins/sqs-source/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/SqsWorker.java
```diff
@@ -172,9 +172,13 @@ private List<DeleteMessageBatchRequestEntry> processSqsEvents(final List<Message
...
+            if (receiveCount < 1) {
```
graytaylor0, line 176: This should be `receiveCount <= 1`, since the receive count will always be 1 when it's grabbed for the first time.

data-prepper-plugins/sqs-source/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/SqsWorker.java
```diff
@@ -172,9 +172,13 @@ private List<DeleteMessageBatchRequestEntry> processSqsEvents(final List<Message
...
+                Duration duration = Duration.between(
+                        Instant.ofEpochMilli(Long.parseLong(message.attributes().get(MessageSystemAttributeName.SENT_TIMESTAMP))),
+                        Instant.now());
+                sqsMessageDelayTimer.record(duration.isNegative() ? Duration.ZERO : duration); // can sometimes record negative durations if message is processed immediately
```
graytaylor0, line 180: Let's add this to a unit test

data-prepper-plugins/sqs-source/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/SqsWorker.java
jmsusanto: both of these system attributes will always exist since we make this receiveMessageRequest call:

       ReceiveMessageRequest.Builder requestBuilder = ReceiveMessageRequest.builder()
                .queueUrl(queueUrl)
                .attributeNamesWithStrings("All")
                .messageAttributeNames("All");

data-prepper-plugins/sqs-source/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/SqsWorker.java
jmsusanto: good catch, changing this right away

data-prepper-plugins/sqs-source/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/SqsWorker.java
jmsusanto: added this

data-prepper-plugins/sqs-source/src/test/java/org/opensearch/dataprepper/plugins/source/sqs/SqsWorkerTest.java
```diff
@@ -330,4 +330,45 @@ void stop_should_set_isStopped_and_call_stop_on_sqsWorkerCommon() {
         worker.stop();
         verify(sqsWorkerCommon, times(1)).stop();
     }
+
+    @Test
+    void processSqsMessages_should_record_sqsMessageDelayTimer_when_approximateReceiveCount_less_than_or_equal_to_one() throws IOException {
+        final long sentTimestampMillis = Instant.now().minusSeconds(5).toEpochMilli();
+        final Message message = Message.builder()
```
graytaylor0, line 337: Nit: You could mock this message instead of building it.

graytaylor0 (dismissed):

data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/OnErrorOption.java
```diff
@@ -0,0 +1,34 @@
+/*
```
san81, line 1: Please use new format file header 

```
/*
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 *
 */
```

data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/OnErrorOption.java
```diff
@@ -0,0 +1,34 @@
...
+            ));
+
+    private final String option;
+
+    OnErrorOption(final String option) {
+        this.option = option;
+    }
+
+    @JsonCreator
+    static OnErrorOption fromOptionValue(final String option) {
```
san81, line 31: This could possibly return a null value. Not sure if you need to consider that at the caller end.

data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/SqsWorkerCommon.java
```diff
@@ -56,18 +55,17 @@ public class SqsWorkerCommon {
     private final Counter sqsVisibilityTimeoutChangedCount;
     private final Counter sqsVisibilityTimeoutChangeFailedCount;
 
-    public SqsWorkerCommon(final SqsClient sqsClient,
-                           final Backoff standardBackoff,
+    public SqsWorkerCommon(final Backoff standardBackoff,
```
san81, line 58: I think, `standardBackoff` logic should be per sqsClient. You don't want to share one instance of the backoff across different sqs clients because it maintains state to track retry attempts and compute delays

san81 (changes requested):
I think, we shouldn't share Backoff object instance between different SQSClients because it maintains state to track retry attempts and compute delays

data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/OnErrorOption.java
jmsusanto: Current implementation will not allow for values other than "RETAIN_MESSAGES" or "DELETE_MESSAGES". Values such as None won't work. Doing `on_error: null `will treat as if that parameter is not overridden and use the default.

data-prepper-plugins/sqs-common/src/main/java/org/opensearch/dataprepper/plugins/source/sqs/common/SqsWorkerCommon.java
jmsusanto: I think it would make more sense to have one Backoff instance on the queue level. This would mean that workers of the same queue will share Backoff.

san81 (approved):

graytaylor0 (approved):

PR #5449, opened 2025-02-20
S3-Source Bug Fix: Handle null object size for non S3-Create Events

### LINKED ISSUE #5448: [BUG] NullPointerException on S3 Delete Event Due to Null Object Size
**Describe the bug**
When processing S3 delete events, data-prepper throws a NullPointerException because the object size is null. 

**To Reproduce**
1. Configure an S3 bucket with SQS Event Notifcations and ObjectDelete enabled.
2. Delete a file from the monitored S3 bucket.
3. The following error will occur:
`[s3-source-sqs-1] ERROR org.opensearch.dataprepper.plugins.source.s3.SqsWorker - Unable to process SQS messages. Processing error due to: Cannot invoke "java.lang.Long.longValue()" because the return value of "org.opensearch.dataprepper.plugins.source.s3.S3EventNotification$S3ObjectEntity.getSizeAsLong()" is null
`

Pipeline config:
```
  S3-pipeline:
    source:
      s3:
        notification_type: sqs
        codec:
          csv:
        compression: none
        sqs:
          queue_url: "https://sqs.us-east-2.amazonaws.com/someQueue"
        aws:
          region: "us-east-2"
          sts_role_arn: "someArn"

    sink:
      - stdout:
```

**Expected behavior**  
The application should gracefully handle S3 delete events without throwing an exception. The object size should default to 0 when not present in the event notification.

**Proposed Fix**
- Add a null check for `getSizeAsLong()` in the `ParsedMessage` class.
- Default object size to `0L` if the value is null.

 - OS: MacOS 14.7 Sonoma
 - Data Prepper 2.10.2

jmsusanto:
### Description
This PR fixes a NullPointerException that occurs when processing non S3-create events such as S3-delete due to the absence of an object size in the event notification. The `getSizeAsLong()` method can return null for delete events, which caused the application to throw an error.
 
### Issues Resolved
Resolves #5448
 

### DIFF
--- a/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/parser/ParsedMessage.java
+++ b/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/parser/ParsedMessage.java
@@ -37,7 +37,8 @@ public class ParsedMessage {
          // S3EventNotification contains only one S3EventNotificationRecord
         this.bucketName = notificationRecords.get(0).getS3().getBucket().getName();
         this.objectKey = notificationRecords.get(0).getS3().getObject().getUrlDecodedKey();
-        this.objectSize = notificationRecords.get(0).getS3().getObject().getSizeAsLong();
+        Long size = notificationRecords.get(0).getS3().getObject().getSizeAsLong();
+        this.objectSize = (size != null) ? size : 0L;
         this.eventName = notificationRecords.get(0).getEventName();
         this.eventTime = notificationRecords.get(0).getEventTime();
         this.failedParsing = false;
--- a/data-prepper-plugins/s3-source/src/test/java/org/opensearch/dataprepper/plugins/source/s3/parser/ParsedMessageTest.java
+++ b/data-prepper-plugins/s3-source/src/test/java/org/opensearch/dataprepper/plugins/source/s3/parser/ParsedMessageTest.java
@@ -134,6 +134,13 @@ class ParsedMessageTest {
             assertThat(actualString, containsString(messageId));
             assertThat(actualString, containsString(testDecodedObjectKey));
         }
+
+        @Test
+        void test_parsed_message_with_null_object_size_defaults_to_zero() {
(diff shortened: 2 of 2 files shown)


data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/parser/ParsedMessage.java
```diff
@@ -37,7 +37,8 @@ public ParsedMessage(final Message message, final boolean failedParsing) {
          // S3EventNotification contains only one S3EventNotificationRecord
         this.bucketName = notificationRecords.get(0).getS3().getBucket().getName();
         this.objectKey = notificationRecords.get(0).getS3().getObject().getUrlDecodedKey();
-        this.objectSize = notificationRecords.get(0).getS3().getObject().getSizeAsLong();
+        Long size = notificationRecords.get(0).getS3().getObject().getSizeAsLong();
```
san81, line 40: Could you please add a test case?

jmsusanto: added a unit test

jmsusanto:
> I actually thought we had code that filtered out the events so that only created events were processed.
@dlvenable  The issue is that we're calling final `Collection<ParsedMessage> s3MessageEventNotificationRecords = sqsMessageParser.parseSqsMessages(sqsMessages);` before the code that handles non-create events. The code that you're referencing acts on the ParsedMessages, but the error occurs when we create these ParsedMessages, so we never get there.

graytaylor0:
Were you able to confirm that with this change a DeleteEvent did not hit an error and the SQS message for it was deleted?

jmsusanto:
> Were you able to confirm that with this change a DeleteEvent did not hit an error and the SQS message for it was deleted?

@graytaylor0 After this change, we don't hit an error and DeleteEvent messages get deleted. The issue was primarily during the creation of the ParsedMessages.

san81 (approved):

PR #5459, opened 2025-02-24
ENH: error message in pipeline and plugin deserialization

chenqi0805:
### Description
This PR enhances the error message in pipeline and plugin deserialization.
* fix null value handling in DataPrepperDurationParser
before
```
waf-access-log-pipeline.source.s3: caused by: Parameter "sqs.visibility_timeout" for plugin "s3" is invalid: text
```
after
```
waf-access-log-pipeline.source.s3: caused by: Parameter "sqs.visibility_timeout" for plugin "s3" is invalid: Durations must use either ISO 8601 notation or simple notations for seconds (60s) or milliseconds (100ms). Whitespace is ignored.
```
* simplify pipeline and plugin deserialization error message from Jackson by adding DataPrepperDeserializationProblemHandler
   1. handleWeirdStringValue
before:
```
waf-access-log-pipeline.source.s3: caused by: Parameter "sqs.visibility_duplication_protection" for plugin "s3" is invalid: Cannot deserialize value of type `java.lang.Boolean` from String "xy": only "true" or "false" recognized
 at [Source: UNKNOWN; byte offset: #UNKNOWN]
```
after:
```
waf-access-log-pipeline.source.s3: caused by: Parameter "sqs.visibility_duplication_protection" for plugin "s3" is invalid: Cannot deserialize value of type `java.lang.Boolean` from String "xy": only "true" or "false" recognized
```
    2. handleMissingInstantiator
before:
```
waf-access-log-pipeline.processor.truncate: caused by: Parameter "entries.null" for plugin "truncate" is invalid: Cannot construct instance of `org.opensearch.dataprepper.plugins.processor.truncate.TruncateProcessorConfig$Entry` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('rand')
```
after:
```
waf-access-log-pipeline.processor.truncate: caused by: Parameter "entries.null" for plugin "truncate" is invalid: Cannot deserialize 'rand' into 'org.opensearch.dataprepper.plugins.processor.truncate.TruncateProcessorConfig$Entry'
```
    3. handleUnexpectedToken
before:
```
waf-access-log-pipeline.source.s3: caused by: Parameter "sqs.visibility_duplication_protection" for plugin "s3" is invalid: Cannot deserialize value of type `java.lang.Boolean` from Array value (token `JsonToken.START_ARRAY`)
```
after:
```
waf-access-log-pipeline.source.s3: caused by: Parameter "sqs.visibility_duplication_protection" for plugin "s3" is invalid: Cannot deserialize value of type `java.lang.Boolean` from Array value.
```

### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-core/src/integrationTest/java/org/opensearch/dataprepper/plugin/DefaultPluginFactoryIT.java
+++ b/data-prepper-core/src/integrationTest/java/org/opensearch/dataprepper/plugin/DefaultPluginFactoryIT.java
@@ -21,6 +21,7 @@ import org.opensearch.dataprepper.model.configuration.PluginSetting;
 import org.opensearch.dataprepper.model.plugin.InvalidPluginConfigurationException;
 import org.opensearch.dataprepper.model.plugin.NoPluginFoundException;
 import org.opensearch.dataprepper.model.source.Source;
+import org.opensearch.dataprepper.pipeline.parser.DataPrepperDeserializationProblemHandler;
 import org.opensearch.dataprepper.plugins.TestObjectPlugin;
 import org.opensearch.dataprepper.plugins.configtest.TestComponentWithConfigInject;
 import org.opensearch.dataprepper.plugins.configtest.TestDISourceWithConfig;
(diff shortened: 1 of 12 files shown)


chenqi0805:
> What are some examples of what certain errors based on the configuration will look like?

@graytaylor0 I updated the description with some typical examples

dlvenable (approved):
Thanks @chenqi0805! This is a nice improvement.

Did you look to see if Jackson provides a mechanism to get the path to the problematic values? It would be nice to say something like " ... in my-pipeline.source.s3.sqs`.

chenqi0805:
> Thanks @chenqi0805! This is a nice improvement.
> 
> Did you look to see if Jackson provides a mechanism to get the path to the problematic values? It would be nice to say something like " ... in my-pipeline.source.s3.sqs`.

@dlvenable Thanks for the call out. I do have another follow up PR in preparation: https://github.com/opensearch-project/data-prepper/pull/5460. It would be nice if we can replace `getParameterPath` with built-in Jackson mechanism. I will look into this further

Issue #5309, opened 2025-01-06
Support AWS Aurora/RDS PostgreSQL as source


oeyh:
**Is your feature request related to a problem? Please describe.**
As a Data Prepper user, I would like to have an rds source to load existing data and stream change events from RDS MySQL databases.

**Describe the solution you'd like**
For export (loading existing data), we can create a snapshot, export it to S3 and read the data from S3
For stream (streaming change events), we can create a Postgres logical replication stream to receive change events.

**Describe alternatives you've considered (Optional)**
Run SQL queries periodically through a JDBC driver to load existing and incremental data from the source database.

**Additional context**
Relevant issue for MySQL: https://github.com/opensearch-project/data-prepper/issues/4561

## Tasks

- [ ] Project setup, source configurations, skeleton code
- [ ] Export implementation 
- [ ] Stream implementation
- [ ] Checkpointing in both export and stream
- [ ] Secret rotation support
- [ ] Add E2E acknowledge support
- [ ] Add data type mapping
- [ ] Add plugin metrics
- [ ] Add aggregate metrics
- [ ] Add integration tests

PR #5494, opened 2025-03-03
Add table filters in Rds source configuration

oeyh:
### Description
Adds table filters in Rds source like this:
```
      # For MySQL
      tables:
        database: "database1"
        include:
          - "table1"
          - "table2"
        # exclude:
        #   - "table2"

      
      # For Postgres
      tables:
        database: "database1"
        include:
          - "schema1.table1"
          - "schema2.table1"
        # exclude:
        #   - "schema1.table2"
```
When no includes/excludes are specified, all tables under the database will be processed. With includes/excludes, specified tables will be included or excluded. Exclude option can override include option.

Corresponding changes are made to retrieve all table names and filter them. Also fix a bug in `BinlogEventListener` so that only CDC events from tables of interest will be processed. SchemaManager interface and implementation are refactored to process multiple tables in a single method.
 
### Issues Resolved
Contributes to #4561 and #5309 
 
### Check List


chenqi0805 (approved):

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/configuration/TableFilterConfigHelper.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class TableFilterConfigHelper {
+
+    /**
+     * This method applies the table filter configuration to the given set of table names.
+     *
+     * @param tableNames        The set of table names to be filtered
+     * @param tableFilterConfig The table filter configuration to be applied
+     */
+    public static void applyTableFilter(Set<String> tableNames, TableFilterConfig tableFilterConfig) {
```
san81, line 24: This method can be part of `TableFilterConfig` class itself.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/schema/MySqlSchemaManager.java
```diff
@@ -51,13 +54,25 @@ public MySqlSchemaManager(ConnectionManager connectionManager) {
...
-    public List<String> getPrimaryKeys(final String fullTableName) {
+    public Map<String, List<String>> getPrimaryKeys(final List<String> fullTableNames) {
+        final Map<String, List<String>> tableToPrimaryKeysMap = new HashMap<>();
+        try (final Connection connection = connectionManager.getConnection()) {
+            for (final String fullTableName : fullTableNames) {
+                tableToPrimaryKeysMap.put(fullTableName, getPrimaryKeysForTable(connection, fullTableName));
+            }
+            return tableToPrimaryKeysMap;
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to get primary keys for tables. ", e);
```
san81, line 65: It looks like, if you are failing to get primary key, you are not throwing any exception. This block will be executed only if you fail to get connection. May be message can say `failed to get connection while trying to get primary keys`  or something like that?

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/schema/MySqlSchemaManager.java
```diff
@@ -74,16 +89,26 @@ public List<String> getPrimaryKeys(final String fullTableName) {
...
-    public Map<String, String> getColumnDataTypes(final String fullTableName) {
+    @Override
+    public Map<String, Map<String, String>> getColumnDataTypes(final List<String> fullTableNames) {
+        final Map<String, Map<String, String>> tableToColumnDataTypesMap =  new HashMap<>();
+        try (Connection connection = connectionManager.getConnection()) {
+            for (final String fullTableName : fullTableNames) {
+                tableToColumnDataTypesMap.put(fullTableName, getColumnDataTypesForTable(connection, fullTableName));
+            }
+            return tableToColumnDataTypesMap;
+        } catch (Exception e) {
```
san81, line 100: do you want this catch block to catch generic `Exception`? if yes then you will be catching the RuntimeException thrown from `getColumnDataTypesForTable` and translating that exception into something else. May be better to have custom exceptions defined. One for connection error and other for failing to get dataTypes or primaryKeys etc? Overall, the exception is not lost but the message is changed.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/schema/MySqlSchemaManager.java
```diff
@@ -94,14 +119,33 @@ public Map<String, String> getColumnDataTypes(final String fullTableName) {
                 return columnsToDataType;
             } catch (final Exception e) {
                 LOG.error("Failed to get dataTypes for database {} table {}, retrying", database, tableName, e);
-                if (retry == NUM_OF_RETRIES) {
-                    throw new RuntimeException(String.format("Failed to get dataTypes for database %s table %s after " +
-                            "%d retries", database, tableName, retry), e);
+            }
+            applyBackoff();
+        }
+        throw new RuntimeException(String.format("Failed to get dataTypes for database %s table %s after " +
```
san81, line 125: You are throwing an exception here but not in the case of `getPrimaryKeysForTable` method. May be that is what you want but just calling out to make sure that is your intention.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/stream/StreamWorker.java
```diff
@@ -62,7 +62,7 @@ public void processStream(final StreamPartition streamPartition) {
             LOG.info("Connect to database to read change events.");
             replicationLogClient.connect();
         } catch (Exception e) {
-            LOG.warn("Error while connecting to replication stream, will retry.");
+            LOG.warn("Error while connecting to replication stream, will retry.", e);
```
san81, line 65: With e added now, it will print the entire stack trace I guess. May be worth adding a NOISY flag here?

oeyh: Thanks for this reminder! Added.

Do we add NOISY flag whenever we print the stack trace? The error here will likely be repeated, but only at a rate of one in every few minutes.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/configuration/TableFilterConfigHelper.java
oeyh: Made the suggested change.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/schema/MySqlSchemaManager.java
oeyh: Good catch. It should throw when failing to get primary key. I also took your suggestion below to separate the connection exception with other type of exceptions through custom exception.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/schema/MySqlSchemaManager.java
oeyh: I took this suggestion and throws different messages for connection errors and other errors.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/schema/MySqlSchemaManager.java
oeyh: You are right. Made the change in other places.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/schema/MySqlSchemaManager.java
oeyh: Also made similar changes to PostgresSchemaManager

san81 (approved):
Looks good to me

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/stream/StreamWorker.java
san81: Not necessarily but I guess, anything that is repeated won't really add value. Noisy will print a couple of times before it starts to block so good to have it for repeated logs

chenqi0805 (approved):

PR #5499, opened 2025-03-04
Give up partition when there is no Ack

dinujoh:
### Description
Give up partition when there is no Ack to avoid partition expiring. When there is no ACK, there is a code flow where partition checkpoint is not updated.
 
### Check List

### DIFF
--- a/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamAcknowledgementManager.java
+++ b/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamAcknowledgementManager.java
@@ -21,6 +21,7 @@ import java.util.function.Consumer;
 public class StreamAcknowledgementManager {
     private static final Logger LOG = LoggerFactory.getLogger(StreamAcknowledgementManager.class);
     private static final int CHECKPOINT_RECORD_INTERVAL = 50;
+    private static final int NO_ACK_PARTITION_TIME_OUT_SECONDS = 900; // 15 minutes
     private final ConcurrentLinkedQueue<CheckpointStatus> checkpoints = new ConcurrentLinkedQueue<>();
     private final ConcurrentHashMap<String, CheckpointStatus> ackStatus = new ConcurrentHashMap<>();
     private final AcknowledgementSetManager acknowledgementSetManager;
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamAcknowledgementManager.java
```diff
@@ -95,19 +97,23 @@ private void monitorAcknowledgment(final ExecutorService executorService, final
...
                         if (checkpointStatus.isNegativeAcknowledgement()) {
-                            // Give up partition and should interrupt parent thread to stop processing stream
-                            if (lastCheckpointStatus != null && lastCheckpointStatus.isPositiveAcknowledgement()) {
-                                checkpoint(lastCheckpointStatus.getResumeToken(), lastCheckpointStatus.getRecordCount());
-                            }
-                            LOG.warn("Acknowledgement not received for the checkpoint {} past wait time. Giving up partition.", checkpointStatus.getResumeToken());
-                            partitionCheckpoint.giveUpPartition();
-                            this.giveupPartitionCount.increment();
+                            LOG.warn("Negative Acknowledgement received for the checkpoint {}. Giving up partition.", checkpointStatus.getResumeToken());
+                            giveUpPartition(lastCheckpointStatus);
```
graytaylor0, line 103: Will the actual worker know if it's partition is given up?

dinujoh: yeah.. the parent thread will be interrupted and will be stopped

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamAcknowledgementManager.java
```diff
@@ -95,19 +97,23 @@ private void monitorAcknowledgment(final ExecutorService executorService, final
...
                             break;
+                        } else {
+                            // no ack received within timeout period (15 secs + 60 secs)
+                            if (!ackWaitDuration.minus(partitionAcknowledgmentTimeout)
+                                    .minusMillis(checkPointIntervalInMs).isNegative()) {
+                                LOG.warn("Acknowledgement not received for the checkpoint {} past wait time. Giving up partition.", checkpointStatus.getResumeToken());
+                                giveUpPartition(lastCheckpointStatus);
```
graytaylor0, line 110: Do you think it's better to give up in this case?

chenqi0805 (dismissed):

sb2k16 (approved):

data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamAcknowledgementManager.java
dinujoh: When no ack is received after timeout, then not sure if we will ever receive it. So we need to restart processing the partition from checkpoint. Updated the no ack timeout to 15 mins, so take care of extreme sink backed up case.

sb2k16 (approved):

graytaylor0 (approved):

PR #5502, opened 2025-03-05
Allow setting non-standard AWS Glue Registry endpoints for testing

seschis:
### Description
Add support for the kafka.schema.registry_url setting for the aws_glue schema registry.  Previously this setting was ignored and the code would only connect to the standard aws glue registry endpoint.  This change allows for running a local aws glue registry using something like localstack or motoserver so that local integration testing can occur.  We noticed this problem when setting up AWS OSIS and being unable to locally perform integration testing in our application.
 
### Issues Resolved
Resolves #[5377]
 
### Check List


data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -227,6 +227,7 @@ <T> void consumeRecords() throws Exception {
             topicMetrics.getNumberOfPollAuthErrors().increment();
             Thread.sleep(10000);
         } catch (RecordDeserializationException e) {
+            LOG.error("Deserialization error while doing poll()", e);
```
seschis, line 230: I originally had a problem where I didn't set the AWS_SECRET_ACCESS_KEY in the data-prepper docker image and it was failing to connect to the instance of [motoserver](https://github.com/getmoto/moto) I was running.  The helpful error message in the stack trace wasn't being displayed as it was about the 3rd "caused-by" down in the trace.  I had to add this error message to see it.

I don't think this is ideal so I wanted some guidance on how you want me to handle this error so that it provides a meaningful message to a user for correcting things.

seschis:
I see that the pipeline currently has some accepted failing tests based on previous merges.  I briefly looked into the failing tests and they didn't make a lot of sense based on the changes I made.  Before I go too deep troubleshooting the Kafka plugin integration test failures, could someone let me know if these are expected and accepted as failing as well.

dlvenable (approved):
This looks good to me. @kkondaka , can you also take a look?

kkondaka (approved):

kkondaka:
@seschis this is breaking existing behavior.  "Previously this setting was ignored and the code would only connect to the standard aws glue registry endpoint." -- this is incorrect. We do use this setting. I need to fix this immediately. I will add another option under `schema:` like `override_endpoint:` which when true this new code will be invoked

PR #5541, opened 2025-03-23
feat: Add configurable stream read constraints for JSON input codec

### LINKED ISSUE #5466: Support configurable stream read constraints max length in the JSON input codec
The default value for stream read constraints is 20,000,000 characters. This means that if any single Event being read by the json codec exceeds this size, then an error is thrown due to it not being parseable

```
com.fasterxml.core.exc.StreamConstraintsException "String value length (20054016 exceeds the maximum allowed (20000000 from StreamReadConstraints.getMaxStringLength()"
```

A configurable option in the JSON codec 

```
source:
  s3:
    codec:
      json:
        max_event_length: 30000000
```

And this could then be wired into the JSON decoder here (https://github.com/opensearch-project/data-prepper/blob/837d1a9a4bee8d1cb05e1f539bb41fa37c6fc070/data-prepper-api/src/main/java/org/opensearch/dataprepper/model/codec/JsonDecoder.java#L33)

```
public JsonDecoder(String keyName, Collection<String> includeKeys, Collection<String> includeKeysMetadata, int maxEventLength) {
        this.keyName = keyName;
        this.includeKeys = includeKeys;
        this.includeKeysMetadata = includeKeysMetadata;
        jsonFactory.setStreamReadConstraints(StreamReadConstraints.builder()
                        .maxStringLength(maxEventLength)
                .build());
    }
```

**Additional context**

saketh-pallempati:
- Introduced a configurable option for setting the maximum event length in the JSON input codec.
- Updated `JsonDecoder` to accept `max_event_length` parameter and apply it to `StreamReadConstraints`.
- Added validation to ensure the maximum event length is within acceptable limits. 

### Description
This change introduces a new configurable option that allows users to set the maximum event length for the JSON input codec. By setting this parameter, users can control the maximum size of JSON strings that the codec will process, ensuring that excessively large events are not processed and potentially causing issues. The `JsonDecoder` has been updated to accept the `max_event_length` parameter and apply it to the `StreamReadConstraints`. Additionally, validation has been added to ensure that the maximum event length is within acceptable limits.

### Issues Resolved
fixes #5466 

### Check List

The documentation issue link : https://github.com/opensearch-project/documentation-website/issues/9370

For more information on following Developer Certificate of Origin and signing off your commits, please check here.

### DIFF
--- a/data-prepper-api/src/main/java/org/opensearch/dataprepper/model/codec/JsonDecoder.java
+++ b/data-prepper-api/src/main/java/org/opensearch/dataprepper/model/codec/JsonDecoder.java
@@ -8,6 +8,7 @@ package org.opensearch.dataprepper.model.codec;
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.StreamReadConstraints;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import org.opensearch.dataprepper.model.event.Event;
@@ -30,10 +31,13 @@ public class JsonDecoder implements ByteDecoder {
     private Collection<String> includeKeys;
     private Collection<String> includeKeysMetadata;
 
-    public JsonDecoder(String keyName, Collection<String> includeKeys, Collection<String> includeKeysMetadata) {
+    public JsonDecoder(String keyName, Collection<String> includeKeys, Collection<String> includeKeysMetadata, int maxEventLength) {
         this.keyName = keyName;
         this.includeKeys = includeKeys;
         this.includeKeysMetadata = includeKeysMetadata;
+        jsonFactory.setStreamReadConstraints(StreamReadConstraints.builder()
+                .maxStringLength(maxEventLength)
+                .build());
     }
 
     public JsonDecoder() {
--- a/data-prepper-plugins/parse-json-processor/src/main/java/org/opensearch/dataprepper/plugins/codec/json/JsonInputCodecConfig.java
+++ b/data-prepper-plugins/parse-json-processor/src/main/java/org/opensearch/dataprepper/plugins/codec/json/JsonInputCodecConfig.java
@@ -11,11 +11,13 @@
 package org.opensearch.dataprepper.plugins.codec.json;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import jakarta.validation.constraints.Min;
 import jakarta.validation.constraints.Size;
 
 import java.util.List;
 
 public class JsonInputCodecConfig {
+    static final Integer DEFAULT_MAX_EVENT_LENGTH = 20000000;
 
     @JsonProperty("key_name")
     @Size(min = 1, max = 2048)
@@ -37,5 +39,13 @@ public class JsonInputCodecConfig {
 
     public List<String> getIncludeKeysMetadata() {
         return includeKeysMetadata;
+    }
+
+    @JsonProperty("max_event_length")
+    @Min(1)
+    private Integer maxEventLength = DEFAULT_MAX_EVENT_LENGTH;
+
+    public Integer getMaxEventLength(){
+        return maxEventLength;
     }
 }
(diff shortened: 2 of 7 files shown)


data-prepper-plugins/parse-json-processor/src/main/java/org/opensearch/dataprepper/plugins/codec/json/JsonInputCodecConfig.java
```diff
@@ -11,11 +11,13 @@
 package org.opensearch.dataprepper.plugins.codec.json;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import jakarta.validation.constraints.Min;
 import jakarta.validation.constraints.Size;
 
 import java.util.List;
 
 public class JsonInputCodecConfig {
+    static final Integer DEFAULT_MAX_EVENT_LENGTH = 20000000;
```
graytaylor0, line 20: Do you think we should just have this default to null to use the current default, and then just not pass it when it's not set? 

Might be a little safer since it will always use the library default

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/codec/JsonDecoder.java
```diff
@@ -30,10 +31,13 @@ public class JsonDecoder implements ByteDecoder {
...
     private Collection<String> includeKeysMetadata;
 
-    public JsonDecoder(String keyName, Collection<String> includeKeys, Collection<String> includeKeysMetadata) {
+    public JsonDecoder(String keyName, Collection<String> includeKeys, Collection<String> includeKeysMetadata, int maxEventLength) {
         this.keyName = keyName;
         this.includeKeys = includeKeys;
         this.includeKeysMetadata = includeKeysMetadata;
+        jsonFactory.setStreamReadConstraints(StreamReadConstraints.builder()
+                .maxStringLength(maxEventLength)
+                .build());
```
graytaylor0, line 40: With the the approach of not setting this for `null`, you could change maxEventLength to Integer and just check 

```
if (maxEventLength != null) {
   jsonFactory.setStreamReadConstraints(StreamReadConstraints.builder()
                .maxStringLength(maxEventLength)
                .build());
}
```

graytaylor0 (dismissed):

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/codec/JsonDecoder.java
saketh-pallempati: Yeah that makes sense will make the change to Jackson Builder only when the variable is set. Will get the new commit ASAP.

saketh-pallempati:
Hi @graytaylor0 ,
The suggestions were implemented in the new commit please do review.

graytaylor0 (approved):

data-prepper-plugins/parse-json-processor/src/main/java/org/opensearch/dataprepper/plugins/codec/json/JsonInputCodecConfig.java
```diff
@@ -38,4 +39,12 @@ public List<String> getIncludeKeys() {
     public List<String> getIncludeKeysMetadata() {
         return includeKeysMetadata;
     }
+
+    @JsonProperty("max_event_length")
+    @Min(1)
```
san81, line 44: is it a good idea to set the Min to Jackson's default max? Is there a case we want this to be below Jackson's default?

saketh-pallempati: Hi @san81 
Do you think the min value should be 20,000,000 which is default for Jackson?
I am actually not sure if there is case where less than jackson's default would make sense.
So any suggestions from your side should do the trick.

chenqi0805 (approved):
Thanks for making the change!

data-prepper-plugins/parse-json-processor/src/main/java/org/opensearch/dataprepper/plugins/codec/json/JsonInputCodecConfig.java
san81: Yes, I think the min value should be 20,000,000, which is the default for Jackson

saketh-pallempati:
> Thanks for making the change!

@chenqi0805 Thanks for approval.

data-prepper-plugins/parse-json-processor/src/main/java/org/opensearch/dataprepper/plugins/codec/json/JsonInputCodecConfig.java
saketh-pallempati: Sure then we will let Jackson's default be the Min value.

dlvenable (approved):
Thank you @saketh-pallempati for this improvement!

saketh-pallempati:
Thanks to all ✌️

PR #5585, opened 2025-04-04
Add test for date patterned index name

### LINKED ISSUE #984: e2e log test failing when index name has DateTime pattern suffix in OpenSearch Sink Config
Opensearch sink configuration now includes creating `index` name which can be a plain string + DateTime pattern as a suffix.
For example, `log-index-%{yyyy.MM.dd}` OR `test-index-%{yyyy.MM.dd.HH}`

Observed a failure in the e2e log test when setting the index with a date-time pattern suffix i.e. `log-index-%{yyyy.MM.dd}` in the `basic-grok-e2e-pipeline.yaml`

```
com.amazon.dataprepper.integration.log.EndToEndBasicLogTest > testPipelineEndToEnd FAILED
    org.opensearch.OpenSearchStatusException at EndToEndBasicLogTest.java:71
```

TomasLongo:
### Description
Creates an additional test to verify correct behaviour with date patterned indices
 
### Issues Resolved
Resolves #984


TomasLongo:
@dlvenable The failing test ran into a timeout. Since the trace isn't affected by this change, a restart of the failed job might be trick.

PR #5589, opened 2025-04-08
Add OTel Metrics String Attributes to Index Template

KarstenSchnitter:
### Description
Introduces keyword mappings for string fields. This was the default for all fields before the latest change in #5539.
 
### Issues Resolved
Fix for regression introduced by #5539.

### Check List

### DIFF
--- a/data-prepper-plugins/opensearch/src/main/resources/index-template/metrics-otel-v1-index-template.json
+++ b/data-prepper-plugins/opensearch/src/main/resources/index-template/metrics-otel-v1-index-template.json
@@ -26,6 +26,16 @@
           }
         },
         {
+          "string_resource_attributes": {
+            "mapping": {
+              "type": "keyword",
+              "ignore_above": 256
+            },
+            "path_match": "resource.attributes.*",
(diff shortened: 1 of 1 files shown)


kkondaka (commented):
You also need to make this change to the file `./data-prepper-plugins/opensearch/src/main/resources/metrics-otel-v1-index-standard-template.json`. Not sure why we have two files but I found that the template files in `./data-prepper-plugins/opensearch/src/main/resources/` directory are used anot the one in the sub directory `index-template`

KarstenSchnitter:
@kkondaka: This is a fix for your PR #5539, which only changed the existing mapping in the `index_template` directory. Therefore, this change is only for that one file. `./data-prepper-plugins/opensearch/src/main/resources/metrics-otel-v1-index-standard-template.json` uses a much simpler mapping, where this change is not required.

Maybe you change the file in #5539 by accident?

kkondaka:
@KarstenSchnitter You are right. It looks like I merged that change by accident.

kkondaka (approved):

PR #5687, opened 2025-05-08
Kds cross account stream

sb2k16:
### Description
This PR is to add support for accessing cross account Kinesis Data streams. This will allow data prepper to securely consume data from streams belonging to different AWS accounts while maintaining proper access controls and performance optimization.

Add Stream ARN Support

* Add new optional field: `stream_arn`  to allow explicit specification of cross-account stream ARNs
* Format: `arn:aws:kinesis:region:account-id:stream/<stream-name>`

Add Consumer ARN support

* Add new optional field: `consumer_arn`  to allow explicit specification of cross-account stream ARNs
* Format: `arn:aws:kinesis:region:account-id:stream/stream-name/consumer/<consumer_arn>:<consumer epoch time>`

### Issues Resolved
Resolves #1082 
 
### Check List


data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/KinesisClientApiHandler.java
```diff
@@ -30,44 +33,84 @@ public class KinesisClientApiHandler {
...
         this.kinesisClient = kinesisClient;
         this.backoff = backoff;
-        this.failedAttemptCount = 0;
         if (maxRetryCount <= 0) {
             throw new IllegalArgumentException("Maximum Retry count should be strictly greater than zero.");
         }
         this.maxRetryCount = maxRetryCount;
     }
 
+    public StreamIdentifier getStreamIdentifierFromStreamArn(final String streamArnString) {
```
dlvenable, line 47: I think you can make this package protected. Remove the `public` modifier.

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/KinesisClientApiHandler.java
```diff
@@ -30,44 +33,84 @@ public class KinesisClientApiHandler {
...
-        failedAttemptCount = 0;
         DescribeStreamSummaryRequest describeStreamSummaryRequest = DescribeStreamSummaryRequest.builder()
                 .streamName(streamName).build();
-        while (failedAttemptCount < maxRetryCount) {
+        DescribeStreamSummaryResponse response = getStreamDescriptionSummary(describeStreamSummaryRequest);
+        String streamIdentifierString = getStreamIdentifierString(response.streamDescriptionSummary());
+        return StreamIdentifier.multiStreamInstance(streamIdentifierString);
+    }
+
+    public String getConsumerArnForStream(final String streamArn, final String consumerName) {
```
dlvenable, line 64: It looks like this is only used internally, so you can make it private.

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/KinesisClientApiHandler.java
```diff
@@ -77,13 +120,18 @@ private void applyBackoff() {
         try {
             Thread.sleep(delayMillis);
         } catch (final InterruptedException e){
+            Thread.currentThread().interrupt();
```
dlvenable, line 123: You reset the interrupt here. But, does anything actually monitor the interrupt status? Maybe you should end the backoff/retry loop.

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/processor/KinesisRecordProcessor.java
```diff
@@ -112,7 +114,18 @@ public KinesisRecordProcessor(final BufferAccumulator<Record<Event>> bufferAccum
...
+                Arn streamIdentifierArn = streamIdentifier.streamArnOptional().get();
+                Arn streamConfigArn = Arn.fromString(streamConfig.getArn());
+                return streamIdentifierArn.equals(streamConfigArn);
+            }
+            if (Objects.nonNull(streamConfig.getArn())) {
+                String streamName = Arn.fromString(streamConfig.getArn()).resource().resource();
+                return streamName.equals(streamIdentifier.streamName());
+            }
+            return streamConfig.getName().equals(streamIdentifier.streamName());
+        }).findAny();
```
dlvenable, line 128: This `findAny` can return one if there are multiple matching. Should we have some check to ensure there is only one?

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/configuration/KinesisStreamConfig.java
```diff
@@ -26,10 +27,17 @@ public class KinesisStreamConfig {
     private static final boolean DEFAULT_ENABLE_CHECKPOINT = false;
 
     @JsonProperty("stream_name")
-    @NotNull
     @Valid
     private String name;
 
+    @JsonProperty("stream_arn")
+    @Valid
+    private String arn;
```
dlvenable, line 35: Rename to `getStreamArn` to make it clearer what this is.

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/processor/KinesisRecordProcessor.java
```diff
@@ -112,7 +114,18 @@ public KinesisRecordProcessor(final BufferAccumulator<Record<Event>> bufferAccum
     }
 
     private KinesisStreamConfig getStreamConfig(final KinesisSourceConfig kinesisSourceConfig) {
-        final Optional<KinesisStreamConfig> kinesisStreamConfig = kinesisSourceConfig.getStreams().stream().filter(streamConfig -> streamConfig.getName().equals(streamIdentifier.streamName())).findAny();
+        final Optional<KinesisStreamConfig> kinesisStreamConfig = kinesisSourceConfig.getStreams().stream().filter(streamConfig -> {
```
dlvenable, line 117: We have a bit of split logic with how we ready the stream config and get the stream identifiers. I see other conditions based on the different options above in `KinesisClientApiHandler` and `KinesisMultiStreamTracker`.

There should be a good way to have a single class that can handle all the logic for determining the `KinesisStreamConfig`/`StreamConfig` for any stream or identifier. Maybe a class that maps all these on initialization?

Then you won't have to have all this split logic.

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/KinesisClientApiHandler.java
dlvenable: Based on my other comments, I think this code can be refactored to have a single place for handling the stream identification.

dlvenable (changes requested):
Thanks @sb2k16 for this new feature! Some refactoring can help isolate the logic.

dlvenable (dismissed):
Thanks @sb2k16 !

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/apihandler/KinesisClientApiRetryHandler.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+        final Duration delayDuration = Duration.ofMillis(delayMillis);
+        log.info("Pausing request for {}.{} seconds due to an error in processing.",
+                delayDuration.getSeconds(), delayDuration.toMillisPart());
+        try {
+            Thread.sleep(delayMillis);
+        } catch (final InterruptedException e) {
+            Thread.currentThread().interrupt();
+            log.error("Thread interrupted while waiting for retry", e);
+            throw new KinesisRetriesExhaustedException("Thread interrupted while waiting for retry");
```
kkondaka, line 74: Is it a good idea to throw an error for this? It is a very rare event. In the worst case, it should not count towards an "attempt"

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/apihandler/KinesisClientApiHandler.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    private void handleConsumerException(CompletionException ex, String streamArn, int attempt) {
+        Throwable cause = ex.getCause();
+        if (cause instanceof KinesisException) {
+            log.error("Kinesis API error while describing stream consumer for stream {}: {}. Attempt {}.",
+                    streamArn, ex.getMessage(), attempt + 1);
+        } else if (cause instanceof com.amazonaws.SdkClientException) {
+            log.error("AWS SDK client error while describing stream consumer for stream {}: {}. Attempt {}.",
+                    streamArn, ex.getMessage(), attempt + 1);
```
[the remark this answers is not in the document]
sb2k16: The `ex.message() will show the exception. However, the line 145 is to have a different message in the logs.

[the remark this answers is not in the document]
dlvenable: The only differene here is the `AWS SDK` versus `Kinesis API`. I think @kkondaka is right that we don't need this.

This is the default `Throwable::toString`. I think this give us what we want.

```
public String toString() {
        String s = getClass().getName();
        String message = getLocalizedMessage();
        return (message != null) ? (s + ": " + message) : s;
    }
```

Something like this will work:

```
            log.error("AWS error while describing stream consumer for stream {}: {}. Attempt {}.",
                    streamArn, ex.toString(), attempt + 1);
```

kkondaka (approved):

dlvenable (approved):

PR #5700, opened 2025-05-19
fix(opensearch): default to OpenSearch if distribution detection fails

### LINKED ISSUE #5673: Assume distribution_version is set to opensearch for OpenSearch source
The opensearch source attempts to auto detect by calling the `GET /` API if the source is an opensearch or elasticsearch cluster. Sometimes, clusters have spoofed responses that impersonate something they are not, and this can lead to failures without the distribution version being set. 

To handle this more gracefully, if distribution_version is not set, and the attempts to call the `GET /` API fail for both the opensearch and elasticsearch clients, the opensearch source should act as if the `distribution_version` is set to OpenSearch, and continue

**Additional context**

saketh-pallempati:
### Description
**Accurate Client Selection**: The logic for choosing between an _OpenSearchAccessor_ and an _ElasticsearchAccessor_ was refined to correctly rely on the determined cluster distribution, especially after attempting to get version information.

**Reliable Default to OpenSearch**: If the cluster distribution cannot be auto-detected (i.e., API calls to both OpenSearch and Elasticsearch clients fail) and the distribution_version is not configured, the strategy now consistently defaults to OpenSearch behavior and uses an _OpenSearchAccessor_.

**Enhanced Edge Case Handling**: The code now better manages situations where an Elasticsearch client needs to be created dynamically, for instance, if an initial OpenSearch client call succeeds but indicates an Elasticsearch distribution.
 
### Issues Resolved
Resolves #5673
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/SearchAccessorStrategy.java
+++ b/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/SearchAccessorStrategy.java
@@ -81,8 +81,8 @@ public class SearchAccessorStrategy {
 
         InfoResponse infoResponse = null;
 
-        PluginComponentRefresher<ElasticsearchClient, OpenSearchSourceConfiguration> elasticsearchClientRefresher =
-                null;
+        PluginComponentRefresher<ElasticsearchClient, OpenSearchSourceConfiguration> elasticsearchClientRefresher = null;
+        
         try {
             infoResponse = clientRefresher.get().info();
             pluginConfigObservable.addPluginConfigObserver(newConfig -> clientRefresher.update(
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/SearchAccessorStrategy.java
```diff
@@ -189,14 +210,58 @@ private boolean versionSupportsPointInTime(final String distribution, final Stri
...
+                                Objects.toString(esBuildFlavor, "unknown"), esVersionNumber);
+                        if (ELASTICSEARCH_OSS_BUILD_FLAVOR.equalsIgnoreCase(esBuildFlavor)) {
+                            effectiveDistribution = ELASTICSEARCH_DISTRIBUTION + "-" + ELASTICSEARCH_OSS_BUILD_FLAVOR;
+                        }
+                        return Pair.of(effectiveDistribution, esVersionNumber);
+                    }
+                    LOG.warn("Version number is null in Elasticsearch API response. Proceeding to default.");
+                }
+            } catch (final Exception e) {
+                LOG.warn("Failed to get cluster info using Elasticsearch client. Will proceed with default behavior.", e);
```
graytaylor0, line 244: This log could maybe indicate that the default behavior is to assume that it is OpenSearch, although you do have that log later on

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/client/SearchAccessorStrategy.java
```diff
@@ -189,14 +210,58 @@ private boolean versionSupportsPointInTime(final String distribution, final Stri
...
+        } else {
+            LOG.info("Elasticsearch client refresher is null, skipping Elasticsearch .info() call.");
+        }
+
+        if (openSearchSourceConfiguration.getDistributionVersion() == null) { 
+            LOG.warn("Failed to auto-detect cluster distribution from API response for both OpenSearch and (if attempted) Elasticsearch clients. " +
+                     "Defaulting to OpenSearch distribution. Consider setting 'distribution_version' in the configuration if this is not an OpenSearch cluster.");
+            return Pair.of(OPENSEARCH_DISTRIBUTION, OPENSEARCH_POINT_IN_TIME_SUPPORT_VERSION_CUTOFF);
+        } else {
+            final String configuredDistributionName = openSearchSourceConfiguration.getDistributionVersion().name();
```
graytaylor0, line 255: I'm not sure we want this check to happen. If users configure distribution version it's probably better to trust it and create that type of accessor. I think there could be cases where the distribution_version is correct but the info response can still return unexpected results. So I would just keep the behavior the same (if distribution_version is set, just force that type of accessor

saketh-pallempati: Sure I was just trying to add some extra validation. Will simplify the case

graytaylor0:
I also tried to run the DCO check and it failed

saketh-pallempati:
> I also tried to run the DCO check and it failed

Yeah some problem on my end i will figure it.

graytaylor0:
It's still failing DCO

saketh-pallempati:
> It's still failing DCO

Sorry that I made a change which should be pushed by my org. Please give it some time.

saketh-pallempati:
Hi @graytaylor0 
Please check the commit for DCO and the code changes

graytaylor0 (approved):
Thanks for cleaning this up!

saketh-pallempati:
Thanks for the approval @graytaylor0

PR #5725, opened 2025-05-28
Set server id for MySQL binlog client

oeyh:
### Description
Set a custom server id for MySQL binlog client before making connection to the MySQL database. 

Before this CR, server id is always 65535 by default and can result in server id conflict when two Data Prepper pipelines try to connect to the same MySQL database's binary log stream.
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/utils/ServerIdGenerator.java
+++ b/data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/utils/ServerIdGenerator.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ *  The OpenSearch Contributors require contributions made to
+ *  this file be licensed under the Apache-2.0 license or a
+ *  compatible open source license.
+ */
+
+package org.opensearch.dataprepper.plugins.source.rds.utils;
+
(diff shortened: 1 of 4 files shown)


data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/utils/ServerIdGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+package org.opensearch.dataprepper.plugins.source.rds.utils;
+
+import java.net.InetAddress;
+import java.util.Random;
+
+public class ServerIdGenerator {
+    static final int MIN_SERVER_ID = 100_000;
+    static final int MAX_SERVER_ID = 999_999;
+
+    public static int generateServerId() {
```
chenqi0805, line 19: Is this value guaranteed to be unique?

chenqi0805 (approved):

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/utils/ServerIdGenerator.java
oeyh: It's not 100% that it will be unique but the possibility of conflict is very low. 

If there's still server id conflict between two Data Prepper instances, the retry logic provided by StreamScheduler will re-generate a server id and reconnect.

PR #5734, opened 2025-05-31
add internal batching of the records for Sagemaker job creator

Zhangxunmt:
### Description
This PR enhances batch job creation by implementing thread-safe record collection across multiple processor worker threads. 

Currently when processing multiple S3 files (e.g., an S3 source with 7 files), records are randomly distributed across worker threads. For example, worker-1 might process files 2,4,6 while worker-2 processes 1,3,5,7, resulting in two separate, potentially undersized batch jobs. Ideally we should only create 1 batch job to process all the files with best efficiency. 

Rather than having each worker thread independently create separate ML batch jobs, all worker threads now contribute records to a shared, thread-safe buffer in the ml processor. The system creates a single consolidated ML batch job when either of these conditions is met:
1. The accumulated records in the buffer reach the maximum batch size threshold (default: 100 records)
2. The buffer remains unchanged (no new records added) for a configurable idle time window (default: 1 minute)

This approach optimizes resource utilization by consolidating what would have been multiple smaller batch jobs into fewer, appropriately-sized batch operations.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/common/SageMakerBatchJobCreator.java
+++ b/data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/common/SageMakerBatchJobCreator.java
@@ -17,15 +17,20 @@ import org.opensearch.dataprepper.model.event.Event;
 import org.opensearch.dataprepper.model.event.EventKey;
 import org.opensearch.dataprepper.model.record.Record;
 import org.opensearch.dataprepper.plugins.ml_inference.processor.MLProcessorConfig;
-import org.opensearch.dataprepper.plugins.ml_inference.processor.exception.MLBatchJobException;
 import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
 
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 
 import static org.opensearch.dataprepper.common.utils.RetryUtil.retryWithBackoff;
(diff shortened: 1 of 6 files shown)


data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/common/SageMakerBatchJobCreator.java
```diff
@@ -50,55 +63,149 @@ public SageMakerBatchJobCreator(final MLProcessorConfig mlProcessorConfig, final
         this.awsCredentialsSupplier = awsCredentialsSupplier;
         this.s3Client = createS3Client(mlProcessorConfig, awsCredentialsSupplier);
         this.dateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMddHHmmss");
+        this.MAX_BATCH_SIZE = mlProcessorConfig.getMaxBatchSize();
+
+        // Initialize batch processing scheduler
+        scheduler.scheduleAtFixedRate(
```
kkondaka: Actually, you don't need this. processor's `doExecute()` is always invoked even when there is no traffic. You can use that.

Zhangxunmt: Great point! I just noticed each worker runs every 3 seconds. just made the change to use the worker threads in the refresh.

data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/common/SageMakerBatchJobCreator.java
```diff
@@ -39,6 +41,13 @@ public class SageMakerBatchJobCreator extends AbstractBatchJobCreator {
     private final S3Client s3Client;
     private final DateTimeFormatter dateTimeFormatter;
 
+    // Added batch processing fields
+    private final ConcurrentLinkedQueue<Record<Event>> batch_records = new ConcurrentLinkedQueue<>();
+    private final ConcurrentLinkedQueue<Record<Event>> processedBatchRecords = new ConcurrentLinkedQueue<>();
+    private final int MAX_BATCH_SIZE;
```
kkondaka, line 47: Why capital letters for this variable? Maybe, you can fix this in the next PR?

kkondaka (approved):

data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/common/SageMakerBatchJobCreator.java
Zhangxunmt: sure will definitely do in the next PR. It might be from the first definition as a constant and forgot to change it after I make it readable from the config.

srikanthjg (approved):

PR #5764, opened 2025-06-06
fix: lastPollTime updating for O365

savit-aluri:
### Fix Office365 Source Duplicate Logs Issue

  ###  Problem
  The Office365 source plugin was experiencing duplicate log entries due to improper handling of `lastPollTime` and time window processing. This was causing:
  1. Same logs being ingested multiple times
  2. Inefficient polling due to not properly advancing through time windows

  ### Changes
  - Modified `Office365Service/Office365Iterator` to properly track time window completion
    - Improved logging around time window processing
    - Kept `lastModifiedAt` as the time that the log was imported into open search to be able to track later.

  ## Testing
  - Verified no duplicate logs are ingested when processing multiple time windows
  - Confirmed lastPollTime advances correctly through time windows
  - Tested window completion state handling

  ## Additional Notes
  - This change leverages `PaginationCrawler`'s built-in lastPollTime tracking instead of custom window tracking
  - Maintains accurate event timestamps while ensuring proper time window progression
  - Improves logging to help track window processing state
 

### Check List
- [N] New functionality includes testing.
- [N] New functionality has a documentation issue. Please link to it in this PR.
  - [N] New functionality has javadoc added

### DIFF
--- a/data-prepper-plugins/saas-source-plugins/source-crawler/src/main/java/org/opensearch/dataprepper/plugins/source/source_crawler/base/PaginationCrawler.java
+++ b/data-prepper-plugins/saas-source-plugins/source-crawler/src/main/java/org/opensearch/dataprepper/plugins/source/source_crawler/base/PaginationCrawler.java
@@ -95,9 +95,11 @@ public class PaginationCrawler implements Crawler<PaginationCrawlerWorkerProgres
 
     private void updateLeaderProgressState(LeaderPartition leaderPartition, Instant updatedPollTime, EnhancedSourceCoordinator coordinator) {
         LeaderProgressState leaderProgressState = leaderPartition.getProgressState().get();
+        Instant oldPollTime = leaderProgressState.getLastPollTime();
         leaderProgressState.setLastPollTime(updatedPollTime);
         leaderPartition.setLeaderProgressState(leaderProgressState);
         coordinator.saveProgressStateForPartition(leaderPartition, DEFAULT_EXTEND_LEASE_MINUTES);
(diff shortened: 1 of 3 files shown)


san81 (approved):
I think this will fix it but from the overall functionality point of view, make sure that frequent polling with the updated last modified time is Ok and given a new last modified time, your window construction is working as expected.

data-prepper-plugins/saas-source-plugins/source-crawler/src/main/java/org/opensearch/dataprepper/plugins/source/source_crawler/base/PaginationCrawler.java
```diff
@@ -95,9 +95,11 @@ public void executePartition(PaginationCrawlerWorkerProgressState state, Buffer<
 
     private void updateLeaderProgressState(LeaderPartition leaderPartition, Instant updatedPollTime, EnhancedSourceCoordinator coordinator) {
         LeaderProgressState leaderProgressState = leaderPartition.getProgressState().get();
+        Instant oldPollTime = leaderProgressState.getLastPollTime();
```
graytaylor0, line 98: When we update this state, is it 100% guaranteed we will never need to go back from before this timestamp to reprocess?

savit-aluri: Yup, again no change to this particular logic here, this was just to add a debug log. In response to your question yes, we will not go back to a previous time for this.

PR #5847, opened 2025-06-30
Chunk queries for duplicate documents, add metrics, track duplicate documents found in queries

graytaylor0:
### Description
Makes the following improvements to querying for existing documents in the OpenSearch sink

* chunk search requests into batches of 1,000. This prevents too many term values from being sent in the same search request
* Improve error log for failed search requests to provide full error rather than just the reason
* When 2 documents with the same term value are found in a query, log and add metric for document id
* Add metric for the same term value being added to the query manager with `addBulkOperation`
* Add metric for tracking the time taken to search for documents
* Fixes issue where metric tracking documents being queried wasn't decremented when events were dropped and released, which could cause process workers to block and deadlock forever.

 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/ExistingDocumentQueryManager.java
+++ b/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/ExistingDocumentQueryManager.java
@@ -48,7 +48,7 @@ public class ExistingDocumentQueryManager implements Runnable {
 
     static final String QUERY_TIME = "queryDuplicatesTime";
 
-    static final String POTENTIAL_DUPLICATES_DELETED = "potentialDuplicatesDeleted";
+    static final String POTENTIAL_DUPLICATES = "potentialDuplicates";
 
     private final Counter eventsDroppedAndReleasedCounter;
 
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/ExistingDocumentQueryManager.java
```diff
@@ -168,19 +188,25 @@ private MsearchRequest buildMultiSearchRequest() {
...
-                                .size(values.size())
-                                .source(source -> source.filter(f -> f.includes(queryTerm)))
-                                .query(Query.of(q -> q
-                                        .terms(TermsQuery.of(t -> t
-                                                .field(queryTerm)
-                                                .terms(TermsQueryField.of(tf -> tf.value(values)))
-                                        ))
-                                ))
-                        ));
+                final int batchSize = 1000;
```
kkondaka, line 191: Do we want to make this configurable?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/ExistingDocumentQueryManager.java
```diff
@@ -246,7 +272,10 @@ private void dropAndReleaseFoundEvents(final MsearchResponse<ObjectNode> msearch
                         final Map<String, QueryManagerBulkOperation> bulkOperationsForIndex = bulkOperationsWaitingForQuery.get(indexForHit);
                         final QueryManagerBulkOperation bulkOperationToRelease = bulkOperationsForIndex.get(queryTermValue);
                         if (bulkOperationToRelease == null) {
-                            LOG.error("bulk operation for term value {} is null", queryTermValue);
+                            // Delete duplicate document
+                            LOG.warn("Bulk operation for term value {} with id {} is null, potentially a duplicate document, deleting", queryTermValue, hit.id());
+                            potentialDuplicatesDeleted.increment();
+                            deleteDuplicateDocument(hit.index(), hit.id());
```
kkondaka: Are we supposed to delete existing documents?

graytaylor0: This happens when we find 2 documents with the same term value, so it does clean up a duplicate. It technically shouldn't happen but If you'd prefer I can just keep the metric and log that there is a duplicate rather than deleting it.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/ExistingDocumentQueryManager.java
graytaylor0: We could make it configurable I suppose.

kkondaka (approved):

PR #5929, opened 2025-07-30
fixes OTLP sink metric names to use dot notation instead of underscores

geekgirl10:
### Description
This PR updates the OTLP sink metrics to use dot notation (.) instead of underscore (_) in metric names.
This aligns with the convention used across other Data Prepper plugins and improves consistency and compatibility with metric backends that treat dot-separated tokens as hierarchy.
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/metrics/OtlpSinkMetrics.java
+++ b/data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/metrics/OtlpSinkMetrics.java
@@ -6,7 +6,6 @@
 package org.opensearch.dataprepper.plugins.sink.otlp.metrics;
 
 import io.micrometer.core.instrument.DistributionSummary;
-import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.instrument.Timer;
 import org.opensearch.dataprepper.metrics.PluginMetrics;
 import org.opensearch.dataprepper.model.configuration.PluginSetting;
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/metrics/OtlpSinkMetrics.java
```diff
@@ -73,7 +73,7 @@ private static Timer buildLatencyTimer(@Nonnull final String pipelineName, @Nonn
      * @return The distribution summary
      */
     private static DistributionSummary buildDistributionSummary(@Nonnull final String pipelineName, @Nonnull final String pluginName, @Nonnull final String metricName) {
```
graytaylor0: Why aren't we using `pluginMetrics` for these?

geekgirl10: This method was originally added by a previous engineer, but I couldn’t find any documentation explaining the reasoning behind not using pluginMetrics here. The main difference is that this approach includes additional configurations like percentiles and histogram settings. That’s likely why it was implemented separately instead of using pluginMetrics

Given that, we plan to update the code to use pluginMetrics for consistency and maintainability.

dlvenable (approved):
Thank you @geekgirl10 for this improvement!

graytaylor0 (approved):

PR #5969, opened 2025-08-05
fix TranslateProcessorConfig to load target.type property correct (#5…

wjyao0316:
…961)

### Description

**What?**

This commit update TranslateProcessorConfig to load the target.type property from pipeline config with Jackson's support. The cause is that converter parameter is not updated while Jackson parsing the target.type properly. The fix removes the unncessary private field converter and instead return targetType.

It also make secondar contstructor protected for unit test only.

**Why?**

Make target.type property in translate processor to take effect.
 
### Issues Resolved
Resolves #[5961]
 
### Check List
- [N/A ] New functionality includes testing.
- [ N/A] New functionality has a documentation issue. Please link to it in this PR.
  - [ N/A] New functionality has javadoc added
- [ Y] Commits are signed with a real name per the DCO


alparish:
DCO check is failing, we need to add sign off for every commit like this:
Signed-off-by: Alekhya Parisha <aparisha@amazon.com>

graytaylor0 (approved):

PR #6023, opened 2025-08-29
Add caching support to AWS lambda processor

### LINKED ISSUE #5788: Add Lookup Processor
Need a processor similar to OTEL's lookup processor or memcached like processor which can lookup a key and cache the value

Enhance Lambda processor to  cache the results of a lambda execution for a short TTL. And lambda input should identify the key and send just the key as input to lambda processor, instead of the entire event.

Alternately, 
1. support lookup processor similar to OTEL 
2. add memcached processor which interacts with external memcache server
3. lookup of S3 with key as S3 object name and value as contents of the object

In all these cases the data is cached locally for a short duration

**Additional context**

kkondaka:
### Description
Add caching support to `aws_lambda` processor.

The cache config can be added to existing lambda config as follows:

```
aws_lambda:
   <existing options>
    cache:
        ttl: <ttl in seconds>
        max_size: <maximum_size_in_bytes>
```

In addition, caching requires a set of keys whose values are used as cache keys. A new `keys` is added to the config at the top level so that even lambda processors without caching can used this option to send only a partial list of keys to the lambda function. The result from the lambda function can be either "merged" with the original Event contents or replace the original event contents. New config for this is 

```
aws_lambda:
   <existing options>
   keys: ["key1", "key2"]
   response_mode: "replace" (default) or "merge"

```

`keys` field is mandatory when using caching.
 
### Issues Resolved
Resolves #5788 
 
### Check List

### DIFF
--- a/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
+++ b/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
@@ -15,6 +15,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.commons.lang3.time.StopWatch;
 import org.opensearch.dataprepper.model.codec.OutputCodec;
 import org.opensearch.dataprepper.model.event.Event;
+import org.opensearch.dataprepper.model.event.JacksonEvent;
 import org.opensearch.dataprepper.model.record.Record;
 import org.opensearch.dataprepper.model.sink.OutputCodecContext;
 import org.opensearch.dataprepper.plugins.codec.json.JsonOutputCodec;
@@ -38,14 +39,16 @@ public class InMemoryBuffer implements Buffer {
     private final long payloadResponseSize;
     private int eventCount;
     private long payloadRequestSize;
+    private List<String> keys;
 
 
     public InMemoryBuffer(String batchOptionKeyName) {
-        this(batchOptionKeyName, new OutputCodecContext());
+        this(batchOptionKeyName, new OutputCodecContext(), null);
     }
 
-    public InMemoryBuffer(String batchOptionKeyName, OutputCodecContext outputCodecContext) {
+    public InMemoryBuffer(String batchOptionKeyName, OutputCodecContext outputCodecContext, List<String> keys) {
         byteArrayOutputStream = new ByteArrayOutputStream(32 * 1024);
+        this.keys = keys;
         records = new ArrayList<>();
         bufferWatch = new StopWatch();
         bufferWatch.start();
@@ -64,6 +67,14 @@ public class InMemoryBuffer implements Buffer {
     public void addRecord(Record<Event> record) {
         records.add(record);
         Event event = record.getData();
+        if (keys != null && keys.size() > 0) {
+            Event newEvent = JacksonEvent.fromEvent(event);
+            newEvent.clear();
+            for (final String key: keys) {
+                newEvent.put(key, event.get(key, Object.class));
+            }
+            event = newEvent;
+        }
         try {
             if (eventCount == 0) {
                 requestCodec.start(this.byteArrayOutputStream, event, this.outputCodecContext);
--- a/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
+++ b/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
@@ -68,8 +68,9 @@ public class InMemoryBuffer implements Buffer {
         records.add(record);
         Event event = record.getData();
         if (keys != null && keys.size() > 0) {
-            Event newEvent = JacksonEvent.fromEvent(event);
-            newEvent.clear();
+            Event newEvent = JacksonEvent.builder()
+                             .withEventType(event.getMetadata().getEventType())
+                             .build();
             for (final String key: keys) {
                 newEvent.put(key, event.get(key, Object.class));
             }
--- a/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/config/LambdaCommonConfig.java
+++ b/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/config/LambdaCommonConfig.java
@@ -15,8 +15,9 @@ import java.time.Duration;
 import lombok.Getter;
 import org.opensearch.dataprepper.model.configuration.PluginModel;
 
-@Getter
+import java.util.List;
 
+@Getter
 public abstract class LambdaCommonConfig {
 
   public static final int DEFAULT_CONNECTION_RETRIES = 3;
@@ -51,6 +52,14 @@ public abstract class LambdaCommonConfig {
   @JsonProperty("response_codec")
   @Valid
   private PluginModel responseCodecConfig;
+
+  @JsonPropertyDescription("Keys to send to lambda")
+  @JsonProperty("keys")
+  private List<String> keys;
+
+  public boolean hasKeys() {
+    return keys != null && keys.size() > 0;
+  }
 
   public abstract InvocationType getInvocationType();
 
--- a/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKey.java
+++ b/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKey.java
@@ -12,9 +12,14 @@ import java.util.List;
 import java.util.ArrayList;
 
 public class LambdaCacheKey {
-    private List<Object> keys;
+    private final List<Object> keys;
 
     public LambdaCacheKey(final List<Object> keys) {
+        for (final Object key : keys) {
+            if (!((key instanceof String) || (key instanceof Long) || (key instanceof Integer))) {
+                throw new RuntimeException("Only String and Long/Integer values are supported in cache");
+            }
+        }
         this.keys = keys;
     }
 
@@ -23,7 +28,7 @@ public class LambdaCacheKey {
         for (final String keyName : keyNames) {
             Object value = event.get(keyName, Object.class);
             if (!((value instanceof String) || (value instanceof Long) || (value instanceof Integer))) {
-                throw new RuntimeException("Only String and Number values are supported in cache");
+                throw new RuntimeException("Only String and Long/Integer values are supported in cache");
             }
             this.keys.add(value);
         }
--- a/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKey.java
+++ b/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKey.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+
+package org.opensearch.dataprepper.plugins.lambda.processor;
+
+import org.opensearch.dataprepper.model.event.Event;
+
+import java.util.List;
+import java.util.ArrayList;
+
+public class LambdaCacheKey {
+    private List<Object> keys;
+
+    public LambdaCacheKey(final List<Object> keys) {
+        this.keys = keys;
+    }
+
+    public LambdaCacheKey(final Event event, final List<String> keyNames) {
+        this.keys = new ArrayList<>();
+        for (final String keyName : keyNames) {
+            Object value = event.get(keyName, Object.class);
+            if (!((value instanceof String) || (value instanceof Long) || (value instanceof Integer))) {
+                throw new RuntimeException("Only String and Number values are supported in cache");
+            }
+            this.keys.add(value);
+        }
+    }
+
+    public int length() {
+        return keys.stream()
+                .mapToInt(key -> {
+                    if (key instanceof String) {
+                        return ((String) key).length();
+                    } else { // instance of Long/Integer
+                        return 8;
+                    }
+                })
+                .sum();
+    }
+
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof LambdaCacheKey)) return false;
+        return keys.equals(((LambdaCacheKey)o).keys);
+    }
+
+    @Override
+    public int hashCode() {
+        return keys.hashCode();
+    }
+
+}
--- a/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
+++ b/data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
@@ -5,12 +5,15 @@
 
 package org.opensearch.dataprepper.plugins.lambda.processor;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.Timer;
 import org.opensearch.dataprepper.aws.api.AwsCredentialsSupplier;
 import org.opensearch.dataprepper.expression.ExpressionEvaluator;
+
 import static org.opensearch.dataprepper.logging.DataPrepperMarkers.NOISY;
+
 import org.opensearch.dataprepper.metrics.PluginMetrics;
 import org.opensearch.dataprepper.model.annotations.DataPrepperPlugin;
 import org.opensearch.dataprepper.model.annotations.DataPrepperPluginConstructor;
@@ -27,7 +30,9 @@ import org.opensearch.dataprepper.model.record.Record;
 import org.opensearch.dataprepper.model.sink.OutputCodecContext;
 import org.opensearch.dataprepper.plugins.codec.json.JsonOutputCodecConfig;
 import org.opensearch.dataprepper.plugins.lambda.common.LambdaCommonHandler;
+
 import static org.opensearch.dataprepper.plugins.lambda.common.LambdaCommonHandler.isSuccess;
+
 import org.opensearch.dataprepper.plugins.lambda.common.ResponseEventHandlingStrategy;
 import org.opensearch.dataprepper.plugins.lambda.common.accumlator.Buffer;
 import org.opensearch.dataprepper.plugins.lambda.common.client.LambdaClientFactory;
@@ -53,13 +58,31 @@ import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.Weigher;
+
+
 @DataPrepperPlugin(name = "aws_lambda", pluginType = Processor.class, pluginConfigurationType = LambdaProcessorConfig.class)
 public class LambdaProcessor extends AbstractProcessor<Record<Event>, Record<Event>> {
+
+    class ByteArrayValueWeigher implements Weigher<LambdaCacheKey, Event> {
+        @Override
+        public int weigh(LambdaCacheKey key, Event value) {
+            return value.toJsonString().length() + key.length();
+        }
+    }
 
     public static final String NUMBER_OF_RECORDS_FLUSHED_TO_LAMBDA_SUCCESS = "recordsSuccessfullySentToLambda";
     public static final String NUMBER_OF_RECORDS_FLUSHED_TO_LAMBDA_FAILED = "recordsFailedToSentLambda";
     public static final String NUMBER_OF_SUCCESSFUL_REQUESTS_TO_LAMBDA = "numberOfRequestsSucceeded";
     public static final String NUMBER_OF_FAILED_REQUESTS_TO_LAMBDA = "numberOfRequestsFailed";
+
+    public static final String NUMBER_OF_CACHE_ENTRIES = "numberOfCacheEntries";
+
+    public static final String CACHE_HIT_COUNT = "cacheHitCount";
+
+    public static final String CACHE_EVICTION_COUNT = "cacheEvictionCount";
     public static final String LAMBDA_LATENCY_METRIC = "lambdaFunctionLatency";
     public static final String REQUEST_PAYLOAD_SIZE = "requestPayloadSize";
     public static final String RESPONSE_PAYLOAD_SIZE = "responsePayloadSize";
@@ -89,6 +112,11 @@ public class LambdaProcessor extends AbstractProcessor<Record<Event>, Record<Eve
     private final ResponseEventHandlingStrategy responseStrategy;
     private final JsonOutputCodecConfig jsonOutputCodecConfig;
     private final CircuitBreaker circuitBreaker;
+    private final Cache<LambdaCacheKey, Event> cache;
+
+    private final LambdaProcessorConfig.CacheConfig cacheConfig;
+
+    private LambdaResponseMode responseMode;
 
     @DataPrepperPluginConstructor
     public LambdaProcessor(final PluginFactory pluginFactory, final PluginSetting pluginSetting,
@@ -112,6 +140,7 @@ public class LambdaProcessor extends AbstractProcessor<Record<Event>, Record<Eve
                 NUMBER_OF_SUCCESSFUL_REQUESTS_TO_LAMBDA);
         this.numberOfRequestsFailedCounter = pluginMetrics.counter(
                 NUMBER_OF_FAILED_REQUESTS_TO_LAMBDA);
+
         this.lambdaLatencyMetric = pluginMetrics.timer(LAMBDA_LATENCY_METRIC);
         this.requestPayloadMetric = pluginMetrics.summary(REQUEST_PAYLOAD_SIZE);
         this.responsePayloadMetric = pluginMetrics.summary(RESPONSE_PAYLOAD_SIZE);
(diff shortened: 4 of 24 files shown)


data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -121,6 +150,35 @@ public LambdaProcessor(final PluginFactory pluginFactory, final PluginSetting pl
...
+            cache = Caffeine.newBuilder()
+                    .maximumWeight(cacheConfig.getMaxSize())
+                    .weigher(new ByteArrayValueWeigher())
+                    .expireAfterAccess(cacheConfig.getTtl(), TimeUnit.SECONDS)
+                    .removalListener(((key, value, cause) -> {
+                    }))
+                    .recordStats()
+                    .build();
+
+             */
```
graytaylor0: You can remove this comment

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -53,13 +58,31 @@
...
+    }
+
     public static final String NUMBER_OF_RECORDS_FLUSHED_TO_LAMBDA_SUCCESS = "recordsSuccessfullySentToLambda";
     public static final String NUMBER_OF_RECORDS_FLUSHED_TO_LAMBDA_FAILED = "recordsFailedToSentLambda";
     public static final String NUMBER_OF_SUCCESSFUL_REQUESTS_TO_LAMBDA = "numberOfRequestsSucceeded";
     public static final String NUMBER_OF_FAILED_REQUESTS_TO_LAMBDA = "numberOfRequestsFailed";
+
+    public static final String NUMBER_OF_CACHE_ENTRIES = "numberOfCacheEntries";
+
+    public static final String CACHE_HIT_COUNT = "cacheHitCount";
```
graytaylor0, line 78: Would it be better to have a wrapper on the Cache and put metrics there rather than having the processor manage them?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -121,6 +150,35 @@ public LambdaProcessor(final PluginFactory pluginFactory, final PluginSetting pl
...
+            responseMode = LambdaResponseMode.MERGE;
+        } else {
+            cache = null;
+            responseMode = lambdaProcessorConfig.getResponseMode();
+        }
+
+
+        //int valueLength = (v instanceof String) ? ((String)v).length() :
+        //       ((Event)v).toJsonString().length();
+
```
graytaylor0: Another comment

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -281,7 +412,22 @@ List<Record<Event>> convertLambdaResponseToEvent(Buffer flushedBuffer,
                         "FlushedBuffer size:{}", parsedEvents.size(), flushedBuffer.getEventCount(),
                 flushedBuffer.getSize());
         lambdaResponseRecordsCounter.increment(parsedEvents.size());
-        return responseStrategy.handleEvents(parsedEvents, originalRecords);
+        return responseStrategy.handleEvents(parsedEvents, originalRecords, (originalEvent, responseEvent) -> {
+
+            if (cacheConfig != null) {
+                putInCache(originalEvent, responseEvent);
+                /*
+                synchronized (cache) {
```
graytaylor0: Another comment

graytaylor0 (approved):

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: I can address this in a follow up PR

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
```diff
@@ -38,14 +39,16 @@ public class InMemoryBuffer implements Buffer {
     private final long payloadResponseSize;
     private int eventCount;
     private long payloadRequestSize;
+    private List<String> keys;
```
san81, line 42: Could have been better to use `List<EventKey>`  instead of String.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
```diff
@@ -64,6 +67,14 @@ public InMemoryBuffer(String batchOptionKeyName, OutputCodecContext outputCodecC
     public void addRecord(Record<Event> record) {
         records.add(record);
         Event event = record.getData();
+        if (keys != null && keys.size() > 0) {
+            Event newEvent = JacksonEvent.fromEvent(event);
+            newEvent.clear();
```
san81: This may not be performant as we are creating an event from an existing event and dropping all the keys. Instead creating a new event with the new keys and then attaching the metadata could have been more efficient.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/config/LambdaCommonConfig.java
```diff
@@ -52,6 +53,14 @@ public abstract class LambdaCommonConfig {
   @Valid
   private PluginModel responseCodecConfig;
 
+  @JsonPropertyDescription("Keys to send to lambda")
```
san81, line 56: do we allow nested keys here?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKey.java
```diff
@@ -0,0 +1,57 @@
...
+
+package org.opensearch.dataprepper.plugins.lambda.processor;
+
+import org.opensearch.dataprepper.model.event.Event;
+
+import java.util.List;
+import java.util.ArrayList;
+
+public class LambdaCacheKey {
+    private List<Object> keys;
```
san81: making this as final is better
`private final List<Object> keys`

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKey.java
```diff
@@ -0,0 +1,57 @@
...
+                })
+                .sum();
+    }
+
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof LambdaCacheKey)) return false;
+        return keys.equals(((LambdaCacheKey)o).keys);
```
san81, line 54: This also check for same exact order of elements in the source and destination. Probably that is good to enforce but worth calling out in the doc strings of this class.  What this means is, if the Object order inside the List differs then they will be treated as two different cache keys.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -53,13 +58,31 @@
...
+import com.github.benmanes.caffeine.cache.Weigher;
+
+
 @DataPrepperPlugin(name = "aws_lambda", pluginType = Processor.class, pluginConfigurationType = LambdaProcessorConfig.class)
 public class LambdaProcessor extends AbstractProcessor<Record<Event>, Record<Event>> {
 
+    class ByteArrayValueWeigher implements Weigher<LambdaCacheKey, Event> {
+        @Override
+        public int weigh(LambdaCacheKey key, Event value) {
+            return value.toJsonString().length() + key.length();
```
san81: This is very expensive operation. It creates entire copy of the event object as a String. We should look for more efficient ways to get the object size than this.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -121,6 +150,35 @@ public LambdaProcessor(final PluginFactory pluginFactory, final PluginSetting pl
...
+                    .build();
+
+             */
+            cache = createCache(lambdaProcessorConfig.getCacheConfig());
+
+            pluginMetrics.gauge(NUMBER_OF_CACHE_ENTRIES, cache, c -> (double)c.estimatedSize());
+            pluginMetrics.gauge(CACHE_HIT_COUNT, cache, c -> (double)c.stats().hitCount());
+            pluginMetrics.gauge(CACHE_EVICTION_COUNT, cache, c ->(double) c.stats().evictionCount());
+
+            responseMode = LambdaResponseMode.MERGE;
```
san81, line 156: This should come from the cache config - right?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -153,6 +212,59 @@ public LambdaProcessor(final PluginFactory pluginFactory, final PluginSetting pl
...
     }
 
+    private Cache<LambdaCacheKey, Event> createCache(LambdaProcessorConfig.CacheConfig cacheConfig) {
+        return Caffeine.newBuilder()
+                .maximumWeight(cacheConfig.getMaxSize())
+                .weigher(new Weigher<LambdaCacheKey, Event>() {
+                    @Override
+                    public int weigh(LambdaCacheKey key, Event value) {
+                        // The weight is the length of the string value
+                        return key.length() + value.toJsonString().length();
```
san81, line 203: `value.toJsonString().length()` is also an expensive operation and creates a String version copy of `value` object that will double the memory usage.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -153,6 +212,59 @@ public LambdaProcessor(final PluginFactory pluginFactory, final PluginSetting pl
...
+
+    @VisibleForTesting
+    long getCacheEvictionCount() {
+        synchronized (cache) {
+            cache.cleanUp();
+            return cache.stats().evictionCount();
+        }
+    }
+    private Event getFromCache(final Event event) {
+        synchronized (cache) {
```
san81, line 237: I think Caffine takes care of synchronizing and operation being thread-safe and this additional synchronization is not necessary.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -153,6 +212,59 @@ public LambdaProcessor(final PluginFactory pluginFactory, final PluginSetting pl
...
+            LambdaCacheKey cacheKey = new LambdaCacheKey(event, lambdaProcessorConfig.getKeys());
+            return cache.getIfPresent(cacheKey);
+        }
+    }
+
+    private void putInCache(final Event originalEvent, final Event responseEvent) {
+
+        synchronized (cache) {
```
san81: I think Caffine takes care of synchronizing and operation being thread-safe and this additional synchronization is not necessary.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -169,9 +281,28 @@ public Collection<Record<Event>> doExecute(Collection<Record<Event>> records) {
...
+            if (cacheConfig != null) {
+                synchronized (cache) {
+                    List<Object> valuesToFindInCache = new ArrayList<>();
+                    boolean unsupportedValue = false;
+                    //LambdaCacheKey cacheKey = new LambdaCacheKey(event, lambdaProcessorConfig.getKeys());
+
+                    //Event cachedValue = cache.getIfPresent(cacheKey);
+                    Event cachedValue = getFromCache(event);
+                    if (cachedValue != null) {
+                        record.getData().merge(cachedValue);
```
san81: do we always do `merge` here? or it depends on the config `response_mode`?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -169,9 +281,28 @@ public Collection<Record<Event>> doExecute(Collection<Record<Event>> records) {
                 resultRecords.add(record);
                 continue;
             }
+            if (cacheConfig != null) {
+                synchronized (cache) {
```
san81: This `synchronized` block is not needed I guess.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
```diff
@@ -281,7 +412,22 @@ List<Record<Event>> convertLambdaResponseToEvent(Buffer flushedBuffer,
                         "FlushedBuffer size:{}", parsedEvents.size(), flushedBuffer.getEventCount(),
                 flushedBuffer.getSize());
         lambdaResponseRecordsCounter.increment(parsedEvents.size());
-        return responseStrategy.handleEvents(parsedEvents, originalRecords);
+        return responseStrategy.handleEvents(parsedEvents, originalRecords, (originalEvent, responseEvent) -> {
+
+            if (cacheConfig != null) {
+                putInCache(originalEvent, responseEvent);
```
san81, line 396: May be my understanding is wrong here, but what I am thinking is that, we need to put it cache after we apply the response mode of merge or replace. But I see that we are applying response mode in 426 but we are already placing it in the cache here.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/StrictResponseEventHandlingStrategy.java
```diff
@@ -35,7 +37,12 @@ public List<Record<Event>> handleEvents(List<Event> parsedEvents,
...
-            originalEvent.clear();
+            if (consumer != null) {
+                consumer.accept(originalEvent, responseEvent);
+            } else {
+                originalEvent.clear();
+            }
+
 
             // Manually copy each key-value pair from the responseEvent to the originalEvent
             Map<String, Object> responseData = responseEvent.toMap();
```
san81, line 48: We still need to honor the `response_mode` setting here? Like it could be merge or replace. Seems like we are simply replacing it here.

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKeyTest.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Map;
+import java.util.UUID;
+
+public class LambdaCacheKeyTest {
+    private static final EventFactory TEST_EVENT_FACTORY = TestEventFactory.getTestEventFactory();
+    @Test
+    public void testLambdaCacheKey() {
+        List<Object> keys = new ArrayList<>();
+        String testString = UUID.randomUUID().toString();
+        keys.add(testString);
```
san81, line 31: We should add a scenario with nested keys

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKeyTest.java
san81, line 1: We can add additional scenarios for 
- failing to create LambdaCacheKeyTest when the keys are not String or number
- test to see if two LambdaCacheKeys are equal (or not) when the order of keys is different
- Empty keys test or null keys test

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorTest.java
```diff
@@ -288,6 +291,238 @@ public void testProcessorDefaults() {
         assertNotNull(batchOptions);
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = {"lambda-processor-cache-string-key.yaml"})
+    public void testLambdaCachingWithOneStringKey(final String configFileName) throws Exception {
```
san81, line 296: Can we also add some assertions about the cacheHit or cacheMiss?

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorTest.java
```diff
@@ -288,6 +291,238 @@ public void testProcessorDefaults() {
...
+        List<Object> values = new ArrayList<>();
+        values.add("value1");
+        values.add(1234567891230L);
+        values.add(222);
+        List<Record<Event>> records = Collections.singletonList(getSampleRecordWithKeysAndValues("message1", lambdaProcessorConfig.getKeys(), values));
+        LambdaProcessor lambdaProcessor = new LambdaProcessor(pluginFactory, pluginSetting, lambdaProcessorConfig,
+                awsCredentialsSupplier, expressionEvaluator, circuitBreaker);
+        populatePrivateFields(lambdaProcessor);
+
+        InvokeResponse invokeResponse = InvokeResponse.builder()
```
san81, line 415: We should also add a negative case where the given key is not present in the event

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/config/LambdaCommonConfig.java
kkondaka: It should work, right?

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
kkondaka: Yeah. I thought about it. I did initial implementation like this. Let me change it to more performant version

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKey.java
kkondaka: Yeah. Let's keep the order for now. We can always change if needed.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: I don't think there are any other ways. I think, it is OK in this case

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: cacheconfig is making sure that it ALWAYS merge when cache is used. So, it's OK either way. But we shouldn't support REPLACE option even by mistake. That's why I am doing this way.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: Yes. As mentioned above, I don't think we have a better way. We discussed few times. But in this context, I think it is OK because lambda call is more latency inducing operation anyways.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: As mentioned above, we should never support REPLACE mode for cache

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: Response mode for caching is always merge

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/StrictResponseEventHandlingStrategy.java
kkondaka: What do you mean? The code above takes care of replace/merge, right?

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorTest.java
kkondaka: This is done many times in the tests.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
kkondaka: I think I will address this in a follow up PR.

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorTest.java
kkondaka: It is done in many of the test cases

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: Done

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: Done

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: Done

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/accumlator/InMemoryBuffer.java
kkondaka: Made the change

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/common/config/LambdaCommonConfig.java
kkondaka: Added tests for nested keys

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKey.java
kkondaka: Done

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: Will try alternate suggestion provided offline. And if possible include it in the followup PR

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
kkondaka: Removed `synchronized`

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/StrictResponseEventHandlingStrategy.java
kkondaka: The consumer is always non-null the else case is just for test cases.

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKeyTest.java
kkondaka: Added a test.

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaCacheKeyTest.java
kkondaka: Added tests

data-prepper-plugins/aws-lambda/src/test/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessorTest.java
kkondaka: Added cacheHit and cacheMiss count based assertions

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
san81: We will try this approach and see if this helps 

```
public static long getObjectSize(Object obj) {
    try {
        ObjectMapper mapper = new ObjectMapper();
        CountingOutputStream cos = new CountingOutputStream(NullOutputStream.NULL_OUTPUT_STREAM);
        mapper.writeValue(cos, obj);
        return cos.getCount();
    } catch (IOException e) {
    }
}
```

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/LambdaProcessor.java
san81: I think, this need not to be the case but we will discuss on this topic and change this logic if required.

data-prepper-plugins/aws-lambda/src/main/java/org/opensearch/dataprepper/plugins/lambda/processor/StrictResponseEventHandlingStrategy.java
san81: Then we can possibly simplify the logic by removing the null check

san81 (approved):
I am good with some of the comments to be addressed in the follow up PRs

PR #6209, opened 2025-10-30
Fix acknowledgement timeout handling in DefaultAcknowledgementSet

san81:
When acknowledgements time out in DefaultAcknowledgementSet, the callback is never invoked. The current implementation cancels the callback future on timeout but doesn't execute it, making caller unaware of expire event and giving the caller an option to take their required action.

The fix in this PR is to call the callback handler with `false` when acknowledgement set expires, which is kind of same as negative acknowledgment. 

  1. Added callbackInvoked flag to prevent double invocation
  2. Modified isDone() method to invoke callback with false on timeout
  3. Updated complete() and release() methods to set the flag when invoking callback

  The fix ensures that when an acknowledgement set times out, the callback is explicitly invoked with false
 
 
### Check List

### DIFF
--- a/data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetTests.java
+++ b/data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetTests.java
@@ -326,15 +326,15 @@ class DefaultAcknowledgementSetTests {
         // Wait for timeout to occur
         Thread.sleep(150);
 
-        // Trigger timeout check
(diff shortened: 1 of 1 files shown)


data-prepper-core/src/main/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSet.java
```diff
@@ -144,19 +153,16 @@ public boolean isDone() {
         return false;
     }
 
-    public Instant getExpiryTime() {
-        return expiryTime;
-    }
-
```
san81, line 150: Unused method

PR #6211, opened 2025-10-30
Migrate Bitnami Zookeeper and Kafka Docker Images

KarstenSchnitter:
This is just a test to see if the Kafka tests do work due to the very same change contained in the commit of this PR.


KarstenSchnitter:
Apparently the change in the Docker images in #6210  was successful.

PR #6262, opened 2025-11-13
Add support for non-array JSON body in HTTP source

### LINKED ISSUE #6264: HTTP source does not support non-array JSON payload
Data Prepper returns an error when receiving a JSON payload that is not wrapped in an array.
For example: `curl -d '{"key":"value"}' http://localhost:2021/log/ingest` returns `Bad request data format. Needs to be json array.`.

Allow HTTP source to handle both JSON array and JSON object payloads.

We would have to add an additional service layer that can handle both payload formats (like Fluent-Bit or Logstash) before Data Prepper.

**Additional context**

couyangm:
Signed-off-by: Chris Ouyang <couyangm@gmail.com>

### Description
Allow HTTP source to handle a JSON payload that is not nested in an array.
 
### Issues Resolved
Resolves #6264
 
### Check List


data-prepper-plugins/http-source-common/src/main/java/org/opensearch/dataprepper/http/codec/JsonCodec.java
```diff
@@ -66,6 +61,16 @@ public void serializeSplit(final HttpData content, final Consumer<String> serial
     }
 
 
+    private List<Map<String, Object>> readHttpData(final HttpData httpData) throws IOException {
+        try {
+            return mapper.readValue(httpData.toInputStream(), new TypeReference<List<Map<String, Object>>>() {});
+        } catch (IOException e) {
```
dlvenable: @couyangm ,

Please see my comment here:

https://github.com/opensearch-project/data-prepper/issues/6264#issuecomment-3642403749

Regarding the code, we should make a couple changes:

1. Let's make this a configuration. I think for now, it could be something like `accept_single_object: true`.
2. The current change is catching an exception to handle a common condition. This can lead to degraded performance. It is better to add a condition where possible. Can we peek at the `HttpData` to see if the first non-whitespace character is `{` or `[` to make the decision?

dlvenable (changes requested):
@couyangm , Thank you for this change. I left a few comments.

data-prepper-plugins/http-source-common/src/main/java/org/opensearch/dataprepper/http/codec/JsonCodec.java
```diff
@@ -51,8 +47,7 @@ public List<String> parse(final HttpData httpData) throws IOException {
 
     @Override
     public void validate(final HttpData content) throws IOException {
-        mapper.readValue(content.toInputStream(),
-                LIST_OF_MAP_TYPE_REFERENCE);
+        readHttpData(content);
```
dlvenable: We need test coverage on these changes as well.

dlvenable (changes requested):

data-prepper-plugins/http-source-common/src/main/java/org/opensearch/dataprepper/http/codec/JsonCodec.java
couyangm: 1. Added `accept_single_object` property to HTTPSourceConfig
2. Enabled `DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY` in `ObjectMapper` to achieve the desired behaviour.

data-prepper-plugins/http-source-common/src/main/java/org/opensearch/dataprepper/http/codec/JsonCodec.java
couyangm: Added tests for `accept_single_object` to `JsonCodecTest`, `HTTPSourceConfigTest`, and `HTTPSourceTest`.

PR #6320, opened 2025-12-03
set retry time interval configurable, increase the http client read timeout

Zhangxunmt:
### Description
1. There's occasionally read time out issues in ml processing invoking ml-commons to create batch jobs. The error message is below. 
```
Failed to execute HTTP request due to IO issue: Read timed out
```

2. Currently the throttled events are retried every 3 seconds based on the processor doExecute() frequency. In some conditions, retry attempts are hammering the remote Bedrock service causing unstoppable throttling errors. 

This PR increased the http client read timeout from 3 to 30 seconds, and implemented a configurable retry time interval with the default value of 60 seconds to reduce the TPS sent to the remote AI server. Since Bedrock throttles request based on requests per minutes, so 60 seconds should be a good default value. 

### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/MLProcessorConfig.java
+++ b/data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/MLProcessorConfig.java
@@ -9,6 +9,8 @@ import com.fasterxml.jackson.annotation.JsonClassDescription;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyDescription;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import org.hibernate.validator.constraints.time.DurationMax;
+import org.hibernate.validator.constraints.time.DurationMin;
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.NotNull;
 import lombok.Getter;
@@ -33,9 +35,7 @@ import java.util.Map;
 public class MLProcessorConfig {
     private static final int DEFAULT_MAX_BATCH_SIZE = 100;
     public static final Duration DEFAULT_RETRY_WINDOW = Duration.ofMinutes(10);
-    public static final int DEFAULT_RETRY_INTERVAL = 60;    // default retry interval is 1 minute
-    private static final int MIN_RETRY_INTERVAL = 3;
-    public static final int MAX_RETRY_INTERVAL = 300;
+    public static final int DEFAULT_RETRY_INTERVAL_SECONDS = 60;    // default retry interval is 1 minute
 
     @JsonProperty("aws")
     @NotNull
@@ -92,9 +92,18 @@ public class MLProcessorConfig {
     @JsonProperty("retry_time_window")
     private Duration retryTimeWindow = DEFAULT_RETRY_WINDOW;
 
-    @JsonPropertyDescription("The retry interval for the throttled records. Default is 60s.")
-    @JsonProperty(value = "retry_interval_seconds", defaultValue = "" + DEFAULT_RETRY_INTERVAL)
-    private int retry_interval_seconds =  DEFAULT_RETRY_INTERVAL;
+    @JsonPropertyDescription("The retry interval for throttled records. " +
+            "Supports ISO_8601 duration notation (\"PT1M\", \"PT30S\") and simple notation (\"60s\", \"2m\"). " +
+            "Valid range: 3 seconds to 5 minutes. Default is 60 seconds.")
+    @ExampleValues({
+            @ExampleValues.Example(value = "\"PT1M\"", description = "ISO-8601 format for 1 minute"),
+            @ExampleValues.Example(value = "\"60s\"", description = "Simple format for 60 seconds"),
+            @ExampleValues.Example(value = "\"2m\"", description = "Simple format for 2 minutes")
+    })
+    @JsonProperty("retry_interval")
+    @DurationMin(seconds = 3)
+    @DurationMax(seconds = 300)
+    private Duration retryInterval = Duration.ofSeconds(DEFAULT_RETRY_INTERVAL_SECONDS);
 
     @JsonProperty("dlq")
     private PluginModel dlq;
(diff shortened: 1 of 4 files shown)


data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/MLProcessorConfig.java
```diff
@@ -33,6 +33,9 @@
 public class MLProcessorConfig {
     private static final int DEFAULT_MAX_BATCH_SIZE = 100;
     public static final Duration DEFAULT_RETRY_WINDOW = Duration.ofMinutes(10);
+    public static final int DEFAULT_RETRY_INTERVAL = 60;    // default retry interval is 1 minute
+    private static final int MIN_RETRY_INTERVAL = 3;
+    public static final int MAX_RETRY_INTERVAL = 300;
```
oeyh: Nit: suggest specifying the time unit in the name, eg. DEFAULT_RETRY_INTERVAL_SECONDS, etc.

data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/MLProcessorConfig.java
```diff
@@ -89,6 +92,10 @@ public class MLProcessorConfig {
     @JsonProperty("retry_time_window")
     private Duration retryTimeWindow = DEFAULT_RETRY_WINDOW;
 
+    @JsonPropertyDescription("The retry interval for the throttled records. Default is 60s.")
+    @JsonProperty(value = "retry_interval_seconds", defaultValue = "" + DEFAULT_RETRY_INTERVAL)
+    private int retry_interval_seconds =  DEFAULT_RETRY_INTERVAL;
```
oeyh: Use camelCase here: `retryIntervalSeconds`

data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/MLProcessorConfig.java
```diff
@@ -115,4 +122,11 @@ public PluginModel getDlq() {
     public Map<String, Object> getDlqPluginSetting() {
         return dlq != null ? dlq.getPluginSettings() : null;
     }
+
+    public int getRetryIntervalSeconds() {
+        if (retry_interval_seconds < MIN_RETRY_INTERVAL || retry_interval_seconds > MAX_RETRY_INTERVAL) {
+            throw new IllegalArgumentException(String.format("retry interval for throttled records of %d seconds is not valid, valid range is %d - %d", retry_interval_seconds, MIN_RETRY_INTERVAL, MAX_RETRY_INTERVAL));
+        }
```
oeyh: See how validation is done here: https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorConfig.java#L217-L220
Recommend doing the same here.

data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/MLProcessorConfig.java
graytaylor0: I would use `Duration` type for this and rename to `retry_interval`. With duration type users can either put an ISO_8601 time ("PT3M") or a simple duration time ("60s"). You can validate it with annotations (ex: https://github.com/opensearch-project/data-prepper/blob/fc0b563b46204d4195a083d5f235264897112f6f/data-prepper-plugins/cloudwatch-logs/src/main/java/org/opensearch/dataprepper/plugins/sink/cloudwatch_logs/config/ThresholdConfig.java#L49)

data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/common/BedrockBatchJobCreator.java
```diff
@@ -150,16 +154,39 @@ private void processRecord(Record<Event> record, List<Record<Event>> resultRecor
...
             return;
         }
 
         try {
+            if (throttledRecords.isEmpty()) {
+                LOG.debug("Queue became empty after acquiring lock, skipping timestamp update");
+                return;
+            }
+
+            LOG.info("Processing {} throttled records ({}s since last retry)",
```
graytaylor0: Could this potentially be a NOISY log? If so you should mark it as NOISY

data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/MLProcessorConfig.java
Zhangxunmt: I will use the hibernate.validator as the validation per taylor's comment. I think it's fine to just use Durations. 
@DurationMin(seconds = 3)
@DurationMax(seconds = 300)

data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/MLProcessorConfig.java
Zhangxunmt: makes sense to me. Initially I think using Duration is an overkill since we don't need a big interval anyways (probably 5 mins is the biggest). But to make it consistent with retryTimeWindow as a Duration too, I think changing it to Duration is fine.

data-prepper-plugins/ml-inference-processor/src/main/java/org/opensearch/dataprepper/plugins/ml_inference/processor/common/BedrockBatchJobCreator.java
Zhangxunmt: Yes it's Ok to make it Noisy. I thought this is only printed every x minutes so probably not that frequent. I will mark it as NOISY in the next revision since it still repeats every 60s in busy conditions.

PR #6342, opened 2025-12-09
Added logging for No indices matched

### LINKED ISSUE #6341: Improve Logging for OpenSearch Source when there are no matching indices.
Currently, if the are no matching indices, there is no log line post -
```
The opensearch source will start processing data at 2025-12-09T08:27:16.977857090Z. It is currently 2025-12-09T08:27:19.854944258Z
```
Which becomes hard to debug. Moreover, there would be no metrics also. Thus, having a clear log message, in case if post applying the `include_name_regex` and `exclude_name_regex` there are no matching indices, would really help.

Add relevant logging for this situation in [following file](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java)

Utkarsh-Aga:
### Description
Adds a logging in case there are no matching indices. Have added a Noisy marker also, as this log can appear multiple times due to Partition Creation
 
### Issues Resolved
Resolves #6341
 
### Check List
- ~~[ ] New functionality includes testing.~~
- ~~[] New functionality has a documentation issue. Please link to it in this PR.~~
  - ~~[ ] New functionality has javadoc added~~

### DIFF
--- a/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
+++ b/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
@@ -18,6 +18,7 @@ import org.opensearch.dataprepper.plugins.source.opensearch.configuration.OpenSe
 import org.opensearch.dataprepper.plugins.source.opensearch.worker.client.ClusterClientFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import static org.opensearch.dataprepper.logging.DataPrepperMarkers.NOISY;
 
 import java.io.IOException;
 import java.util.Collections;
@@ -26,7 +27,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.function.Function;
 import java.util.regex.Matcher;
-import java.util.stream.Collectors;
 
 public class OpenSearchIndexPartitionCreationSupplier implements Function<Map<String, Object>, List<PartitionIdentifier>> {
 
@@ -82,10 +82,18 @@ public class OpenSearchIndexPartitionCreationSupplier implements Function<Map<St
 
         LOG.debug("Found {} indices", indicesResponse.valueBody().size());
 
-        return indicesResponse.valueBody().stream()
-                .filter(osIndicesRecord -> shouldIndexBeProcessed(osIndicesRecord.index()))
-                .map(indexRecord -> PartitionIdentifier.builder().withPartitionKey(indexRecord.index()).build())
-                .collect(Collectors.toList());
+        final List<PartitionIdentifier> partitions = new java.util.ArrayList<>();
+        int matchCount = 0;
+        for (final var esIndicesRecord : indicesResponse.valueBody()) {
+            if (shouldIndexBeProcessed(esIndicesRecord.index())) {
+                partitions.add(PartitionIdentifier.builder().withPartitionKey(esIndicesRecord.index()).build());
+                matchCount++;
+            }
+        }
+        if (matchCount == 0) {
+            LOG.warn(NOISY, "No indices matched the configured regex patterns after applying include/exclude filters");
+        }
+        return partitions;
     }
 
     private List<PartitionIdentifier> applyForElasticSearchClient(final Map<String, Object> globalStateMap) {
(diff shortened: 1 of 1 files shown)


data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
```diff
@@ -82,10 +82,18 @@ private List<PartitionIdentifier> applyForOpenSearchClient(final Map<String, Obj
...
+        final List<PartitionIdentifier> partitions = new java.util.ArrayList<>();
+        int matchCount = 0;
+        for (final var esIndicesRecord : indicesResponse.valueBody()) {
+            if (shouldIndexBeProcessed(esIndicesRecord.index())) {
+                partitions.add(PartitionIdentifier.builder().withPartitionKey(esIndicesRecord.index()).build());
+                matchCount++;
+            }
+        }
+        if (matchCount == 0) {
+            LOG.info(NOISY, "No indices matched the configured regex patterns after applying include/exclude filters");
```
dlvenable: Should this be a warn log?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
```diff
@@ -82,10 +82,18 @@ private List<PartitionIdentifier> applyForOpenSearchClient(final Map<String, Obj
...
+        }
```
dlvenable, line 95: Is there any existing metric that we can use for alarming? If not, maybe we should add one.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
Utkarsh-Aga: Initially thought to have it at warn level, but then thought since the include_name_regex and exclude_name_regex are user configured, so this log line just provides a information that based on the filters there are no matching index, hence kept it at INFO level. I am open to suggestion, if we should move it to warn log ?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
Utkarsh-Aga: Currently, there are no metrics for this specific case. The nearest one we have is `indicesProcessed` but that can report value zero because of number of reasons. Should we add a new one as part of this pull request or should create another for it ? Additionally, is there a recommended name for that metrics ?
Thank you.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
dlvenable: @Utkarsh-Aga , Do you think there is particular value in running this without finding any indexes? We generally use WARN level for configuration problems in the pipeline. This seems like it would indicate that the user configured their pipeline incorrectly to me.

@graytaylor0 , Any thoughts on this?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
dlvenable: I'd think that a metric like `indicesFound` could make sense. But, it is isn't necessary in this PR. @graytaylor0 , Any thoughts on having a metric like this?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
graytaylor0: I do lean towards agreeing with @dlvenable here. It's unlikely someone would configure a pipeline that has no existing indexes matching the pattern on startup.

Technically, they could have a new index get created that matches the pattern after the pipeline starts, but we would then process an empty or near empty index since it's a one-time migration via point in time. Either way, if someone intends that, then WARN can't hurt. It is far more likely that someone misconfigures the pattern.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/source/opensearch/worker/OpenSearchIndexPartitionCreationSupplier.java
graytaylor0: We have a metric for `partitionsCreated`. There could be value in showing that we have at least discovered indexes through a metric though.

dlvenable:
@Utkarsh-Aga , Let's change the log level to WARN. You can keep the metrics as-is since we have another metric we can use here.

Utkarsh-Aga:
Okay sure, I have updated the log level to warn

graytaylor0 (approved):

PR #6346, opened 2025-12-09
Do not clear offsets after failure to commit offsets due to rebalance exception

graytaylor0:
### Description
This change handles RebalanceInProgressException by not clearing offsets on this exception and retrying, rather than completely clearing offsets and not committing them, which can lead to duplicate processing unnecessarily (in the case that retry would succeed after rebalance)

Tested by modifying the integration test locally to kill one of the consumers and verifying that the data is still processed.
 
### Check List

### DIFF
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
@@ -17,6 +17,7 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.errors.RebalanceInProgressException;
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.TopicPartition;
@@ -354,11 +355,15 @@ public class KafkaCustomConsumer implements Runnable, ConsumerRebalanceListener 
             offsetsToCommit.forEach(((partition, offset) -> updateCommitCountMetric(partition, offset)));
             try {
                 consumer.commitSync(offsetsToCommit);
+                lastCommitTime = currentTimeMillis;
+            } catch (final RebalanceInProgressException ex) {
+                LOG.error("Failed to commit offsets in topic {} due to rebalance in progress", topicName, ex);
+                return;
             } catch (Exception e) {
                 LOG.error("Failed to commit offsets in topic {}", topicName, e);
             }
+
             offsetsToCommit.clear();
-            lastCommitTime = currentTimeMillis;
         }
     }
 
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -354,11 +355,20 @@ private void commitOffsets(boolean forceCommit) {
             offsetsToCommit.forEach(((partition, offset) -> updateCommitCountMetric(partition, offset)));
             try {
                 consumer.commitSync(offsetsToCommit);
+                lastCommitTime = currentTimeMillis;
+            } catch (final RebalanceInProgressException ex) {
+                LOG.error("Failed to commit offsets in topic {} due to rebalance in progress", topicName, ex);
+                try {
+                    Thread.sleep(100);
```
dlvenable: Why do we sleep? Is it related to retrying? If so, can we add some check that we are not still rebalancing?

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -354,11 +355,20 @@ private void commitOffsets(boolean forceCommit) {
...
+                } catch (final InterruptedException ie) {
+                    throw new RuntimeException("Interrupted while waiting to retry after failing to commit offsets");
+                }
+                return;
             } catch (Exception e) {
```
dlvenable, line 362: Currently, this exception set the lastCommitTime. But, now we don't. Is that ok?

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
graytaylor0: The sleep is meant to add some backoff to let the rebalance complete without just spamming commit offsets. Supposedly we have to call `poll` again to complete the rebalance though, so maybe sleep is not that useful.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
graytaylor0: It seems like it was incorrectly saying we committed offsets when we didn't if we get an exception

kkondaka: @dlvenable it is correct because the last commit time should be set only if commit really happens.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
kkondaka: Yes. It looks like sleep() is not needed.

graytaylor0: Pushed a revision removing the sleep

PR #6349, opened 2025-12-11
Add forward_to support to opensearch sink

kkondaka:
### Description
`forward_to` option to all sinks is added via `SinkContext` but the sink level implementation for opensearch sink is not done. This PR adds the support for forwarding events after successful delivery to opensearch sink.

Also added check to not allow more than one forwarding pipeline because that would require acknowledgement support. Forwarding to multiple pipelines an be done after initially forwarding to one pipeline, if it is really needed.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/SinkForwardConfig.java
+++ b/data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/SinkForwardConfig.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.dataprepper.model.configuration;
 
+import org.opensearch.dataprepper.model.plugin.InvalidPluginConfigurationException;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
@@ -31,7 +32,7 @@ public class SinkForwardConfig {
         @JsonProperty("with_data") final Map<String, Object> withData,
         @JsonProperty("with_metadata") final Map<String, Object> withMetadata) {
         if (pipelineNames.size() != 1) {
-            throw new IllegalArgumentException("Supports only one forwarding pipeline");
+            throw new InvalidPluginConfigurationException("Supports only one forwarding pipeline");
         }
         this.pipelineNames = pipelineNames;
         this.withData = withData;
--- a/data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SinkForwardConfigTest.java
+++ b/data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SinkForwardConfigTest.java
@@ -5,6 +5,7 @@
 
 package org.opensearch.dataprepper.model.configuration;
 
+import org.opensearch.dataprepper.model.plugin.InvalidPluginConfigurationException;
 import org.junit.jupiter.api.Test;
 
 import static org.hamcrest.CoreMatchers.equalTo;
@@ -42,12 +43,12 @@ public class SinkForwardConfigTest {
         List<String> pipelines = List.of("pipeline1", "pipeline2");
         Map<String, Object> withData = mock(Map.class);
         Map<String, Object> withMetadata = mock(Map.class);
-        assertThrows(IllegalArgumentException.class, ()->new SinkForwardConfig(pipelines, withData, withMetadata));
+        assertThrows(InvalidPluginConfigurationException.class, ()->new SinkForwardConfig(pipelines, withData, withMetadata));
     }
 
     @Test
     void empty_pipelines_list_throws_exception() {
-        assertThrows(IllegalArgumentException.class, ()->new SinkForwardConfig(List.of(), Map.of(), Map.of()));
+        assertThrows(InvalidPluginConfigurationException.class, ()->new SinkForwardConfig(List.of(), Map.of(), Map.of()));
     }
 }
 


data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/SinkForwardConfig.java
```diff
@@ -30,6 +30,9 @@ public SinkForwardConfig(
         @JsonProperty("pipelines") final List<String> pipelineNames,
         @JsonProperty("with_data") final Map<String, Object> withData,
         @JsonProperty("with_metadata") final Map<String, Object> withMetadata) {
+        if (pipelineNames.size() != 1) {
+            throw new RuntimeException("Supports only one forwarding pipeline");
```
dlvenable: This should throw `InvalidPipelineConfigurationException`.

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SinkForwardConfigTest.java
```diff
@@ -27,13 +28,21 @@ void testDefaults() {
...
         Map<String, Object> withData = mock(Map.class);
         Map<String, Object> withMetadata = mock(Map.class);
         SinkForwardConfig sinkForwardConfig = new SinkForwardConfig(pipelines, withData, withMetadata);
         assertThat(sinkForwardConfig.getPipelineNames(), equalTo(pipelines));
         assertThat(sinkForwardConfig.getWithData(), equalTo(withData));
         assertThat(sinkForwardConfig.getWithMetadata(), equalTo(withMetadata));
     }
+
+    @Test
+    void testInvalidValues() {
```
dlvenable: I'm unclear on this. Is this testing all invalid values? When testing invalid values, you should use all valid values except one specific invalid value. Otherwise, you don't know what you tested against or if it works.

Additionally, you should test invalid values on both ends: empty list and list with two pipelines.

Also, make the names clearer.

e.g.

```
empty_pipelines_list_throws_exception()
pipelines_list_with_two_or_more_pipelines_throws_exception()
```

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/sink/SinkContextTest.java
```diff
@@ -93,17 +93,21 @@ public void testForwardToPipelinesWithPipelineMap() {
         verify(forwardPipeline2, times(0)).sendEvents(eq(records));
         sinkForwardRecordsContext.addRecords(records);
         assertThat(sinkContext.forwardRecords(sinkForwardRecordsContext, Map.of("datakey1", "datavalue1"), Map.of("metadataKey1", "metadataValue1")), equalTo(true));
-        verify(forwardPipeline1, times(1)).sendEvents(eq(records));
-        verify(forwardPipeline2, times(1)).sendEvents(eq(records));
+        verify(forwardPipeline1, times(1)).sendEvents(any());
```
dlvenable, line 96: Why are we loosening these verifies? This doesn't seem right.

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/sink/SinkForwardRecordsContextTest.java
```diff
@@ -66,5 +66,7 @@ public void testSinkForwardRecordContextWithForwardingPipelines() {
         sinkForwardRecordsContext.addRecords(List.of(record2, record3));
         List<Record<Event>> records = sinkForwardRecordsContext.getRecords();
         assertThat(records.size(), equalTo(3));
+        sinkForwardRecordsContext.clearRecords();
```
dlvenable, line 53: This should be it's own test.

```
sinkForwardRecordsContext_clearRecords_removes_all_records()
```

data-prepper-plugins/opensearch/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSinkIT.java
```diff
@@ -846,6 +849,36 @@ public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
         }
     }
 
+    @Test
+    public void testForwardingRecords() throws IOException, InterruptedException {
```
dlvenable: This this a better name and scope down what it is doing. 

```
output_forwards_created_documents_to...()
```

Also, as this is an IT, I think it would be better to actually forward these to another OpenSearch index.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkRetryStrategy.java
```diff
@@ -384,6 +391,7 @@ private AccumulatingBulkRequest<BulkOperationWrapper, BulkRequest> createBulkReq
             final AccumulatingBulkRequest requestToReissue = bulkRequestSupplier.get();
             final ImmutableList.Builder<FailedBulkOperation> nonRetryableFailures = ImmutableList.builder();
             int index = 0;
+            List<BulkOperationWrapper> successfulOperations = new ArrayList<>();
```
dlvenable: Initialize this array with the total number of operations, assuming the best case.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -592,6 +598,27 @@ private void flushBatch(AccumulatingBulkRequest accumulatingBulkRequest) {
     });
   }
 
+  private void successfulOperationsHandler(final List<BulkOperationWrapper> successfulOperations) {
```
dlvenable: This needs unit tests. There are a lot of conditions to cover as well.

data-prepper-plugins/opensearch/src/test/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkRetryStrategyTests.java
```diff
@@ -129,6 +129,16 @@ public BulkRetryStrategy createObjectUnderTest(
         return new BulkRetryStrategy(
                 requestFunction,
                 logFailure,
+                (operations) -> {
```
dlvenable, line 132: This moves some unnecessary logic into the tests. I think you can just make a `mock(Consumer.class)` and then call `verify(successOperationsHandler).accept(successOperations)`.

We probably need tests to verify that failed operations are not included.

dlvenable (changes requested):

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
graytaylor0: Did you consider making this handle List<Event> and moving the logic to a higher level to make it re-usable between the sinks? Use bulk operation wrapper makes this very coupled to opensearch sink

graytaylor0:
The OpenSearchSinkIT are all failing with this error

```
 OpenSearchSinkIT > testOpenSearchIndexWithDate(String) > [2] yyyy-MM-dd FAILED
    java.lang.NullPointerException at OpenSearchSinkIT.java:159

OpenSearchSinkIT > testOpenSearchIndexWithDate(String) > [3] dd-MM-yyyy FAILED
    java.lang.NullPointerException at OpenSearchSinkIT.java:159

OpenSearchSinkIT > testOpenSearchRouting(String) > [1]  FAILED
    java.lang.NullPointerException at OpenSearchSinkIT.java:159

OpenSearchSinkIT > testOpenSearchRouting(String) > [2] info/ids/rid FAILED
    java.lang.NullPointerException at OpenSearchSinkIT.java:159

OpenSearchSinkIT > testOpenSearchRouting(String) > [3] rid FAILED
    java.lang.NullPointerException at OpenSearchSinkIT.java:159
```

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SinkForwardConfigTest.java
kkondaka: Fair point. Added two negative tests to test 0 and two pipelines. Added one positive test to test with one pipeline.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/SinkForwardConfig.java
kkondaka: @dlvenable I do not think it can throw `InvalidPipelineConfigurationException` because it is in a different package that depends on `data-prepper-api` package. It will introduce circular dependency between the packages.

kkondaka: Changed it to `IllegalArgumentException`

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/sink/SinkContextTest.java
kkondaka: No. We are not loosening the verifies. Since the `forwardRecords()` clears the `records` It can't be verified like that anymore. The original `records` would now be empty and the verification would fail

data-prepper-plugins/opensearch/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSinkIT.java
kkondaka: @dlvenable Yes, it is IT but to test forwarding to another Opensearch sink requires complete pipeline functionality. We are not forwarding to another sink but forwarding to a pipeline. Goal is to verify that it is forwarding to pipelines. I do not think we need to mock entire pipeline. We could probably send to a sink directly from 
`sinkContext.forwardRecords()` But sending another opensearch sink really doesn't test any more functionality than what is being tested already.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkRetryStrategy.java
kkondaka: @dlvenable You mean this?
```
successfulOperations = new ArrayList<>(request.getOperations.size());
```

data-prepper-plugins/opensearch/src/test/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkRetryStrategyTests.java
kkondaka: @dlvenable Unfortunately, that wont work. The tests count on eventhandle getting released. So, it is important to release the event handle in this Consumer.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
kkondaka: @graytaylor0 by "handle list", you mean "event handle list"? We can't forward event handles. It needs to be events. I don't think this part of the code would be that much re-usable in other Sinks.  We are moving towards using only pipeline DLQ, which means all sinks need to keep the events, and just event handles. Due to legacy code we have multiple conditions in opensearch.

data-prepper-plugins/opensearch/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSinkIT.java
dlvenable: Yea, I think you are right here. This is an integration test of OpenSearch functionality, not core Data Prepper.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkRetryStrategy.java
dlvenable: Yea.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/SinkForwardConfig.java
dlvenable: @kkondaka ,

We can use `InvalidPluginConfigurationException` instead.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
graytaylor0: I meant to say `List<Event>` sorry

PR #6357, opened 2025-12-15
remove usage of buffer accumulator from Kafka custom consumer

graytaylor0:
### Description
This change removes the buffer accumulator in favor of directly calling `buffer.writeAll`. This is because BufferAccumulator does not throw TimeoutException, and instead will do infinite back off and retry to write to the buffer during TimeoutExceptions. This causes the consumer thread processing the records to get blocked, which stops `poll` from getting called on the consumer, which can result in the consumer being removed from the group after the `max.poll.interval.ms` expires, which leads to Kafka rebalances. 
 
### Check List

### DIFF
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
@@ -17,13 +17,12 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
-import org.apache.kafka.common.errors.RebalanceInProgressException;
-import org.apache.kafka.common.header.Header;
-import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.AuthenticationException;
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -492,23 +488,19 @@ private <T> Record<Event> getRecord(ConsumerRecord<String, T> consumerRecord, in
...
         long numRetries = 0;
         while (true) {
             LOG.debug("In while loop for processing records, paused = {}", paused);
             try {
-                if (numRetries == 0) {
-                    bufferAccumulator.add(record);
-                } else {
-                    bufferAccumulator.flush();
-                }
+                buffer.writeAll(eventRecords, BUFFER_WRITE_TIMEOUT);
```
kkondaka, line 503: I think the "topicMetrics.getNumberOfBufferSizeOverflows().increment();" needs to be modified as well to keep track of how times we got this timeout exception. Maybe check for "SizeOverflowException" can be removed as well?

graytaylor0: Well we are still tracking `SizeOverflowException` only in the code below where we check the exception type. If you look at the buffer code you can see that we already have a metric for TimeoutException here (https://github.com/opensearch-project/data-prepper/blob/860e59301dd4f3d9210827877338c4753a50aac4/data-prepper-api/src/main/java/org/opensearch/dataprepper/model/buffer/AbstractBuffer.java#L128). Under the hood the BufferAccumulator was just calling buffer.writeAll() so it's the same code path as far as metrics go

PR #6365, opened 2025-12-22
Add rolling buffer to Prometheus sink

kkondaka:
### Description
Add rolling buffer to Prometheus sink
1. A time based window can be configured using "out_of_order_window" config parameter which is used to keep last few milli seconds ( defined in out_of_order_window with default value of 5 seconds) of buffer as rolling buffer. Any events before this time are flushed and after flushing them, older events are rejected. 
2. This rolling buffer is maintained for each metric separately. Time window of metric1 doesn't interfere with time window of metric2
3. for a given metric, samples are returned in time sorted way
4. for a given metric, labels are sorted by the name
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusSinkBufferWriter.java
+++ b/data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusSinkBufferWriter.java
@@ -9,51 +9,104 @@
   */
 package org.opensearch.dataprepper.plugins.sink.prometheus.service;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import org.opensearch.dataprepper.common.sink.SinkBufferEntry;
 import org.opensearch.dataprepper.common.sink.SinkBufferWriter;
 import org.opensearch.dataprepper.common.sink.SinkFlushableBuffer;
 import org.opensearch.dataprepper.common.sink.SinkFlushContext;
 import org.opensearch.dataprepper.common.sink.SinkMetrics;
+import org.opensearch.dataprepper.plugins.sink.prometheus.configuration.PrometheusSinkConfiguration;
+import software.amazon.awssdk.utils.Pair;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 public class PrometheusSinkBufferWriter implements SinkBufferWriter {
 
+    private static final int MAX_EVENTS_PER_SERIES = 1000;
     // Each buffer entry is a metric.
     // Duplicate entries for the same metric at the same time is not allowed
     // If there are multiple entries in the buffer for the same metric (with different time stamps),
     // They must be sorted by time before sending to Prometheus
-    private final Map<String, Map<Long, SinkBufferEntry>> buffer;
+    private final Map<String, PrometheusSinkBufferWriterEntry> buffer;
     private final SinkMetrics sinkMetrics;
+    private final long outOfOrderWindowMillis;
+    private final long maxEvents;
+    private final long maxRequestSize;
 
-    public PrometheusSinkBufferWriter(SinkMetrics sinkMetrics) {
+    public PrometheusSinkBufferWriter(final PrometheusSinkConfiguration sinkConfig, final SinkMetrics sinkMetrics) {
         this.buffer = new HashMap<>();
         this.sinkMetrics = sinkMetrics;
+        this.outOfOrderWindowMillis = sinkConfig.getOutOfOrderWindow().toMillis();
+        this.maxEvents = sinkConfig.getThresholdConfig().getMaxEvents();
+        this.maxRequestSize = sinkConfig.getThresholdConfig().getMaxRequestSizeBytes();
     }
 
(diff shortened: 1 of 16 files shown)


data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusSinkBufferWriter.java
```diff
@@ -9,51 +9,104 @@
...
+        long curMaxEvents = maxEvents;
+        long curMaxSize = maxRequestSize;
+        long reqSize = 0L;
+        for (Map.Entry<String, PrometheusSinkBufferWriterEntry> entry : buffer.entrySet()) {
+            if (curMaxEvents == 0 || curMaxSize == 0) {
+                break;
+            }
+            Pair<List<PrometheusSinkBufferEntry>, Long> result = entry.getValue().getEntriesReadyToFlush(curMaxEvents, curMaxSize);
+            if (!result.left().isEmpty()) {
+                curMaxEvents -= result.left().size();
```
sb2k16, line 98: How do we handle the negative case?

data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusSinkFlushableBuffer.java
```diff
@@ -39,15 +40,15 @@ public PrometheusSinkFlushableBuffer(List<SinkBufferEntry> buffer, final SinkMet
...
     public SinkFlushResult flush() {
-        if (buffer.size() == 0) {
+        if (buffer.isEmpty()) {
             return null;
         }
 
         PrometheusHttpSender httpSender = sinkFlushContext.getHttpSender();
         final Remote.WriteRequest.Builder writeRequestBuilder = Remote.WriteRequest.newBuilder();
-        List<Types.TimeSeries> allTimeSeries = new ArrayList<>();
+        List<Types.TimeSeries> allTimeSeries = new ArrayList<>(buffer.size() * 2);
```
sb2k16, line 49: why are we using 2?

data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusSinkBufferWriter.java
sb2k16: I think we are missing some tests to cover this.

data-prepper-plugins/prometheus-sink/src/test/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusSinkServiceTest.java
```diff
@@ -51,14 +51,14 @@ public class PrometheusSinkServiceTest {
...
     private ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory().enable(YAMLGenerator.Feature.USE_PLATFORM_LINE_BREAKS));
-        
 
     private static final String SINK_YAML =
             "        url: \"http://localhost:8080/test\"\n" +
             "        threshold:\n"+
             "          max_events: 2\n" +
             "          flush_interval: 10\n"+
             "        connection_timeout: 10\n"+
+            "        out_of_order_window: 0\n" +
```
sb2k16, line 61: Perhaps, we should add a test for default behavior without this setting.

data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusSinkBufferWriter.java
kkondaka: Result empty would continue to the next. 

There are parameterized tests in the unit test file that covers different cases.

data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusSinkFlushableBuffer.java
kkondaka: It is just a heuristic that AI came up with to preallocate.

data-prepper-plugins/prometheus-sink/src/test/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusSinkServiceTest.java
kkondaka: AMPIT tests cover for this setting set to different values. By default it's value is 5 seconds which is covered in those integration tests

kkondaka: But I can add another unit test as a fast follow-up. But don't want to delay this for that.

srikanthjg (approved):

PR #6369, opened 2025-12-29
Add support for DECIMAL logical type in fixed-length byte arrays when parsing Parquet files

### LINKED ISSUE #6339: [BUG] Rds does not handle mysql decimal data types when precision value is 19 or higher
**Describe the bug**
Data Prepper RDS source does not handle MySQL DECIMAL data types with precision ≥19. When RDS exports snapshots to S3, values ≥2^63 are encoded as FIXED_LEN_BYTE_ARRAY in Parquet format (per [AWS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_ExportSnapshot.data-types.html)). The NumericTypeHandler does not support this encoding, causing pipeline failures.

To Reproduce
1. Create MySQL table with DECIMAL(38,0) column containing values ≥2^63
2. Export RDS snapshot to S3
3. Configure Data Prepper pipeline with RDS source pointing to the export
4. Pipeline fails with error:
```
java.lang.IllegalArgumentException: Unsupported value type. The value is of type: class java.util.ArrayList
  at org.opensearch.dataprepper.plugins.source.rds.datatype.mysql.handler.NumericTypeHandler.handleNumericType
```

**Expected behavior**
RDS source should handle all MySQL DECIMAL types regardless of precision, converting FIXED_LEN_BYTE_ARRAY (ArrayList<Byte>) to BigInteger for values that exceed INT64 range.

**Screenshots**

 - OS: [e.g. Ubuntu 20.04 LTS]

**Additional context**

divbok:
### Description
Adds proper deserialization of DECIMAL logical types stored as GenericFixed in Parquet files. Converts fixed byte arrays to BigDecimal using the schema's scale, then serializes as numeric values. Includes fallback handling for non-decimal fixed arrays as JSON byte objects.

This resolves issues where mysql decimal columns with large precision were not handled. 
 
### Issues Resolved
Resolves #6339 
 
### Check List

### DIFF
--- a/data-prepper-plugins/parquet-codecs/src/main/java/org/opensearch/dataprepper/plugins/codec/parquet/GenericRecordJsonEncoder.java
+++ b/data-prepper-plugins/parquet-codecs/src/main/java/org/opensearch/dataprepper/plugins/codec/parquet/GenericRecordJsonEncoder.java
@@ -21,6 +21,7 @@ import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericContainer;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericEnumSymbol;
+import org.apache.avro.generic.GenericFixed;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.generic.IndexedRecord;
 import org.apache.commons.text.StringEscapeUtils;
@@ -84,8 +85,8 @@ public class GenericRecordJsonEncoder {
                 if (fieldSchema.getType() == Schema.Type.UNION) {
                     for (Schema s : fieldSchema.getTypes()) {
                         if (s.getType() != Schema.Type.NULL) {
-                            if (s.getType() == Schema.Type.BYTES &&
-                                    s.getLogicalType() instanceof LogicalTypes.Decimal) {
+                            if ((s.getType() == Schema.Type.BYTES || s.getType() == Schema.Type.FIXED)
+                                    && s.getLogicalType() instanceof LogicalTypes.Decimal) {
                                 serialize(logicalTypeConverter.apply(getField(datum, f.name(), f.pos())), buffer, seenObjects, ((LogicalTypes.Decimal) s.getLogicalType()).getScale());
                                 serializedDecimal = true;
                                 break;
@@ -172,7 +173,22 @@ public class GenericRecordJsonEncoder {
             buffer.append("\"");
             buffer.append(datum);
             buffer.append("\"");
-        } else if (datum instanceof GenericData) {
+        } else if (datum instanceof GenericFixed ) {
+            GenericFixed fixed = (GenericFixed) datum;
+            byte[] bytes = fixed.bytes();
+            if (decimalScale != null) {
+                BigInteger unscaledValue = new BigInteger(bytes);
+                BigDecimal decimal = new BigDecimal(unscaledValue, decimalScale);
+                buffer.append(decimal.toString());
+            }
+            //Fallback mechanism
+            else {
+                buffer.append("{\"bytes\": \"");
+                writeEscapedString(new String(bytes, StandardCharsets.ISO_8859_1), buffer);
+                buffer.append("\"}");
+            }
+        }
+        else if (datum instanceof GenericData) {
             if (seenObjects.containsKey(datum)) {
                 buffer.append(TOSTRING_CIRCULAR_REFERENCE_ERROR_TEXT);
                 return;
--- a/data-prepper-plugins/parquet-codecs/src/test/java/org/opensearch/dataprepper/plugins/codec/parquet/GenericRecordJsonEncoderTest.java
+++ b/data-prepper-plugins/parquet-codecs/src/test/java/org/opensearch/dataprepper/plugins/codec/parquet/GenericRecordJsonEncoderTest.java
@@ -316,4 +316,69 @@ class GenericRecordJsonEncoderTest {
         assertEquals("{\"amount\": null}", json);
     }
 
+    @Test
+    void serialize_WithFixedDecimalLogicalType_UsesScaleFromSchema() {
+        BigDecimal value = new BigDecimal("12.34").setScale(2);
+        byte[] decimalBytes = value.unscaledValue().toByteArray();
+        
+        Schema decimalSchema = new Schema.Parser().parse(
+                "{ \"type\": \"record\", \"name\": \"DecimalRecord\", \"fields\": [" +
+                        "{\"name\": \"amount\", \"type\": [\"null\", {\"type\":\"fixed\",\"size\":" + decimalBytes.length +
+                        ",\"name\":\"DecimalFixed\",\"logicalType\":\"decimal\",\"precision\":4,\"scale\":2}]}" +
+                        "] }"
+        );
+
+        GenericRecord record = new GenericData.Record(decimalSchema);
+        
+        Schema fixedSchema = decimalSchema.getField("amount").schema().getTypes().get(1);
+        GenericData.Fixed fixedValue = new GenericData.Fixed(fixedSchema, decimalBytes);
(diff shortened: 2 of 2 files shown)


data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/datatype/mysql/handler/NumericTypeHandler.java
```diff
@@ -107,4 +114,29 @@ private static BigInteger bitSetToBigInteger(BitSet bitSet) {
...
     }
+
+    private Number handleByteArray(final Object value) {
+        if (value instanceof byte[]) {
+            return new BigDecimal(new BigInteger((byte[]) value));
+        }
+
+        if (value instanceof Map) {
+            Object data = ((Map<?, ?>)value).get(BYTES_KEY);
+            if (data instanceof byte[]) {
```
oeyh: Is `data` always of `byte[]` type? Do we need an else block here?

divbok: Did not encounter the Map use case. In my tests the values were an ArrayList type

divbok:
> Thanks for the fix! Were you able to verify the changes with export and stream data?

Yes, I was able to verify the changes.

data-prepper-plugins/rds-source/src/test/java/org/opensearch/dataprepper/plugins/source/rds/datatype/mysql/handler/NumericTypeHandlerTest.java
```diff
@@ -113,7 +113,12 @@ private static Stream<Arguments> provideNumericTypeData() {
 
                 // BIT tests
                 Arguments.of(MySQLDataType.BIT, "bit_col", BitSet.valueOf(new byte[]{ 4, 3, 2, 1 }), new BigInteger("16909060")), // BitSet interprets the bytes in little-endian order
-                Arguments.of(MySQLDataType.BIT, "bit_col", Map.of("bytes", new byte[]{ 1, 2, 3, 4 }), new BigInteger("16909060")) // Direct BigInteger interprets the bytes in big-endian order.
+                Arguments.of(MySQLDataType.BIT, "bit_col", Map.of("bytes", new byte[]{ 1, 2, 3, 4 }), new BigInteger("16909060")), // Direct BigInteger interprets the bytes in big-endian order.
+
+                //DECIMAL tests represented as byte arrays
+                Arguments.of(MySQLDataType.DECIMAL, "decimal_col", new ArrayList<>(List.of(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16)), new BigDecimal("1339673755198158349044581307228491536")),
```
san81: Could be worth testing negative and fractional values as well, converted to BigDecimal?

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/datatype/mysql/handler/NumericTypeHandler.java
```diff
@@ -107,4 +114,29 @@ private static BigInteger bitSetToBigInteger(BitSet bitSet) {
         }
         return result;
     }
+
+    private Number handleByteArray(final Object value) {
+        if (value instanceof byte[]) {
+            return new BigDecimal(new BigInteger((byte[]) value));
```
dlvenable: Each of these conditions at some point calls this: `new BigDecimal(new BigInteger(someBytes)`. Refactor this out into a method so that changes to one affect others.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/datatype/mysql/handler/NumericTypeHandler.java
dlvenable: I think we should still have the else condition to log this case if it ever happens. Otherwise you will get the exception on line 139, which would make it seem that we didn't find the right type.

divbok:
Previous implementation failed to apply the correct decimal scale during deserialization, resulting in incorrect numeric values. The fixed length byte array when passed directly to MYSQL Numeric handler loses the scale information. This change extracts the scale from the Avro schema's Decimal logical type and applies it when converting fixed byte arrays to BigDecimal, ensuring accurate decimal representation in the output JSON when reading from the Parquet files.

data-prepper-plugins/parquet-codecs/src/test/java/org/opensearch/dataprepper/plugins/codec/parquet/GenericRecordJsonEncoderTest.java
```diff
@@ -316,4 +316,69 @@ void serialize_WithNullDecimalLogicalType_ReturnsNull() {
...
+
+        assertEquals("{\"amount\": 12.3456}", json);
+    }
+
+    @Test
+    void serialize_WithNonDecimalFixedType_ReturnsEscapedString() {
+        byte[] tokenBytes = new byte[] { 34, 92, 13, 10, 9};
+        Schema tokenSchema = new Schema.Parser().parse(
+                "{ \"type\": \"record\", \"name\": \"MyRecord\", \"fields\": [" +
+                        "{\"name\": \"token\", \"type\": {\"type\":\"fixed\",\"name\":\"TokenFixed\",\"size\":4}}" +
```
oeyh: Size should be 5?

data-prepper-plugins/parquet-codecs/src/main/java/org/opensearch/dataprepper/plugins/codec/parquet/GenericRecordJsonEncoder.java
```diff
@@ -172,7 +173,22 @@ private void serialize(final Object datum, final StringBuilder buffer,
...
             buffer.append(datum);
             buffer.append("\"");
-        } else if (datum instanceof GenericData) {
+        } else if (datum instanceof GenericFixed ) {
+            GenericFixed fixed = (GenericFixed) datum;
+            byte[] bytes = fixed.bytes();
+            if (decimalScale != null) {
+                BigInteger unscaledValue = new BigInteger(bytes);
+                BigDecimal decimal = new BigDecimal(unscaledValue, decimalScale);
+                buffer.append(decimal.doubleValue());
```
oeyh: Downstream Json parser may treat it as Double but I suggest we preserve the precision here with `decimal.toPlainString()` or `decimal.toString()`

oeyh (commented):
Thanks for making the change. With this change, what's the type that MySQL NumericHandler will get for Decimals?

divbok:
> Thanks for making the change. With this change, what's the type that MySQL NumericHandler will get for Decimals?

It will be Number

data-prepper-plugins/parquet-codecs/src/test/java/org/opensearch/dataprepper/plugins/codec/parquet/GenericRecordJsonEncoderTest.java
divbok: Will make this change.

data-prepper-plugins/parquet-codecs/src/main/java/org/opensearch/dataprepper/plugins/codec/parquet/GenericRecordJsonEncoder.java
divbok: Made the change

san81 (approved):

PR #6370, opened 2025-12-30
Fix Data Prepper router to send records through routing strategy before sending to the sinks

kkondaka:
### Description
Fix Data Prepper router to send records through routing strategy before sending to the sinks

Basic issue is that when there are multiple sinks (say 2), the records are sent to them serially by sending all records to sink1 first and when that's complete (`sink.output()` returns) the records are sent to second sink. If the first sink writes to the external sink without buffering, the events are "released" resulting in successful ACK to the source even before the records are sent to the second sink. Usually this is not an issue because the events are buffered before sending to external sink. But it is possible that some may not go through buffering resulting in premature acknowledgement to the source.

This fix addresses it by modifying the routing code to go through routing strategy before sending the events to sinks.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-core/src/integrationTest/java/org/opensearch/dataprepper/integration/Router_ThreeRoutesDefaultIT.java
+++ b/data-prepper-core/src/integrationTest/java/org/opensearch/dataprepper/integration/Router_ThreeRoutesDefaultIT.java
@@ -9,31 +9,44 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.opensearch.dataprepper.model.event.Event;
+import org.opensearch.dataprepper.model.event.DefaultEventHandle;
 import org.opensearch.dataprepper.model.event.JacksonEvent;
 import org.opensearch.dataprepper.model.record.Record;
 import org.opensearch.dataprepper.plugins.InMemorySinkAccessor;
 import org.opensearch.dataprepper.plugins.InMemorySourceAccessor;
 import org.opensearch.dataprepper.test.framework.DataPrepperTestRunner;
 
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.ArgumentMatchers.any;
 import static org.awaitility.Awaitility.await;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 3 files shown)


sb2k16 (approved):

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/pipeline/router/Router.java
```diff
@@ -41,6 +42,8 @@ public <C> void route(
         Objects.requireNonNull(dataFlowComponents);
         Objects.requireNonNull(componentRecordsConsumer);
 
+        final Map<Object, Collection<Record>> componentRecords = new HashMap<>();
```
san81, line 45: Just a small suggestion.  Instead of Object type, it can be of `C` type so that we can avoid explicit type conversion later in the code on line 67.

```
final Map<C, Collection<Record>> componentRecords = new HashMap<>();
```

san81: Also, Not sure if order matters here. With HashMap, the order of sending the events to different sinks will change, which could potentially create a different behavior? May be better to use LinkedHash to keep the behavior consistent with the existing code.

```
final Map<C, Collection<Record>> componentRecords = new LinkedHashMap<>();
```

data-prepper-core/src/integrationTest/resources/org/opensearch/dataprepper/pipeline/route/three-route-with-default-route.yaml
```diff
@@ -4,6 +4,7 @@ routing-pipeline:
   source:
     in_memory:
       testing_key: ConditionalRoutingIT
+      acknowledgments: true
```
dlvenable, line 7: Let's not change the existing tests. You should add a different set of tests here instead.

data-prepper-core/src/integrationTest/java/org/opensearch/dataprepper/integration/Router_ThreeRoutesDefaultIT.java
```diff
@@ -120,8 +167,12 @@ void test_default_route() {
 
     private List<Record<Event>> createEvents(final String value, final int numberToCreate) {
         return IntStream.range(0, numberToCreate)
-                .mapToObj(i -> Map.of(KNOWN_CONDITIONAL_KEY, value, "arbitrary_field", UUID.randomUUID().toString()))
-                .map(map -> JacksonEvent.builder().withData(map).withEventType("TEST").build())
+                .mapToObj(i -> Map.of(KNOWN_CONDITIONAL_KEY, value, "id", i, "arbitrary_field", UUID.randomUUID().toString()))
+                .map(map -> {
+                    DefaultEventHandle eventHandle = eventHandles.get(value+map.get("id"));
+                    assertNotNull(eventHandle);
+                    return JacksonEvent.builder().withData(map).withEventType("TEST").withEventHandle(eventHandle).build();
```
dlvenable, line 174: Let's use the EventFactory here and going forward.

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/pipeline/router/Router.java
dlvenable: I agree with both of these comments.

dlvenable: Nice. I see that these are now in #6371.

PR #6372, opened 2025-12-30
Add metric tracking time between poll calls for kafka consumer

graytaylor0:
### Description
This change adds a metric to the KafkaCustomerConsumer `actualPollInterval` that tracks how often the consumers are calling poll. This metric is useful because if poll is not called often enough based on the `max.poll.interval.ms` setting, there will be Rebalances on the kafka topic.

### Check List


kkondaka (approved):
Thanks for renaming the metric as discussed.

PR #6394, opened 2026-01-07
Improve RemotePeerForwarder Performance

KarstenSchnitter:
### Description
I am looking into improving the performance of the `RemotePeerForwarder`. I introduced a JMH benchmark to measure improvements. This lead to a first set of changes to improve throughput especially on smaller batch size. The next change removes synchronization during forwarding of the requests. Finally, I improved the receiving of requests by introducing a heuristic to abandon the batchDelay when enough data is expected in the queue. Major performance gains achieved by this PR:

- caching of local IPs reduces latencies on forwarding of small batches
- asynchronous forwarding removes synchronization between different peers
- reading from a (expectedly) full buffer is done with minimum delay to avoid capping throughput by `batchDelay`
 
### Issues Resolved

- #2147
 
### Check List

### DIFF
--- a/data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
+++ b/data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
@@ -22,12 +22,10 @@ import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.net.UnknownHostException;
 import java.time.Duration;
-import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -56,6 +54,7 @@ class RemotePeerForwarder implements PeerForwarder {
     private final Set<String> identificationKeys;
     final ConcurrentHashMap<String, LinkedBlockingQueue<Record<Event>>> peerBatchingQueueMap;
     private final ConcurrentHashMap<String, Long> peerBatchingLastFlushTimeMap;
+    private final ConcurrentHashMap<String, Boolean> localAddressCache;
 
     private final Counter recordsActuallyProcessedLocallyCounter;
     private final Counter recordsToBeProcessedLocallyCounter;
@@ -99,6 +98,7 @@ class RemotePeerForwarder implements PeerForwarder {
         this.pipelineWorkerThreads = pipelineWorkerThreads;
         peerBatchingQueueMap = new ConcurrentHashMap<>();
         peerBatchingLastFlushTimeMap = new ConcurrentHashMap<>();
+        localAddressCache = new ConcurrentHashMap<>();
         
         recordsActuallyProcessedLocallyCounter = pluginMetrics.counter(RECORDS_ACTUALLY_PROCESSED_LOCALLY);
         recordsToBeProcessedLocallyCounter = pluginMetrics.counter(RECORDS_TO_BE_PROCESSED_LOCALLY);
@@ -149,13 +149,13 @@ class RemotePeerForwarder implements PeerForwarder {
             final Collection<Record<Event>> records,
             final Set<String> identificationKeys
     ) {
-        final Map<String, List<Record<Event>>> groupedRecords = new HashMap<>();
+        final Map<String, List<Record<Event>>> groupedRecords = new HashMap<>(hashRing.getPeerCount());
 
         // group records based on IP address calculated by HashRing
         for (final Record<Event> record : records) {
             final Event event = record.getData();
 
-            final List<String> identificationKeyValues = new LinkedList<>();
+            final List<String> identificationKeyValues = new ArrayList<>(identificationKeys.size());
             int numMissingIdentificationKeys = 0;
             for (final String identificationKey : identificationKeys) {
                 final Object identificationKeyValue = event.get(identificationKey, Object.class);
@@ -178,21 +178,23 @@ class RemotePeerForwarder implements PeerForwarder {
     }
 
     private boolean isAddressDefinedLocally(final String address) {
-        final InetAddress inetAddress;
-        try {
-            inetAddress = InetAddress.getByName(address);
-        } catch (final UnknownHostException e) {
-            return false;
-        }
-        if (inetAddress.isAnyLocalAddress() || inetAddress.isLoopbackAddress()) {
-            return true;
-        } else {
+        return localAddressCache.computeIfAbsent(address, addr -> {
+            final InetAddress inetAddress;
             try {
-                return NetworkInterface.getByInetAddress(inetAddress) != null;
-            } catch (final SocketException e) {
+                inetAddress = InetAddress.getByName(addr);
+            } catch (final UnknownHostException e) {
                 return false;
             }
-        }
+            if (inetAddress.isAnyLocalAddress() || inetAddress.isLoopbackAddress()) {
+                return true;
+            } else {
+                try {
+                    return NetworkInterface.getByInetAddress(inetAddress) != null;
+                } catch (final SocketException e) {
+                    return false;
+                }
+            }
+        });
     }
 
     private List<Record<Event>> batchRecordsForForwarding(final String destinationIp, final List<Record<Event>> records) {
@@ -294,12 +296,14 @@ class RemotePeerForwarder implements PeerForwarder {
     }
 
     private boolean shouldFlushBatch(final String destinationIp) {
-        final long currentTime = System.currentTimeMillis();
-        final long millisSinceLastFlush = currentTime - peerBatchingLastFlushTimeMap.getOrDefault(destinationIp, System.currentTimeMillis());
-        final Duration durationSinceLastFlush = Duration.of(millisSinceLastFlush, ChronoUnit.MILLIS);
+        final LinkedBlockingQueue<Record<Event>> queue = peerBatchingQueueMap.get(destinationIp);
+        if (queue.size() >= forwardingBatchSize) {
+            return true;
+        }
 
-        final boolean shouldFlushDueToTimeout = durationSinceLastFlush.compareTo(forwardingBatchTimeout) >= 0;
-        return shouldFlushDueToTimeout || peerBatchingQueueMap.get(destinationIp).size() >= forwardingBatchSize;
+        final long currentTime = System.currentTimeMillis();
+        final long millisSinceLastFlush = currentTime - peerBatchingLastFlushTimeMap.getOrDefault(destinationIp, currentTime);
+        return millisSinceLastFlush >= forwardingBatchTimeout.toMillis();
     }
 
     void processFailedRequestsLocally(final AggregatedHttpResponse httpResponse, final Collection<Record<Event>> records) {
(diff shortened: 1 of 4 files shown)


data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
```diff
@@ -178,21 +187,23 @@ private Map<String, List<Record<Event>>> groupRecordsBasedOnIdentificationKeys(
...
-        }
-        if (inetAddress.isAnyLocalAddress() || inetAddress.isLoopbackAddress()) {
-            return true;
-        } else {
+        return localAddressCache.computeIfAbsent(address, addr -> {
+            final InetAddress inetAddress;
             try {
-                return NetworkInterface.getByInetAddress(inetAddress) != null;
-            } catch (final SocketException e) {
+                inetAddress = InetAddress.getByName(addr);
```
dlvenable, line 193: I take it that this call uses Java's DNS cache. Why do you think caching it helps so much?

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
```diff
@@ -134,11 +135,19 @@ public Collection<Record<Event>> forwardRecords(final Collection<Record<Event>>
...
+        final int timeout = expectSubstantialBatch ? 0 : batchDelay;
+        final Map.Entry<Collection<Record<Event>>, CheckpointState> readResult = peerForwarderReceiveBuffer.read(timeout);
 
         final Collection<Record<Event>> records = readResult.getKey();
         final CheckpointState checkpointState = readResult.getValue();
 
+        // Heuristic: expect substantial batches only if we're receiving more than 50% of forwarding batch size.
+        // - When true: use non-blocking reads (0ms) to maximize throughput during high-load periods
+        // - When false: use configured batchDelay to accumulate records and prevent fragmentation
+        final int minBatchSizeForNonBlocking = forwardingBatchSize / 2;
```
dlvenable, line 148: Should we make this percentage configurable in `data-prepper-config.yaml`?

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
```diff
@@ -134,11 +135,19 @@ public Collection<Record<Event>> forwardRecords(final Collection<Record<Event>>
     }
 
     public Collection<Record<Event>> receiveRecords() {
-        final Map.Entry<Collection<Record<Event>>, CheckpointState> readResult = peerForwarderReceiveBuffer.read(batchDelay);
+        // Adaptive timeout: use non-blocking (0ms) or configured delay based on expected batch size
+        final int timeout = expectSubstantialBatch ? 0 : batchDelay;
+        final Map.Entry<Collection<Record<Event>>, CheckpointState> readResult = peerForwarderReceiveBuffer.read(timeout);
```
dlvenable, line 140: If the buffer is already full, does it not immediately return? What case do you think is making this take a long time when there are a lots of requests?

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
```diff
@@ -232,52 +243,38 @@ private List<Record<Event>> populateBatchingQueue(final String destinationIp, fi
...
+            final List<Record<Event>> currentBatch = recordsToForward;
             try {
                 final CompletableFuture<AggregatedHttpResponse> responseFuture =
-                        peerForwarderClient.serializeRecordsAndSendHttpRequest(recordsToForward, destinationIp, pluginId, pipelineName);
-                forwardingRequestsMap.put(responseFuture, recordsToForward);
-                for (Record<Event> record: recordsToForward) {
+                        peerForwarderClient.serializeRecordsAndSendHttpRequest(currentBatch, destinationIp, pluginId, pipelineName);
+
+                // Process response asynchronously without blocking
+                responseFuture
```
dlvenable, line 260: Do you think there is any issue with this possibly making _too many_ requests? Is there something in Armeria or `CompletableFuture` that will limit the total requests?

dlvenable (commented):
Thanks @KarstenSchnitter for these improvements! I'm excited to see the results of this. A few high-level questions:

* Do you have benchmark results you can share from before the changes and after?
* Have you tested this with real Data Prepper nodes? If so, what are your findings.

Also, we should probably have some unit tests for these behaviors.

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
KarstenSchnitter: This made a tremendous improvement with the JMH tests. As I understand, this code is to check, whether the own IP address is used. Caching it help especially with small batch sizes, where this code gets called more often.

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
KarstenSchnitter: I thought about this. The reasoning here is, that it is much more important to have this heuristic at all then where the actual limit is. It is a little hard to explain and tune this value if it were exposed. But there is nothing going against exposing it.

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
KarstenSchnitter: This is about returning early when the buffer is not entirely full but full enough. The receive buffer is always alternating with forwarding data. Waiting for the buffer full or timeout creates a hard limit on the throughput by the inverse batchDelay the batchSize and the number of workers.

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/peerforwarder/RemotePeerForwarder.java
KarstenSchnitter: The number of requests is limited by the client thread count in the peer forwarder client. This config allows to control how many requests are made. I think there is no need for extra synchronisation here.

KarstenSchnitter:
@dlvenable: I can provide the JMH results before and after the code changes. Would that help to understand the impact. They would be depending on my machine but you could see the relative improvements.

KarstenSchnitter:
I ran the JMH benchmarks introduced with the first commit in this PR and compared with the benchmark results with the currently last commit. The result are down below. Here is a performance comparison:


| nodes | records | baseline (ops/s) | improved (ops/s) | delta  |                                                                                                                                                                                                                      
|-------|---------|-----------------|------------------|-------|
|         1 |        100 |   12,012 ± 2,397 |     10,619 ± 1,732 |  _-12%_ |                                                                                                                                                                                                                      
|         1 |     1,000 |           1,610 ± 81 |           1,790 ± 47 |  _+11%_ |                                                                                                                                                                                                                      
|         1 |    5,000 |             359 ± 29 |              307 ± 14 |  _-14%_ |                                                                                                                                                                                                                      
|         1 |  50,000 |           28.3 ± 1.3 |             26.9 ± 1.2 |    _-5%_ |
|         2 |       100 |         1,977 ± 105 |    10,642 ± 2,720 |+438%|
|         2 |    1,000 |             914 ± 30 |           1,444 ± 85 | +58% |
|         2 |    5,000 |               270 ± 7 |              264 ± 31 |    _-2%_ │
|         2 | 50,000 |            16.3 ± 1.7 |              21.7 ± 1.0 | +33% |
|         4 |       100 |             736 ± 34 |      9,224 ± 1,747 |+1153%|
|         4 |    1,000 |           499 ± 126 |         1,237 ± 145 |+148% |
|         4 |   5,000 |             206 ± 21 |              260 ± 15 |  +26% |
|         4 | 50,000 |           15.2 ± 0.9 |             21.4 ± 1.1 |   +41% |
 
The large improvements for small record numbers come from the first commit in particular the cached DNS lookups. The second and third commit improve the actual forwarding loop in multi-node setups and skip the read delay on full pipelines (high throughput). Changes below 15% are statistically insignificant.

Original:

```
Benchmark                                             (nodeCount)  (recordCount)   Mode  Cnt      Score      Error  Units
RemotePeerForwarderBenchmark.benchmarkForwardRecords            1            100  thrpt    5  12011,864 ± 2397,057  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            1           1000  thrpt    5   1610,144 ±   81,295  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            1           5000  thrpt    5    358,806 ±   29,385  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            1          50000  thrpt    5     28,311 ±    1,268  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            2            100  thrpt    5   1977,318 ±  105,486  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            2           1000  thrpt    5    914,486 ±   30,344  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            2           5000  thrpt    5    269,807 ±    7,452  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            2          50000  thrpt    5     16,301 ±    1,672  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            4            100  thrpt    5    735,829 ±   33,802  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            4           1000  thrpt    5    499,197 ±  125,630  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            4           5000  thrpt    5    205,993 ±   20,602  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords            4          50000  thrpt    5     15,214 ±    0,923  ops/s
```

Improved:

```
Benchmark                                             (networkLatencyMs)  (nodeCount)  (recordCount)   Mode  Cnt      Score      Error  Units
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            1            100  thrpt    5  10619,076 ± 1731,733  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            1           1000  thrpt    5   1789,886 ±   47,027  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            1           5000  thrpt    5    307,253 ±   13,556  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            1          50000  thrpt    5     26,943 ±    1,184  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            2            100  thrpt    5  10641,924 ± 2720,015  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            2           1000  thrpt    5   1443,829 ±   85,013  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            2           5000  thrpt    5    264,101 ±   31,455  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            2          50000  thrpt    5     21,738 ±    0,998  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            4            100  thrpt    5   9224,068 ± 1747,447  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            4           1000  thrpt    5   1236,758 ±  144,839  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            4           5000  thrpt    5    260,311 ±   14,853  ops/s
RemotePeerForwarderBenchmark.benchmarkForwardRecords                   0            4          50000  thrpt    5     21,399 ±    1,136  ops/s
```

@dlvenable let me know your take on this.

PR #6415, opened 2026-01-15
Adding functionality to read from specific timestamps for KDS source

### LINKED ISSUE #6366: Add support for reading from specific timestamps in Kinesis source plugin
Currently, the Kinesis source plugin supports initial_position with values `LATEST` and `TRIM_HORIZON`, but does not support the `AT_TIMESTAMP` option. Users cannot specify a point in time from which to start reading Kinesis stream data.

This limitation makes it difficult to:

- Replay data from a specific point in time
- Resume processing from a known timestamp after an outage
- Process only recent data (e.g., last 24 hours) without reading from the beginning of the stream

Add support for `AT_TIMESTAMP` initial position with two configuration options:

Relative duration (lookback) - Specify how far back from the current time to start reading
Absolute timestamp - Specify an exact timestamp to start reading from
Configuration Examples

**Relative Duration (Lookback)**
```
kinesis-pipeline:
  source:
    kinesis:
      streams:
        - stream_name: my-stream
          initial_position: AT_TIMESTAMP
          initial_position_lookback: "P3DT12H"  # Start from 3 days 12 hours ago
```
**Absolute Timestamp**

```
kinesis-pipeline:
  source:
    kinesis:
      streams:
        - stream_name: my-stream
          initial_position: AT_TIMESTAMP
          initial_position_timestamp: "2024-01-15T10:30:00Z"
```

### AWS KCL Reference

This feature leverages the AWS Kinesis Client Library (KCL) `InitialPositionInStreamExtended` class:

```java
InitialPositionInStreamExtended.newInitialPositionAtTimestamp(Date timestamp)
```

**Related AWS Documentation**

ShardFilter `AT_TIMESTAMP` - Returns all shards whose start timestamp is less than or equal to the given timestamp and end timestamp is greater than or equal to the given timestamp or still open.
**Note on Checkpointing Behavior**

The initial_position settings (including `AT_TIMESTAMP`) only apply when:
- The lease table is empty (first run)
- New shards are discovered

Once checkpoints exist in the lease table, the consumer resumes from the checkpointed position.

divbok:
### Description
Adds support for starting Kinesis stream reads from a specific timestamp, configurable via relative duration (lookback) or absolute timestamp. Users can specify the timestamps in two ways:

1. Relative duration(lookback) - Specify how far back from the current time to start reading

```
kinesis-pipeline:
  source:
    kinesis:
      streams:
        - stream_name: my-stream
          initial_position: AT_TIMESTAMP
          lookback_duration: "P3DT12H"  # Start from 3 days 12 hours ago
```
2. Absolute timestamp - Specify an exact timestamp to start reading from
```
kinesis-pipeline:
  source:
    kinesis:
      streams:
        - stream_name: my-stream
          initial_position: AT_TIMESTAMP
          initial_timestamp: "2024-01-15T10:30:00Z" #Specified in UTC timezone format
```
 
### Issues Resolved
Resolves #6366 
 
### Check List

### DIFF
--- a/data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/configuration/KinesisStreamConfig.java
+++ b/data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/configuration/KinesisStreamConfig.java
@@ -11,6 +11,8 @@
 package org.opensearch.dataprepper.plugins.kinesis.source.configuration;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.datatype.jsr310.deser.LocalDateTimeDeserializer;
 import jakarta.validation.Valid;
 import lombok.Getter;
 import org.opensearch.dataprepper.plugins.codec.CompressionOption;
@@ -18,6 +20,7 @@ import software.amazon.awssdk.arns.Arn;
 import software.amazon.kinesis.common.InitialPositionInStream;
 
 import java.time.Duration;
+import java.time.LocalDateTime;
 import java.util.Objects;
 
 @Getter
@@ -43,6 +46,13 @@ public class KinesisStreamConfig {
 
     @JsonProperty("checkpoint_interval")
     private Duration checkPointInterval = MINIMAL_CHECKPOINT_INTERVAL;
+
+    @JsonProperty("range")
+    private Duration range;
+
+    @JsonDeserialize(using = LocalDateTimeDeserializer.class)
+    @JsonProperty("initial_timestamp")
+    private LocalDateTime initialTimestamp;
 
     public InitialPositionInStream getInitialPosition() {
         return initialPosition.getPositionInStream();
--- a/data-prepper-plugins/kinesis-source/src/test/java/org/opensearch/dataprepper/plugins/kinesis/source/KinesisMultiStreamTrackerTest.java
+++ b/data-prepper-plugins/kinesis-source/src/test/java/org/opensearch/dataprepper/plugins/kinesis/source/KinesisMultiStreamTrackerTest.java
@@ -28,7 +28,10 @@ import software.amazon.kinesis.processor.FormerStreamsLeasesDeletionStrategy;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -37,6 +40,7 @@ import java.util.UUID;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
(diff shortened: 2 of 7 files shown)


data-prepper-plugins/kinesis-source/src/test/java/org/opensearch/dataprepper/plugins/kinesis/source/configuration/KinesisSourceConfigTest.java
```diff
@@ -234,4 +236,43 @@ void testSourceConfigWithStreamArnConsumerArn() {
...
+
+        assertEquals(streamConfigs.size(), 2);
+
+        for (KinesisStreamConfig kinesisStreamConfig: streamConfigs) {
+            assertTrue(kinesisStreamConfig.getName().contains("stream"));
+            assertNull(kinesisStreamConfig.getStreamArn());
+            assertNull(kinesisStreamConfig.getConsumerArn());
+            assertEquals(kinesisStreamConfig.getInitialPosition(), InitialPositionInStream.AT_TIMESTAMP);
+
+            if ("stream-1".equals(kinesisStreamConfig.getName())) {
```
dlvenable: It's better to avoid conditionals in tests.

I'd recommend that you create a map using Java streams.

Then, use assertions to make sure the map has the expected objects with expected values. This will avoid both a loop and conditions.

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/configuration/KinesisStreamConfig.java
```diff
@@ -44,6 +48,13 @@ public class KinesisStreamConfig {
     @JsonProperty("checkpoint_interval")
     private Duration checkPointInterval = MINIMAL_CHECKPOINT_INTERVAL;
 
+    @JsonProperty("lookback_duration")
```
dlvenable: Is "lookback duration" a Kinesis term? If not, we may want to consider terms that we already using in Data Prepper. I think the S3 source currently uses `range` for this.

data-prepper-plugins/kinesis-source/src/test/java/org/opensearch/dataprepper/plugins/kinesis/source/KinesisMultiStreamTrackerTest.java
```diff
@@ -28,15 +28,16 @@
...
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.*;
```
sb2k16: we should use explicit imports.

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/configuration/KinesisStreamConfig.java
```diff
@@ -44,6 +47,13 @@ public class KinesisStreamConfig {
...
+    private Duration lookbackDuration;
+
+    @JsonDeserialize(using = LocalDateTimeDeserializer.class)
+    @JsonProperty("initial_timestamp")
```
sb2k16, line 54: we should have a validation mechanism to ensure that either of these 2 attributes are defined in the pipeline yaml.

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/configuration/KinesisStreamConfig.java
divbok: No, its not a kinesis term. Doesn't `lookback_duration` make more sense? I can change it to `range` too to keep it consistent across Data prepper

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/configuration/KinesisStreamConfig.java
divbok: I have provided this in `KinesisMultiStreamTracker` class at Line 92

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/configuration/KinesisStreamConfig.java
nsa-softpay: I think `initial_offset` makes sense here because it's consistent with `initial_position`/`initial_timestamp` and it makes it clear that the "offset" is only used *initiallly* and not if there is already a kinesis checkpoint available in the dynamodb coordination table.

dlvenable: Use of `initial_` makes sense. But, I think Kinesis doesn't use the term offset. Instead they are sequence numbers.

PR #6417, opened 2026-01-16
Remove SingleThread annotation from Drop Events Processor

kkondaka:
### Description
Remove SingleThread annotation from Drop Events Processor. There is no state maintained in the processor to use SingleThread annotation. Multiple worker threads should be able to use one processor object.
 
### Issues Resolved
 
### Check List


data-prepper-plugins/drop-events-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/drop/DropEventsProcessor.java
```diff
@@ -24,7 +23,6 @@
 import java.util.Collections;
 import java.util.stream.Collectors;
 
-@SingleThread
```
graytaylor0, line 27: Is there any context for why this was added in the first place?

dlvenable: We originally added this annotation because the expression evaluator was not thread safe. That was resolved in #1840. So this processor doesn't need to have per-thread implementations.

PR #6424, opened 2026-01-23
Flush remaining data to S3 during shutdown

Subrahmanyam-Gollapalli:
### Description
Add shutdown method to ensure all remaining buffered data is flushed to S3 before the sink shuts down
 
### Issues Resolved
Prevents data loss when the sink is stopped.
 
### Check List

### DIFF
--- a/data-prepper-plugins/s3-sink/src/test/java/org/opensearch/dataprepper/plugins/sink/s3/S3SinkServiceTest.java
+++ b/data-prepper-plugins/s3-sink/src/test/java/org/opensearch/dataprepper/plugins/sink/s3/S3SinkServiceTest.java
@@ -665,6 +665,88 @@ class S3SinkServiceTest {
         verify(s3ObjectsForceFlushedCounter, times(2)).increment();
     }
 
+    @Test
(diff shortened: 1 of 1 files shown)


data-prepper-plugins/s3-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/s3/S3SinkService.java
```diff
@@ -248,4 +248,44 @@ private void checkAggregateThresholdsAndFlushIfNeeded(final List<CompletableFutu
             }
         }
     }
+
+    /**
+     * Force flush all remaining S3 groups to S3. This is called during shutdown
+     * to ensure all buffered data is uploaded before the sink shuts down.
+     */
+    void flushAllRemainingGroups() {
```
dlvenable, line 256: Please add unit tests on this to ensure that we do correctly flush all the groups.

graytaylor0: Thanks for making this change! The change looks good but we should have some unit tests

PR #6432, opened 2026-01-26
feat: Add configurable lease interval for crawler source

chrisale000:
## Description

This PR adds support for configurable lease interval in the crawler source plugin, allowing users to customize the leader scheduler's lease interval instead of using a hardcoded value.

## Changes
- Added `getLeaseInterval()` method to `CrawlerSourceConfig` interface with default value of 1 minute
- Modified `CrawlerSourcePlugin` to use the configurable lease interval from the source configuration

## Testing
- [x] Unit tests pass
- [x] Confirmed M365 maintained 1 minute lease interval. 

Signed-off-by: Alexander Christensen <alchrisk@amazon.com>

### DIFF
--- a/data-prepper-plugins/saas-source-plugins/source-crawler/src/main/java/org/opensearch/dataprepper/plugins/source/source_crawler/base/CrawlerSourceConfig.java
+++ b/data-prepper-plugins/saas-source-plugins/source-crawler/src/main/java/org/opensearch/dataprepper/plugins/source/source_crawler/base/CrawlerSourceConfig.java
@@ -45,4 +45,14 @@ public interface CrawlerSourceConfig {
     default Duration getDurationToDelayRetry() {
         return DEFAULT_MAX_DURATION_TO_DELAY_RETRY;
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/saas-source-plugins/source-crawler/src/main/java/org/opensearch/dataprepper/plugins/source/source_crawler/base/CrawlerSourceConfig.java
```diff
@@ -45,4 +45,14 @@ default Duration getDurationToGiveUpRetry() {
...
     }
+
+    /**
+     * Gets the lease interval for the leader scheduler.
+     * Defaults to 1 minute if not overridden.
+     *
+     * @return Duration for lease interval
+     */
+    default Duration getLeaseInterval() {
+        return Duration.ofMinutes(1);
```
graytaylor0, line 56: This default is a little low. The default is normally 10 minutes at this time I think.

PR #6443, opened 2026-01-29
Fix delete_source removing parsed field when writing to root in parse…

DayneD89:
… JSON processor

When source field name matches a key in the parsed JSON and delete_source is true with destination set to root (the default), the source was deleted after writeToRoot(), removing the parsed value instead of the original. Now deletes source before writeToRoot() when writing to root.

### Description
Fixes an oversight in the parse_json processor where if processing a json with a field value matching `source` field and `delete_source` set to true causes the added field to be removed.
 
### Issues Resolved
I haven't opened a ticket, the contribution guide says 'You open an issue to discuss any significant work - we would hate for your time to be wasted.' and I don't consider this a significant change?
 
### Check List


dlvenable (approved):
Thank you @DayneD89 for finding this issue and fixing it!

DayneD89:
> Thank you @DayneD89 for finding this issue and fixing it!

No problem, tbf it's because I need it fixed so it's selfishness really. What's your process for bringing this in? Do I need to do anything or leave it with you and wait for a release with it in?

PR #6444, opened 2026-01-29
refactor(source-crawler): enable direct injection of VendorAPIMetricsRecorder

chrisale000:
### Description
- Add `@Named` and `@Inject` annotations for direct component scanning
- Office365 uses `enableSubscriptionMetrics()` at earliest initialization stage
- Maintain full backward compatibility with existing injection.
- Remove Rest Client Logging as this logging is covered in the crawler and the request can be recreated.

 ### Testing: Verified both patterns generate metrics successfully
- Spun Up Multiple pipelines locally and confirmed: 
   - Direct injection via @Named/@Inject (subscription metrics disabled by default) ✓
   - Office365: Early enablement in Source constructor ensures metrics available throughout lifecycle ✓
   - Comprehensive test coverage for enablement functionality and constructor behavior ✓
   - Backwards compatibility ✓

### Issues Resolved
N/A
 
### Check List


data-prepper-plugins/saas-source-plugins/microsoft-office365-source/src/main/java/org/opensearch/dataprepper/plugins/source/microsoft_office365/auth/Office365AuthenticationProvider.java
```diff
@@ -21,18 +21,18 @@
...
+import javax.inject.Named;
 import java.time.Instant;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * OAuth2 implementation of the Office365AuthProvider.
  */
-@Component
+@Named
```
wjyao0316, line 35: Can you elaborate why changing the annotation here is necessary?

chrisale000: The reason we are moving from `@Component` to `@Named` is to enable direct dependency injection without requiring explicit configuration classes. `@Component` is part of Spring's dependency injection framework and requires components to be explicitly registered in a configuration class (like the ones being removed in this refactor). In contrast, `@Named` is part of the Java CDI (Contexts and Dependency Injection) standard from the `javax.inject` package, which supports automatic component scanning and injection without needing configuration classes.

PR #6479, opened 2026-02-02
Add otel-apm-service-map processor

### LINKED ISSUE #6482: Add a processor for application performance monitoring service map
Need to OTEL trace spans and generate Application Performance Monitoring (APM) service map relationships and metrics.

The `otel_apm_service_map` processor analyzes OpenTelemetry trace spans to automatically generate Application Performance Monitoring (APM) service map relationships and metrics. It creates structured events that can be visualized as service topology graphs, showing how services communicate with each other and their performance characteristics.

**Additional context**

kkondaka:
### Description
The `otel_apm_service_map` processor analyzes OpenTelemetry trace spans to automatically generate Application Performance Monitoring (APM) service map relationships and metrics. It creates structured events that can be visualized as service topology graphs, showing how services communicate with each other and their performance characteristics.
 
### Issues Resolved
Resolves #6482 
 
### Check List


data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/OtelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+import java.util.TreeMap;
+import java.util.concurrent.BrokenBarrierException;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+
+@SingleThread
+@DataPrepperPlugin(name = "otel_apm_service_map", pluginType = Processor.class,
+        pluginConfigurationType = OtelApmServiceMapProcessorConfig.class)
+public class OtelApmServiceMapProcessor extends AbstractProcessor<Record<Event>, Record<Event>> implements RequiresPeerForwarding {
```
dlvenable: Let's rename all the classes that start with `Otel` to `OTel` to remain consistent with other class names in the project.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/utils/ApmServiceMapMetricsUtil.java
```diff
@@ -0,0 +1,379 @@
...
+     * Create histogram buckets from raw duration values
+     * Uses O-Tel Java SDK bucket: 0.0 ms to 10 sec
+     * https://opentelemetry.io/docs/specs/otel/metrics/sdk/?utm_source=chatgpt.com#explicit-bucket-histogram-aggregation
+     *
+     * @param durations List of duration values in seconds
+     * @return HistogramBuckets with counts and bounds
+     */
+    public static HistogramBuckets createHistogramBucketsFromDurations(final List<Double> durations) {
+        // Standard latency buckets in seconds
+        final List<Double> explicitBounds = Arrays.asList(0.0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1,
```
dlvenable: You should be able to make this a static list so that you don't need to create a new `List` each call. Also, use `List.of(...)`.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/utils/ApmServiceMapMetricsUtil.java
```diff
@@ -0,0 +1,379 @@
...
+
+    /**
+     * Create histogram buckets from raw duration values
+     * Uses O-Tel Java SDK bucket: 0.0 ms to 10 sec
+     * https://opentelemetry.io/docs/specs/otel/metrics/sdk/?utm_source=chatgpt.com#explicit-bucket-histogram-aggregation
+     *
+     * @param durations List of duration values in seconds
+     * @return HistogramBuckets with counts and bounds
+     */
+    public static HistogramBuckets createHistogramBucketsFromDurations(final List<Double> durations) {
```
dlvenable: You can make this `private`.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/utils/ApmServiceMapMetricsUtil.java
```diff
@@ -0,0 +1,379 @@
...
+     * Create a JacksonStandardHistogram metric from collected latency durations
+     *
+     * @param metricName Name of the metric
+     * @param description Description of the metric
+     * @param durations List of duration values in seconds
+     * @param labels Labels for the metric
+     * @param timestamp Timestamp for the metric
+     * @return JacksonStandardHistogram metric event
+     */
+    public static JacksonMetric createJacksonStandardHistogram(final String metricName,
```
dlvenable: You can make this private.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtil.java
```diff
@@ -0,0 +1,379 @@
...
+                                                              final Instant timestamp) {
+        final long timestampNanos = getTimeNanos(timestamp);
+        final long startTimeNanos = timestampNanos; // For histogram metrics, start time can be same as timestamp
+
+        // Create histogram buckets from raw duration values
+        final HistogramBuckets buckets = createHistogramBucketsFromDurations(durations);
+
+        final Map<String, Object> labelsWithRandomKey = new HashMap<>();
+        labelsWithRandomKey.putAll(labels);
+        labelsWithRandomKey.put("randomKey", UUID.randomUUID().toString());
```
dlvenable, line 316: What is `randomKey`?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/utils/ApmServiceMapMetricsUtil.java
```diff
@@ -0,0 +1,379 @@
...
+     *
+     * @param metricName Name of the metric
+     * @param description Description of the metric
+     * @param value Value of the metric
+     * @param labels Labels for the metric
+     * @param timestamp Timestamp for the metric
+     * @param exemplars List of exemplars for the metric
+     * @return JacksonSum metric event
+     */
+    public static JacksonMetric createJacksonSumMetric(final String metricName,
```
dlvenable: Make this private.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/OtelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+
+                    final Instant connectionAnchorTimestamp = getAnchorTimestampFromSpan(clientSpan, currentTime);
+
+                    final ServiceConnection serviceConnection = new ServiceConnection(
+                            clientService,
+                            serverService,
+                            connectionAnchorTimestamp
+                    );
+
+                    final Event connectionEvent = JacksonEvent.builder()
```
dlvenable: This plugin should get the `EventFactory` in the constructor and use it to create this event.

Example usage:

https://github.com/opensearch-project/data-prepper/blob/38b8db5159f34a76e93b0fddfa24eaf2b549ab6f/data-prepper-core/src/main/java/org/opensearch/dataprepper/core/pipeline/router/RouterCopyRecordStrategy.java#L100

I see that you also create `JacksonSum` and a few others that don't have support in the event factory yet. That is fine to keep as is. But, for a normal `JacksonEvent` like this one, we should use the factory.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/model/internal/ThreeWindowTraceData.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Data structure to hold three-window trace processing data
+ */
+public class ThreeWindowTraceData {
+    public final Collection<SpanStateData> processingSpans;
```
dlvenable: Make this `private` and create getter methods.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/model/internal/ClientSpanDecoration.java
```diff
@@ -0,0 +1,39 @@
...
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * Decoration for CLIENT spans containing pre-computed relationship data
+ * (groupByAttributes are read directly from SpanStateData to avoid duplication)
+ */
+public class ClientSpanDecoration implements Serializable {
+    public final String parentServerOperationName;
```
dlvenable: All of these classes expose fields. Make them private and provide getter methods.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/model/ServiceConnection.java
```diff
@@ -0,0 +1,93 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.util.Objects;
+
+/**
+ * Represents the connection between two services.
+ */
+public class ServiceConnection {
```
dlvenable, line 22: Who defines this model? Is it related to an OpenSearch index?

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/model/ServiceTest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Test
+    void testEquals() {
+        Service.KeyAttributes keyAttributes = new Service.KeyAttributes("prod", "test-service");
+        Service service1 = new Service(keyAttributes);
+        Service service2 = new Service(keyAttributes);
+
+        assertEquals(service1, service2);
+        assertNotEquals(service1, null);
```
dlvenable, line 57: Split different test cases into different tests.

Also, we should have a test that `equals` works with unequal objects.

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/model/ServiceTest.java
```diff
@@ -0,0 +1,115 @@
...
+            assertEquals("test-service", keyAttributes.getName());
+        }
+
+        @Test
+        void testEquals() {
+            Service.KeyAttributes keyAttributes1 = new Service.KeyAttributes("prod", "test-service");
+            Service.KeyAttributes keyAttributes2 = new Service.KeyAttributes("prod", "test-service");
+
+            assertEquals(keyAttributes1, keyAttributes2);
+            assertNotEquals(keyAttributes1, new Service.KeyAttributes("dev", "test-service"));
```
dlvenable, line 106: Similar comments as above. Create different `@Test` test cases for different test cases.

data-prepper-plugins/otel-apm-service-map-processor/README.md
```diff
@@ -0,0 +1,325 @@
...
+- **Span relationships**: Uses parent-child relationships to build complete call chains
+
+## Configuration
+
+### Basic Configuration
+
+```yaml
+processor:
+  - otel_apm_service_map:
+      window_duration: 60
```
dlvenable: This is a Data Prepper duration so we should update this example.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+@SingleThread
+@DataPrepperPlugin(name = "otel_apm_service_map", pluginType = Processor.class,
+        pluginConfigurationType = OtelApmServiceMapProcessorConfig.class)
+public class OtelApmServiceMapProcessor extends AbstractProcessor<Record<Event>, Record<Event>> implements RequiresPeerForwarding {
+
+    private static final String SPANS_DB_SIZE = "spansDbSize";
+    private static final String SPANS_DB_COUNT = "spansDbCount";
+
+    private static final Logger LOG = LoggerFactory.getLogger(OtelApmServiceMapProcessor.class);
+    private static final String EVENT_TYPE_OTEL_APM_SERVICE_MAP = "OTelAPMServiceMap";
```
dlvenable: It seems that we don't have a convention on values for these types. Some are all upper case and some are all lowercase. Now we have a mixed camel case. I also think we want to have a smaller set of event types. The goal of this should be to promote easy filtering downstream. Can this just be either `SERVICE_MAP` or `service_map`?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+                apmEvents.addAll(generateServiceConnectionsFromEphemeralDecorations(traceData, currentTime, metricsStateByKey));
+                apmEvents.addAll(generateServiceOperationDetailsFromEphemeralDecorations(traceData, currentTime, metricsStateByKey));
+            }
+        }
+
+        final List<JacksonMetric> metrics = ApmServiceMapMetricsUtil.createMetricsFromAggregatedState(metricsStateByKey);
+        metrics.sort(Comparator.comparing(JacksonMetric::getTime));
+
+        final List<Record<Event>> apmEventsSorted = new ArrayList<>();
+        apmEventsSorted.addAll(metrics.stream().map(metric -> new Record<Event>(metric)).collect(Collectors.toList()));
```
dlvenable, line 402: This results in sending out both service map data and metric data from the same source. Is this how we want this processor to operate? For a simple pipeline that means that one OpenSearch index gets both.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+        apmEventsSorted.addAll(apmEvents);
+
+        return apmEventsSorted;
```
dlvenable, line 405: Do we not send out the raw trace information? Is the expectation to have multiple pipelines similar to the existing trace pipeline? The README below indicates that there is just one pipeline.

data-prepper-plugins/otel-apm-service-map-processor/README.md
```diff
@@ -0,0 +1,325 @@
...
+- **Window size impact**: Larger `window_duration` values require more storage
+- **Trace volume**: Memory usage scales with the number of concurrent traces
+
+### Storage Requirements
+
+- **Database path**: Ensure sufficient disk space at the configured `db_path`
+- **Cleanup**: Old database files are automatically cleaned up during window rotation
+- **I/O performance**: Use fast storage (SSD) for better performance
+
+### Scaling Guidelines
```
dlvenable: Are these actual numbers? Should we just remove this section?

data-prepper-plugins/otel-apm-service-map-processor/README.md
```diff
@@ -0,0 +1,325 @@
...
+|--------------|-------------------|
+| < 10k spans/sec | 2-4 GB heap |
+| 10k-50k spans/sec | 4-8 GB heap |
+| > 50k spans/sec | 8+ GB heap |
+
+## Troubleshooting
+
+### Common Issues
+
+#### High Memory Usage
```
dlvenable: These look like random AI generated notes. Do we really need this here?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/OtelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ */
+
+package org.opensearch.dataprepper.plugins.processor;
```
dlvenable: This processor should be moved into its own package. This will enable easier controls over logging in Log4J configurations.

```
org.opensearch.dataprepper.plugins.processor.otel_apm_service_map
```

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/OtelApmServiceMapProcessorTest.java
```diff
@@ -0,0 +1,1096 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.lenient;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class OtelApmServiceMapProcessorTest {
```
dlvenable: Also include a plugin framework test. This will give automatic tests. And you can add a test against actual configuration files.

Here is a small example:

https://github.com/opensearch-project/data-prepper/blob/666c7269d87df7849d4f3d761c95a3507c7a0239/data-prepper-plugins/drop-events-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/drop/DropEventsProcessorIT.java#L33-L34

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtil.java
kkondaka: It seems "randomKey" is to differentiate two different metrics with same labels but different data points

data-prepper-plugins/otel-apm-service-map-processor/README.md
```diff
@@ -0,0 +1,325 @@
+# OpenTelemetry APM Service Map Processor
+
+## Overview
+
+The `otel_apm_service_map` processor analyzes OpenTelemetry trace spans to automatically generate Application Performance Monitoring (APM) service map relationships and metrics. It creates structured events that can be visualized as service topology graphs, showing how services communicate with each other and their performance characteristics.
```
KarstenSchnitter, line 5: Can you add a link to documentation on the APM service maps? It does not become clear, what kind of data structure is created.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+
+            final Map<String, Object> attributes = span.getAttributes();
+            if (attributes != null) {
+                combinedAttributes.putAll(attributes);
+            }
+
+            final Map<String, Object> resource = span.getResource();
+            if (resource != null) {
+                combinedAttributes.put("resource", resource);
+            }
```
KarstenSchnitter, line 320: Doesn't this ignore instrumentation scope attributes?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/OtelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+            final Map<String, String> groupByAttrs = extractGroupByAttributes(span);
+            final Map<String, Object> spanAttributes = extractSpanAttributes(span);
+
+            try {
+                final byte[] traceId = Hex.decodeHex(span.getTraceId());
+                final SpanStateData spanStateData = new SpanStateData(
+                        serviceName,
+                        Hex.decodeHex(spanId),
+                        parentSpanId.isEmpty() ? null : Hex.decodeHex(parentSpanId),
+                        traceId,
```
KarstenSchnitter: [nit-pick] Somewhat asymmetric handling of span and trace ids. You could hex-decode the span ids earlier as well.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+     * Adds the data for spans from the ResourceSpans object to the current window
+     *
+     * @param records Input records that will be modified/processed
+     * @return If the window is reached, returns a list of ServiceDetails and ServiceRemoteDetails events.
+     * Otherwise, returns an empty set.
+     */
+    @Override
+    public Collection<Record<Event>> doExecute(Collection<Record<Event>> records) {
+        final Collection<Record<Event>> apmEvents = windowDurationHasPassed() ? evaluateApmEvents() : EMPTY_COLLECTION;
+        final Map<byte[], Collection<SpanStateData>> batchStateData = new TreeMap<>(SignedBytes.lexicographicalComparator());
```
KarstenSchnitter, line 165: Using `byte[]` as key does actually use the reference of the byte array not its value. This might be wrong here, when different byte arrays contain the same trace id. Try using the String or a ByteBuffer instead.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+        return currentWindow.size() == 0;
+    }
+
+    @Override
+    public void shutdown() {
+        previousWindow.delete();
+        currentWindow.delete();
+        if (nextWindow != null) {
+            nextWindow.delete();
+        }
```
KarstenSchnitter, line 203: This might create a resource leak, if for example `previousWindow.delete()` throws an exception.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(OtelApmServiceMapProcessor.class);
+    private static final String EVENT_TYPE_OTEL_APM_SERVICE_MAP = "OTelAPMServiceMap";
+    private static final Collection<Record<Event>> EMPTY_COLLECTION = Collections.emptySet();
+    private static final String SPAN_KIND_SERVER = "SPAN_KIND_SERVER";
+    private static final String SPAN_KIND_CLIENT = "SPAN_KIND_CLIENT";
+
+    // TODO: This should not be tracked in this class, move it up to the creator
+    private static final AtomicInteger processorsCreated = new AtomicInteger(0);
+    private static Instant previousTimestamp;
```
KarstenSchnitter, line 86: It is hard to see, that this Instant is actually protected against concurrent writes by the isMasterInstance(). Can you consider making this safety more explicit?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+
+    /**
+     * This method checks for master instance and let master instance process the current window and rotate the window.
+     *
+     * @return Set of Record<Event> containing json representation of ServiceConnection and ServiceOperationDetail found
+     */
+    private Collection<Record<Event>> evaluateApmEvents() {
+        LOG.debug("Evaluating APM service map events with three-window semantics");
+        try {
+            allThreadsCyclicBarrier.await();
```
KarstenSchnitter, line 344: I do not know, how the data is distributed amongst the processors, but this synchronization point might create a hard throughput limit waiting for the last processor to receive data after a time-Window closes.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+
+            Collection<Record<Event>> apmEvents = new HashSet<>();
+            if (isMasterInstance()) {
+                apmEvents = processCurrentWindowSpans();
```
KarstenSchnitter, line 348: Isn't this a major bottleneck, where all spans are processed by a single thread only? How does this actually work, when there are multiple instances of Data Prepper running?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+                rotateWindows();
```
KarstenSchnitter, line 349: Would it be possible to do the window rotation and unblock all other threads, while the main thread is processing the current window spans?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/OtelApmServiceMapProcessor.java
```diff
@@ -0,0 +1,868 @@
...
+     */
+    private Collection<Record<Event>> processCurrentWindowSpans() {
+        final Collection<Record<Event>> apmEvents = new HashSet<>();
+        final Instant currentTime = clock.instant();
+
+        final EphemeralSpanDecorations ephemeralDecorations = new EphemeralSpanDecorations();
+
+        final Map<MetricKey, MetricAggregationState> metricsStateByKey = new HashMap<>();
+
+        final Map<byte[], Collection<SpanStateData>> previousSpansByTraceId = buildSpansByTraceIdMap(previousWindow);
```
KarstenSchnitter: This map should already have been calculated. Is there a way to reuse the last result where previousWindow was currentWindow?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: Yes, there will be multiple pipelines with first pipeline sending to opensearch and the second sending to the new service map processor, just like the current one. In theory we may be able to deprecate the existing service_map processor in future.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: We can use conditional routing to send to different opensearch indexes, right?

data-prepper-plugins/otel-apm-service-map-processor/README.md
kkondaka: @dlvenable not sure I understand what you meant here. This the `window_duration` config option for this new processor.

data-prepper-plugins/otel-apm-service-map-processor/README.md
kkondaka: @KarstenSchnitter you are right but the existing code has been doing this for a long time in the service map processor!!! See this - https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/mapdb-processor-state/src/main/java/org/opensearch/dataprepper/plugins/processor/state/MapDbProcessorState.java. @dlvenable what do you think?

kkondaka: @KarstenSchnitter the following code takes care of the comparison, so I think the existing code is fine
```


    private static class SignedByteArraySerializer extends SerializerByteArray {
        @Override
        public int compare(byte[] o1, byte[] o2) {
            return SignedBytes.lexicographicalComparator().compare(o1, o2);
        }
    }
```

kkondaka: But for the new Maps in this new code, I will switch to String.

data-prepper-plugins/otel-apm-service-map-processor/README.md
kkondaka: I will remove.

data-prepper-plugins/otel-apm-service-map-processor/README.md
kkondaka: removed them

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: Added them.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: I don't see how `delete()` can throw exception. But even if it does, this is invoked at the time of processor shutdown which is invoked during DataPrepper shutdown. So, I do not think it's a concern.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: This is how the current service map works too.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: Maybe possible but Service map processor is not needed to be highly efficient, it just needs to be accurate

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/OtelApmServiceMapProcessor.java
kkondaka: Not sure I understand your point. The windows maintained so that we handle the cases where map could not be built because of some slow (out of order) data.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
KarstenSchnitter: Is this relying on some kind of peer forwarding or does it just assume, that the instance eventually sees traces covering all services? Can it happen, that a node is missing in the service map just because no trace containing it was ever seen by the master instance?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtil.java
dlvenable: Is this part of the OTel specification?

data-prepper-plugins/otel-apm-service-map-processor/README.md
dlvenable: This should say:

```
window_duration: 60s
```

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
dlvenable: Yes, we can use conditional routing and that makes sense.

My main concern is that it may not be intuitive and expected. And it means any pipeline using this processor must have some sort of conditional routing.

data-prepper-plugins/otel-apm-service-map-processor/README.md
dlvenable: I'm not sure what motivated the use of `byte[]` in the keys. I don't think we can really output a `byte[]` from Data Prepper very effectively now. Perhaps it was to avoid issues with conversion to string. We need to be sure that the object has a useful `toString`. By default, `toString` is the reference. So serialization to `byte[]` does offer better likelihood of correct hashing. 

What types are being put into the map?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtil.java
```diff
@@ -0,0 +1,379 @@
...
+import static org.opensearch.dataprepper.plugins.otel.codec.OTelProtoCommonUtils.convertUnixNanosToISO8601;
+
+/**
+ * Utility class for handling APM service map metrics generation and processing
+ */
+public final class ApmServiceMapMetricsUtil {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ApmServiceMapMetricsUtil.class);
+    // Standard latency buckets in seconds
+    private static final List<Double> explicitBounds = List.of(0.0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1,
```
dlvenable: Rename to `EXPLICIT_BOUNDS` per Java standards.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: Agreed. But this how ServiceMapStatefulProcessor is also doing. I do not think performance is that important here because the service map is built at an interval of 3+ minutes.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: For now, this is doing exactly what we do in ServiceMapStatefulProcessor. Definitely should revisit in future.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: Fixed.

data-prepper-plugins/otel-apm-service-map-processor/README.md
kkondaka: I think it was primarily traceId which is defined to be byte array. I think it is OK to use byte array as key as long as comparator is overridden. I have modified the code to not use byte array as key in other places. But some places that use existing library functions, still use byte array as key and it is handled correctly.

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/OtelApmServiceMapProcessorTest.java
kkondaka: Done

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/utils/ApmServiceMapMetricsUtil.java
kkondaka: Made it package private.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtil.java
kkondaka: This is put in "attributes" which is part of OTel specification.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/utils/ApmServiceMapMetricsUtil.java
kkondaka: Made it package private.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/utils/ApmServiceMapMetricsUtil.java
kkondaka: Made it package private

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/model/internal/ThreeWindowTraceData.java
kkondaka: Done.

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/model/ServiceTest.java
kkondaka: Done.

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/model/ServiceTest.java
kkondaka: Fixed above but here I donot think it deserves separate test cases. It is just constructing the object and validating the attributes. It is just one API.

data-prepper-plugins/otel-apm-service-map-processor/README.md
kkondaka: Fixed.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: Yes. I have updated README.md to include routing.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/model/ServiceConnection.java
kkondaka: It is defined by APM

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: change to SERVICE_MAP

kkondaka:
@KarstenSchnitter Please review again. I cannot merge while you have "request for changes"

data-prepper-plugin-framework/src/main/java/org/opensearch/dataprepper/plugin/ClasspathPluginProvider.java
```diff
@@ -87,13 +87,18 @@ private Map<String, Map<Class<?>, Class<?>>> scanForPlugins() {
 
         final Map<String, Map<Class<?>, Class<?>>> pluginsMap = new HashMap<>(dataPrepperPluginClasses.size());
         for (final Class<?> concretePluginClass : dataPrepperPluginClasses) {
+            final DataPrepperPlugin annotation = concretePluginClass.getAnnotation(DataPrepperPlugin.class);
```
dlvenable: What are these changes for?

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/utils/OTelSpanDerivationUtil.java
```diff
@@ -8,7 +8,7 @@
  *
  */
 
-package org.opensearch.dataprepper.plugins.otel.common.utils;
+package org.opensearch.dataprepper.plugins.otel.utils;
```
dlvenable: Let's avoid utils. I think just `common` will be fine.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
kkondaka: Our service map processor relies on peer forwarding.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/model/ServiceConnection.java
dlvenable: What do you mean by APM? I don't know of an APM standard or spec. OTel has a spec.

kkondaka: There is no standard spec. It is defined by our APM team to be used by Opensearch UI team. In fact this is changing now. Vamsi has a PR with changes.

data-prepper-plugin-framework/src/main/java/org/opensearch/dataprepper/plugin/ClasspathPluginProvider.java
kkondaka: Removed these changes.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/utils/OTelSpanDerivationUtil.java
kkondaka: Done.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/OtelApmServiceMapProcessor.java
kkondaka: Done.

KarstenSchnitter (approved):

PR #6481, opened 2026-02-03
Add derived server-side attributes to OTel trace raw processor

vamsimanohar:
## Summary

Adds 4 derived attributes to SERVER spans to simplify UI visualization layer queries without complex conditional logic.

### New Derived Attributes

**Added to all SERVER spans:**
- `derived.fault` - "0" or "1" 
- `derived.error` - "0" or "1"
- `derived.operation` - Operation name string
- `derived.environment` - Environment string

### Extraction Logic

**Error/Fault:**
- HTTP 5xx status codes → `fault=1, error=0`
- HTTP 4xx status codes → `fault=0, error=1` 
- Span status ERROR → `fault=1, error=0`
- Otherwise → `fault=0, error=0`

**Operation:**
- Uses HTTP method + first URL path section when available: `"GET /payment"`
- Falls back to span name
- Default: `"UnknownOperation"`

**Environment:**
- Extracts from `resource.attributes.deployment.environment.name`
- Falls back to `resource.attributes.deployment.environment` 
- Default: `"generic:default"`

### Implementation

- `OTelSpanDerivationUtil` in common module for reusability
- Single method call added to `OTelTraceRawProcessor.doExecute()`
- Only processes spans where `kind == "SERVER"`

### Test Coverage

- **Unit tests:** 21 test methods covering all derivation scenarios
- **Processor tests:** Updated to verify derived attributes are added
- **Edge cases:** Null handling, invalid data, missing attributes

### Files Changed

**Added:**
- `data-prepper-plugins/common/src/main/java/.../OTelSpanDerivationUtil.java`
- `data-prepper-plugins/common/src/test/java/.../OTelSpanDerivationUtilTest.java`

**Modified:** 
- `OTelTraceRawProcessor.java` - Added derivation call
- `OTelTraceRawProcessorTest.java` - Updated for derived attributes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,13 @@ data-prepper-main/src/test/resources/logstash-conf/logstash-filter.yaml
 *.iml
 .kiro
 .vscode
+
+# Manual testing files (should not be committed)
+MANUAL_TEST.md
+data-prepper-config.yaml
+pipelines.yaml
+run-dataprepper.sh
+trace-tester.py
+
+# Claude session notes
+claude.md
--- a/.whitesource
+++ b/.whitesource
@@ -26,7 +26,6 @@
     "issueType": "DEPENDENCY"
   },
   "remediateSettings": {
-    "enableRenovate": true,
     "workflowRules": {
       "enabled": true
     }
--- a/data-prepper-plugins/common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/utils/OTelSpanDerivationUtil.java
+++ b/data-prepper-plugins/common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/utils/OTelSpanDerivationUtil.java
@@ -1,0 +1,307 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ */
+
+package org.opensearch.dataprepper.plugins.otel.common.utils;
+
+import org.opensearch.dataprepper.model.trace.Span;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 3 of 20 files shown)


vamsimanohar:
> Thanks for providing this change. I am wondering two things here:
> 
> * Why put this logic in the default OTelTraceRawProcessor and not create a separate processor providing this enrichment?
> * Can the same result be achieved by configuring existing processors?

This processor is created to explicitly enrich the trace analytics UI experience, so we are using the same processor.

I couldn't see a way to do it with existing processors as it involves a quite bit of logic to dervie from the presence of multiple different fields.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/trace/JacksonSpan.java
```diff
@@ -245,6 +245,22 @@ public void setTraceGroupFields(final TraceGroupFields traceGroupFields) {
         }
     }
 
+    @Override
+    public void putAttribute(final String key, final Object value) {
```
kkondaka: I think we should have a common JacksonOTelEvent or something like that for putAttribute(). But for now, do the "putAttribute" in the OtelSpanDerivationUtil.java itself

data-prepper-plugins/common/src/main/java/org/opensearch/dataprepper/plugins/processor/util/OTelSpanDerivationUtil.java
```diff
@@ -0,0 +1,285 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ */
+
+package org.opensearch.dataprepper.plugins.processor.util;
```
kkondaka: Since this is in "common" the package name should not have "processor"

kkondaka (approved):

PR #6486, opened 2026-02-04
Remove default codec and require codec for sqs sink

graytaylor0:
### Description
The SQS sink has default values that are invalid and throw an InvalidPluginConfigurationException. The combination of defaults with codec `ndjson` and `threshold.max_events_per_message` at 25 are invalid together. 

Changing the require codec makes it more clear to users what codec is being used and aligns more closely with other data prepper plugins, and does not throw unexpected validation exceptions from defaults.

 
### Check List

### DIFF
--- a/data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkConfig.java
+++ b/data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkConfig.java
@@ -68,12 +68,5 @@ public class SqsSinkConfig {
              return (groupId == null && deDupId == null);
         }
     }
-
-    @AssertTrue(message = "ndjson codec doesn't support max events per message greater than 1")
-    boolean isValidCodecConfig() {
-        if ((codec != null && codec.getPluginName().equals("ndjson")) && thresholdConfig.getMaxEventsPerMessage() > 1)
-            return false;
-        return true;
-    }
 }
 
--- a/data-prepper-plugins/sqs-sink/src/test/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkConfigTest.java
+++ b/data-prepper-plugins/sqs-sink/src/test/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkConfigTest.java
@@ -114,21 +114,6 @@ public class SqsSinkConfigTest {
         
     }
 
-    @Test
-    void TestValidCodecConfig() throws Exception {
-        reflectivelySetField(sqsSinkConfig, "codec", null);
-        reflectivelySetField(sqsSinkConfig, "thresholdConfig", sqsThresholdConfig);
-        when(sqsThresholdConfig.getMaxEventsPerMessage()).thenReturn(2);
-        assertTrue(sqsSinkConfig.isValidCodecConfig());
-        when(sqsThresholdConfig.getMaxEventsPerMessage()).thenReturn(1);
-        assertTrue(sqsSinkConfig.isValidCodecConfig());
-        PluginModel codec = mock(PluginModel.class);
-        when(codec.getPluginName()).thenReturn("ndjson");
-        reflectivelySetField(sqsSinkConfig, "codec", codec);
-        when(sqsThresholdConfig.getMaxEventsPerMessage()).thenReturn(2);
-        assertFalse(sqsSinkConfig.isValidCodecConfig());
-    }
-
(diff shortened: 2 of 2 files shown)


data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSink.java
```diff
@@ -60,16 +59,12 @@ public SqsSink(final PluginSetting pluginSetting,
...
-            String codecPluginName = codecConfiguration.getPluginName();
-            if (!codecPluginName.equals("json") && !codecPluginName.equals("ndjson")) {
-                throw new RuntimeException(String.format("Codec {} not supported.", codecPluginName));
-            }
-            codecPluginSettings = new PluginSetting(codecConfiguration.getPluginName(),
-                codecConfiguration.getPluginSettings());
-        } else {
-            codecPluginSettings = new PluginSetting("ndjson", Map.of());
+        String codecPluginName = codecConfiguration.getPluginName();
+        if (!codecPluginName.equals("json") && !codecPluginName.equals("ndjson")) {
```
dlvenable, line 63: We should allow for other codecs. I think the only time we want to be restrictive on the codec is when the customer is not using batching (ie. `max_events_per_batch` == 1). In that case, `ndjson` should be the only permissible codec.

Customers may want to serialize in Ion, XML, or other codecs.

data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkConfig.java
```diff
@@ -68,9 +69,9 @@ boolean isValidConfig() {
         }
     }
 
-    @AssertTrue(message = "ndjson codec (default codec) doesn't support max events per message greater than 1")
+    @AssertTrue(message = "ndjson codec doesn't support max events per message greater than 1")
     boolean isValidCodecConfig() {
-        if ((codec == null || codec.getPluginName().equals("ndjson")) && thresholdConfig.getMaxEventsPerMessage() > 1)
+        if ((codec != null && codec.getPluginName().equals("ndjson")) && thresholdConfig.getMaxEventsPerMessage() > 1)
```
dlvenable: I also think it is acceptable to use `ndjson` with more than one event per message.

I think this is the key condition we want to return `false` or invalid for.

```
if ((codec != null && !codec.getPluginName().equals("ndjson")) && thresholdConfig.getMaxEventsPerMessage() == 1)
```

data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSink.java
graytaylor0: It is not clear why the codecs are so limited here

data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkConfig.java
graytaylor0: It is not clear why this limitation was put in place.

dlvenable:
I'm ok to leave these restrictions for now. We can expand them later if desired.

data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkConfig.java
dlvenable: Here is what the SQS sink will do with these combinations:

### ndjson with max_events_per_message == 1

The SQS message looks like:

```
{"fieldA" : "valueA"}
```

### ndjson with max_events_per_message > 1

The SQS message looks like:

```
{"fieldA" : "valueA"}
{"fieldB" : "valueB"}
{"fieldC" : "valueC"}
```

### json with max_events_per_message == 1

The SQS message looks like:

```
{"events" : [{"fieldA" : "valueA"}]
```

### json with max_events_per_message > 1

The SQS message looks like:

```
{"events" : [{"fieldA" : "valueA"}, {"fieldB" : "valueB"}, {"fieldC" : "valueC"}]
```


If a customer sets `max_events_per_message` and uses `json`, they probably don't get the results they expect. They should use `ndjson`.

So I think my initial comment is still valid.

PR #6488, opened 2026-02-04
feat(otlp-sink): Add support for metrics and logs with OTLP encoding

viquer:
Extended OTLP sink to support all three signal types (traces, metrics, logs) with full OTLP protocol encoding/decoding capabilities.

Key Changes:
- Extended OTelProtoCodec interface with convertToResourceMetrics() and convertToResourceLogs() methods for encoding
- Implemented full OTLP encoding in OTelProtoStandardCodec for all metric types (Gauge, Sum, Histogram, Summary, ExponentialHistogram) and logs
- Added OtlpEventEncoder to convert Data Prepper events to OTLP protobuf
- Added OtlpSignalType enum to distinguish between traces, metrics, and logs
- Updated OtlpHttpSender to handle all three signal types with proper partial success handling for each type
- Updated OtlpSinkBuffer to batch and flush events by signal type
- Renamed metrics from span-specific to generic record names:
  - incrementRejectedSpansCount -> incrementRejectedRecordsCount
  - incrementFailedSpansCount -> incrementFailedRecordsCount
- Added comprehensive round-trip tests validating encode/decode fidelity
- Added multi-signal integration tests

This enables OTLP -> modify attributes -> OTLP pipelines for all signal types with full data fidelity preservation.

### Check List


data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/buffer/OtlpSinkBuffer.java
```diff
@@ -120,51 +126,67 @@ void restartWorker() {
...
                 final long now = System.currentTimeMillis();
-                final Record<Span> record = queue.poll(100, TimeUnit.MILLISECONDS);
+                final Record<Event> record = queue.poll(100, TimeUnit.MILLISECONDS);
 
                 if (record != null) {
+                    final Event event = record.getData();
+                    final OtlpSignalType signalType = OtlpSignalType.fromEvent(event);
+
+                    // If signal type changes, flush current batch first
+                    if (currentSignalType != null && currentSignalType != signalType && !batch.isEmpty()) {
```
kkondaka: This is OK but batching may be limited. Another option is to batch each type separately, which may be little bit more involved. Another option is to have 3 different sinks one for each type (log/metrics/traces). This can be achieved using conditional routing in the config like 

```
   routes:
      - logs: 'getEventType() == "LOG"'
      - traces: 'getEventType() == "TRACE"'
      - metrics: 'getEventType() == "METRIC"'
   sink:
      - otlp_sink:
              ...<other config>
              route:
                  - logs
       - otlp_sink:
              ...<other config>
              route:
                  - metrics

Comment (viquer) [reply data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/buffer/OtlpSinkBuffer.java]: I don't think it's worth to start overcomplicating the code, as you mentioned. Added documentation on the README.md for now, if we there is customer requests to do it differently to improve batching, we could definitely work this out. Let me know if that works.

Comment (dlvenable) [reply data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/buffer/OtlpSinkBuffer.java]: I think having a batch per signal makes the most sense. And having a single `otlp` sink is the ideal solution for customers.

I think ideally the buffer is signal agnostic. Then elsewhere you have a `Map<Signal, OtlpSinkBuffer>`.

If we want a short term solution, perhaps add configuration option for allowed signals. Then make it allow only one at a time.

```
sink:
  - otlp:
       allowed_signals: [metric]
```

File: data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/http/OtlpHttpSender.java [hunk 2]
```diff
@@ -139,23 +148,53 @@ public void send(@Nonnull final List<Pair<ResourceSpans, EventHandle>> batch) {
...
 
-    private Pair<byte[], byte[]> getPayloadAndCompressedPayload(final List<Pair<ResourceSpans, EventHandle>> batch) {
-        final ExportTraceServiceRequest request = ExportTraceServiceRequest.newBuilder()
-                .addAllResourceSpans(batch.stream().map(Pair::left).collect(Collectors.toList()))
-                .build();
-        final byte[] payload = request.toByteArray();
+    private Pair<byte[], byte[]> getPayloadAndCompressedPayload(final List<Pair<Object, EventHandle>> batch, final OtlpSignalType signalType) {
+        final byte[] payload;
+        
+        switch (signalType) {
```
dlvenable: Let's have a common base interface and avoid the switch.

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/http/OtlpHttpSender.java
```diff
@@ -173,64 +212,86 @@ private HttpRequest buildHttpRequest(final byte[] compressedPayload) {
...
-
-                // Optimistically release all as true, no per-span granularity
-                releaseAllEventHandle(batch, true);
-            } else {
-                sinkMetrics.incrementRecordsOut(spans);
-                releaseAllEventHandle(batch, true);
+            long rejectedCount = 0;
+            String errorMessage = "";
+            
+            switch (signalType) {
```
dlvenable: We have a few switches around the code. I think a common interface that exposes all the different methods that are variable would be ideal. Then we can avoid these switches and keep the code for each signal consolidated.

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/OtlpSignalType.java
```diff
@@ -0,0 +1,38 @@
+/*
```
dlvenable, line 1: All new files need the latest license header.

data-prepper-plugins/otlp-sink/src/test/java/org/opensearch/dataprepper/plugins/sink/otlp/OtlpSinkTest.java
```diff
@@ -1,14 +1,13 @@
-/*
```
dlvenable, line 1: Please add the latest license header.

data-prepper-plugins/otlp-sink/build.gradle
```diff
@@ -54,6 +54,13 @@ dependencies {
 
     // Metrics
     implementation 'io.micrometer:micrometer-core'
+    
+    // Test dependencies
```
dlvenable: No need for this comment.

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/metrics/OtlpSinkMetrics.java
```diff
@@ -69,21 +69,23 @@ public void registerQueueGauges(final BlockingQueue<?> queue) {
...
+     * Increments the count of records that were explicitly rejected by the OTLP endpoint.
+     * This applies to all signal types: spans, metrics, and logs.
      *
-     * @param count The number of spans rejected.
+     * @param count The number of records rejected.
      */
-    public void incrementRejectedSpansCount(final long count) {
-        pluginMetrics.counter("rejectedSpansCount").increment(count);
+    public void incrementRejectedRecordsCount(final long count) {
+        pluginMetrics.counter("rejectedRecordsCount").increment(count);
```
dlvenable, line 78: This is a breaking change since we already have this metric. 

It might be ok to make. Or we add a new metric and have a metric per signal.

data-prepper-plugins/otel-proto-common/src/test/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoEncoderRoundTripTest.java
```diff
@@ -0,0 +1,680 @@
...
+        @Test
+        void testEmptyAttributesRoundTrip() throws Exception {
+            final ExportTraceServiceRequest originalRequest = 
+                buildExportTraceServiceRequestFromJsonFile("test-request-no-spans.json");
+            
+            // This should handle gracefully even with no spans
+            final List<Span> spans = decoder.parseExportTraceServiceRequest(originalRequest, TEST_TIME);
+            
+            // If there are spans, encode them
+            for (Span span : spans) {
```
dlvenable, line 602: Assert the length first. Or this loop doesn't test anything and passes.

data-prepper-plugins/otel-proto-common/src/test/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoEncoderRoundTripTest.java
```diff
@@ -0,0 +1,680 @@
...
+
+        @Test
+        void testHistogramWithoutExplicitBounds() throws Exception {
+            final ExportMetricsServiceRequest originalRequest = 
+                buildExportMetricsServiceRequestFromJsonFile("test-histogram-metrics-no-explicit-bounds.json");
+            
+            final Collection<Record<? extends Metric>> records = 
+                decoder.parseExportMetricsServiceRequest(originalRequest, TEST_TIME);
+            
+            for (Record<? extends Metric> record : records) {
```
dlvenable, line 616: Assert the length first. Or this loop doesn't test anything and passes.

data-prepper-plugins/otel-proto-common/src/test/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoEncoderRoundTripTest.java
```diff
@@ -0,0 +1,680 @@
...
+                final Metric metric = record.getData();
+                if (metric instanceof JacksonHistogram) {
```
dlvenable, line 618: Let's avoid conditions in tests. When is this condition expected to be false? It may make sense to have multiple tests. Some without Histogram and one with Histograms.

Also, let's base it on `Histogram` and not `JacksonHistogram`.

data-prepper-plugins/otel-proto-common/src/test/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoEncoderRoundTripTest.java
```diff
@@ -0,0 +1,680 @@
...
+            
+            // Verify gauge data point is included with actual value
+            assertThat(encodedMetric.hasGauge(), equalTo(true));
+            assertThat(encodedMetric.getGauge().getDataPointsCount(), equalTo(1));
+            
+            final var originalDataPoint = originalMetric.getGauge().getDataPoints(0);
+            final var encodedDataPoint = encodedMetric.getGauge().getDataPoints(0);
+            
+            // Verify the actual numeric value is preserved
+            if (originalDataPoint.hasAsDouble()) {
```
dlvenable: Let's split this into two tests. One for doubles and one for ints. You can use `@Nested` to have common setup.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoStandardCodec.java
```diff
@@ -930,6 +938,438 @@ protected AnyValue objectToAnyValue(final Object obj) throws UnsupportedEncoding
...
+            final io.opentelemetry.proto.metrics.v1.Metric.Builder metricBuilder = 
+                io.opentelemetry.proto.metrics.v1.Metric.newBuilder();
+            
+            // Set common fields
+            metricBuilder.setName(metric.getName() != null ? metric.getName() : "");
+            metricBuilder.setDescription(metric.getDescription() != null ? metric.getDescription() : "");
+            metricBuilder.setUnit(metric.getUnit() != null ? metric.getUnit() : "");
+            
+            // Handle different metric types
+            if (metric instanceof JacksonGauge) {
```
dlvenable: Do not use the `Jackson` classes when reading. Use the interface instead. e.g. `Gauge` instead of `JacksonGauge`.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoStandardCodec.java
```diff
@@ -930,6 +938,438 @@ protected AnyValue objectToAnyValue(final Object obj) throws UnsupportedEncoding
...
+                    constructExponentialHistogram((JacksonExponentialHistogram) metric));
+            } else {
+                throw new UnsupportedEncodingException(
+                    "Unsupported metric type: " + metric.getClass().getName());
+            }
+            
+            return metricBuilder.build();
+        }
+
+        protected io.opentelemetry.proto.metrics.v1.Gauge constructGauge(final JacksonGauge gauge)
```
dlvenable: Use `Gauge`.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoStandardCodec.java
```diff
@@ -930,6 +938,438 @@ protected AnyValue objectToAnyValue(final Object obj) throws UnsupportedEncoding
...
+                gauge.getAttributes(),
+                gauge.getExemplars(),
+                gauge.getFlags()
+            );
+            
+            builder.addDataPoints(dataPoint);
+            return builder.build();
+        }
+
+        protected io.opentelemetry.proto.metrics.v1.Sum constructSum(final JacksonSum sum)
```
dlvenable: Use `Sum`

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoStandardCodec.java
```diff
@@ -930,6 +938,438 @@ protected AnyValue objectToAnyValue(final Object obj) throws UnsupportedEncoding
...
+                sum.getAttributes(),
+                sum.getExemplars(),
+                sum.getFlags()
+            );
+            
+            builder.addDataPoints(dataPoint);
+            return builder.build();
+        }
+
+        protected io.opentelemetry.proto.metrics.v1.Histogram constructHistogram(
```
dlvenable, line 1095: Use `Histogram`. I see a few places like this. Only use the `JacksonX` classes when constructing.

dlvenable (changes requested):
Thank you @viquer . This is a great contribution to have and will greatly improve the sink.

tillwulfram:
@viquer thank you for the great work and contribution, really appreciate it!

I had a look at the code and wanted to raise two related gaps for non-AWS backends — apologies if I've missed something that's already there! And sry for the long comment, just wanted to explain my thoughts :)

---

### 1. Custom headers support

Would it be possible to add a `headers` config field? This would cover bearer tokens, basic auth, and any backend that requires custom headers for authentication — all in one simple map.

**`something like`**:
```java
@Getter
@JsonProperty("headers")
private Map<String, String> headers = Collections.emptyMap();
```
**`and avoid duplicate headers in sender`**:
```java
config.getHeaders().forEach(headersBuilder::set);
```
**for usage like this**
```yaml
sink:
  - otlp:
      endpoint: "https://my-backend.example.com/v1/logs"
      headers:
        Authorization: "Bearer <token>"
        API-Key: "otlp-endpoint-api-key"
```
---

One thing worth noting: `isAwsConfigValid()` currently enforces `awsConfig != null`, which means `aws: {}` is always required. Since there are valid use cases without any auth at all (e.g. sending cluster-internally) or just one
```java
return awsConfig != null || headers != null;
```
---

### 2. TLS configuration (`insecure` / custom CA)

Currently `buildWebClient()` has no TLS options, which means self-signed or internal CA certs (common in private clusters) will fail with a handshake error.
Adding options like `insecure: true` and a custom CA path (similar to what the `opensearch` sink already supports) would make the plugin usable in a lot more environments.

---

If neither is on your roadmap anytime soon, please let me know!

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/metrics/OtlpSinkMetrics.java
viquer: This was intentional. AFAIK the plugin has never been used in production, I could be wrong. @kkondaka  has more context. If we want to keep it then we would likely have metrics per signal.

LaszloKovacs9001:
The work from this PR has been incorporated into #6768 with DCO fixes and additional enhancements (configurable SigV4 service name, additional headers, explicit region config). Thank you @viquer for the great foundation!

dlvenable:
@viquer , Thank you for starting this work! I'm going to close this PR. If you want further changes to the merged PR, please open a new PR.

PR #6250, opened 2025-11-07
Otel logs source http service

TomasLongo:
## Description
This PR adds an HTTP-Service to the OtelLogsSource, which runs alongside the gRPC-Service under a separate endpoint.

Both Services share the same underlying Armeria-Server, running under the same port.

## Breakdown

The PR is kind of heavy. The OtelLogsSource had to be restructured resulting in the refactoring of the present tests. 

A quick breakdown of the changes will certainly help with the review

### One Server, two services

`OtelLogsSource` is now inhabitetd by two services running under different paths behind the same port.
The path of the services can be configured by the two fields `path` and `http_path`

- `path` is the already present field for configuring path of the grpc service and keeps its purpose
- `http_path` will set the path for the new http service

### **Unframed Requests**

Since unframed requests served the purpose of sending plain `http` requests to the grpc server, this functionality has been dropped. However, for backwardscompatibilty of the config, it is still present in the config data structure.

### **CreateServer**

The class `CreateServer` deals with creating both, http and gRPC servers. However, in its current form, the intended use is to create a server with a single service. Two quick attempts to rewrite the Class for the purpose if this PR yielded no results.

So the final version  of the server creation loosely resembles the functionality found in `CreateServer`.

### Refactoring of tests

Many tests were making sure that the former gRPC server, with unframed requests enabled, would handle http requests correctly. I have dropped this tests completely, replacing them with a separate test class that handles tests regarding the http functionality. I added a few improvements along the way like

- Getting rid of unnecessary mocks
- Introducing parameterized tests
- Using unified test data and fixtures

## Decisions made and additional considerations

**Getting completely rid of unframed requests**

`enable_unframed_requests` has no effect any more, since its purpose has been replaced with the new http service

**Configuration**

A new field, `http_path` has been added to specify the path of the http service. However, one could extend the config to

- enable the client to specify only on of both service by omitting one of both paths. Currently, both services will be instantiated.

**CreateServer**

I think that the intention of `CreateServer`, to provide a unified mechanism to bootstrap armeria servers, is absolutely the right thing todo. However, in its current form, it only supports creating servers with a single service.

It's worth thinking about how to refactor `CreateServer` to streamline server creation for future plugins. The challenge I encountered was how to best model the fact, that different services might need different server capabilities/configurations.

Figuring this out, would have gone well beyond the scope of this PR.
 
### Issues Resolved
Resolves #[Issue number to be closed when this PR is merged]
 
### Check List
- [ ] New functionality includes testing.
[...]

PR #6494, opened 2026-02-06
Fix OTEL trace source broken by PR 5322

kkondaka:
### Description
Fix OTEL trace source broken by PR #5322

The PR was started before OTEL standard support was added committed long after OTEL standard support was added. This resulted in regression of support for OTEL standard in otel-trace-source
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/otel-trace-source/src/main/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/ArmeriaHttpService.java
+++ b/data-prepper-plugins/otel-trace-source/src/main/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/ArmeriaHttpService.java
@@ -14,9 +14,6 @@ import org.opensearch.dataprepper.model.buffer.Buffer;
 import org.opensearch.dataprepper.model.record.Record;
 import org.opensearch.dataprepper.model.trace.Span;
 import org.opensearch.dataprepper.plugins.otel.codec.OTelProtoCodec;
-import org.opensearch.dataprepper.plugins.otel.codec.OTelProtoOpensearchCodec;
-import org.opensearch.dataprepper.plugins.otel.codec.OTelProtoStandardCodec;
-import org.opensearch.dataprepper.plugins.otel.codec.OTelOutputFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -52,9 +49,9 @@ public class ArmeriaHttpService {
     private final DistributionSummary payloadSizeSummary;
     private final Timer requestProcessDuration;
 
-    public ArmeriaHttpService(Buffer<Record<Object>> buffer, final OTelOutputFormat otelOutputFormat, final PluginMetrics pluginMetrics, final int bufferWriteTimeoutInMillis) {
+    public ArmeriaHttpService(Buffer<Record<Object>> buffer, final OTelProtoCodec.OTelProtoDecoder otelProtoDecoder, final PluginMetrics pluginMetrics, final int bufferWriteTimeoutInMillis) {
         this.buffer = buffer;
-        this.oTelProtoDecoder = (otelOutputFormat == OTelOutputFormat.OPENSEARCH) ? new OTelProtoOpensearchCodec.OTelProtoDecoder() : new OTelProtoStandardCodec.OTelProtoDecoder();
+        this.oTelProtoDecoder = otelProtoDecoder;
         this.bufferWriteTimeoutInMillis = bufferWriteTimeoutInMillis;
 
         requestsReceivedCounter = pluginMetrics.counter(REQUESTS_RECEIVED);
--- a/data-prepper-plugins/otel-trace-source/src/test/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/ArmeriaHttpServiceTest.java
+++ b/data-prepper-plugins/otel-trace-source/src/test/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/ArmeriaHttpServiceTest.java
@@ -20,11 +20,14 @@ import java.util.Collection;
 import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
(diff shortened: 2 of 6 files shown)


KarstenSchnitter (changes requested):
Thanks for fixing this, I think, you missed the ArmeriaHttpService: <https://github.com/kkondaka/kk-data-prepper-f2/blob/fb32017455fb5092ecf9cae4d938c7589617624b/data-prepper-plugins/otel-trace-source/src/main/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/ArmeriaHttpService.java#L55>

data-prepper-plugins/otel-trace-source/src/test/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/HttpServiceTest.java
```diff
@@ -0,0 +1,120 @@
...
+        when(oTelTraceSourceConfig.getRequestTimeoutInMillis()).thenReturn(5000);
+        when(oTelTraceSourceConfig.getAuthentication()).thenReturn(null);
+
+        ArmeriaHttpService result = httpService.create(serverBuilder, buffer);
+
+        assertNotNull(result);
+    }
+
+    @Test
+    void testCreateWithAuthentication() {
```
TomasLongo, line 102: Is there a reason to assert internals instead of treating the service as black box and assert its behaviour?

This is a good example for it. The test could bootstrap the trace source and make sure that it doesn't accept requests that do not carry authentication information.

In this form, the test depends on the exact invocation of some internal component.

I'm addressing this here, because in #6250 I followed my line of reasoning getting rid of all the mocking, bootstrapping the service and testing with actual requests against it. Is that something to avoid?

data-prepper-plugins/otel-trace-source/src/test/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/ArmeriaHttpServiceTest.java
```diff
@@ -0,0 +1,219 @@
...
+            return null;
+        }).when(buffer).writeAll(any(), any(Integer.class));
+        try (MockedStatic<ServiceRequestContext> mocked = mockStatic(ServiceRequestContext.class)) {
+            mocked.when(ServiceRequestContext::current).thenReturn(staticServiceRequestContext);
+            armeriaHttpService = new ArmeriaHttpService(buffer, outputFormat, pluginMetrics, 5000);
+            ExportTraceServiceRequest request = createExportTraceRequest();
+
+            ExportTraceServiceResponse response = armeriaHttpService.exportTrace(request);
+
+            assertNotNull(response);
```
dlvenable, line 123: Let's move as much out of the try-with-resources block as possible. I'm fairly sure that the rest of this can go out of it.

data-prepper-plugins/otel-trace-source/src/test/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/ArmeriaHttpServiceTest.java
```diff
@@ -0,0 +1,219 @@
...
+            verify(requestsReceivedCounter).increment();
+            verify(payloadSizeSummary).record(request.getSerializedSize());
+            verify(successRequestsCounter).increment();
+            assertThat(recordsReceived.size(), equalTo(1));
+            Span span = recordsReceived.get(0).getData();
+            if (outputFormat == OTelOutputFormat.OTEL) {
```
dlvenable: Let's avoid conditions. Instead, make an `ArgumentsProvider` class which produces these as inputs.

Something like this:

```
    static class OTelOutputFormatToServiceNameKey implements ArgumentsProvider {
        @Override
        public Stream<? extends Arguments> provideArguments(final ExtensionContext extensionContext) {
            return Stream.of(
                    arguments(OTelOutputFormat.OPENSEARCH, "attributes/resource.attributes.service@name"),
                    arguments(OTelOutputFormat.OTEL, "resource/attributes/service.name")
            );
        }
    }
```

Then you just have:

```
assertThat(span.get(serviceNameKey, String.class), equalTo(TRACE_SERVICE_NAME));
```

data-prepper-plugins/otel-trace-source/src/test/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/HttpServiceTest.java
dlvenable: I think as a unit test, it is appropriate to have something like this. Though, we should also have tests that are more black box. Using a real server and web client provide much better tests for authentication.

data-prepper-plugins/otel-trace-source/src/main/java/org/opensearch/dataprepper/plugins/source/oteltrace/http/ArmeriaHttpService.java
```diff
@@ -50,9 +52,9 @@ public class ArmeriaHttpService {
     private final DistributionSummary payloadSizeSummary;
     private final Timer requestProcessDuration;
 
-    public ArmeriaHttpService(Buffer<Record<Object>> buffer, final PluginMetrics pluginMetrics, final int bufferWriteTimeoutInMillis) {
+    public ArmeriaHttpService(Buffer<Record<Object>> buffer, final OTelOutputFormat otelOutputFormat, final PluginMetrics pluginMetrics, final int bufferWriteTimeoutInMillis) {
         this.buffer = buffer;
-        this.oTelProtoDecoder = new OTelProtoOpensearchCodec.OTelProtoDecoder();
+        this.oTelProtoDecoder = (otelOutputFormat == OTelOutputFormat.OPENSEARCH) ? new OTelProtoOpensearchCodec.OTelProtoDecoder() : new OTelProtoStandardCodec.OTelProtoDecoder();
```
dlvenable: This is the same as the change above. We'd be better off injecting the `OTelProtoDecoder` into these classes and making this decision in a single location.

dlvenable (changes requested):
I left some comments. I also see a number of failing GitHub Actions.

kkondaka:
Actually I wanted @TomasLongo to write the unit tests separately because they were not there when the PR was merged. But I quickly generated few unit tests using AI :-)

kkondaka:
@KarstenSchnitter please review this.

KarstenSchnitter (approved):
Thanks for providing the fixe and adding the tests.

PR #6507, opened 2026-02-12
Convert Kafka Header values to Strings

kkondaka:
### Description
See the discussion - https://github.com/opensearch-project/data-prepper/discussions/4917
The issue is that kafka source keeps all headers to bytes because kafka header values are byte arrays. But if the input data has headers as strings they are converted to byte arrays by Kafka on the producer side. The consumer side (ie Data Prepper kafka source) keeps them byte arrays. This can cause problem if the original header value is a string. 

So, it makes sense to convert the header values that can be converted to Strings to Strings. Those values that cannot be converted to Strings are left as bytes. Customer could potentially use type conversion processor to convert them to desired type.

 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
@@ -510,11 +510,16 @@ public class KafkaCustomConsumer implements Runnable, ConsumerRebalanceListener 
             Map<String, Object> headerData = new HashMap<>();
             for (Header header: headers) {
                 byte[] headerValue = header.value();
-                try {
-                    headerData.put(header.key(), ((headerValue != null) ? new String(header.value(), StandardCharsets.UTF_8) : null));
-                } catch (Exception e) {
-                    headerData.put(header.key(), headerValue);
+                if (headerValue == null) {
+                    headerData.put(header.key(), null);
+                    continue;
                 }
+                String strValue = new String(header.value(), StandardCharsets.UTF_8);
+                if (Arrays.equals(headerValue, strValue.getBytes())) {
+                    headerData.put(header.key(), strValue);
+                    continue;
+                }
+                headerData.put(header.key(), headerValue);
             }
             eventMetadata.setAttribute("kafka_headers", headerData);
         }
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
@@ -505,22 +505,8 @@ public class KafkaCustomConsumer implements Runnable, ConsumerRebalanceListener 
         if (kafkaKeyMode == KafkaKeyMode.INCLUDE_AS_METADATA) {
             eventMetadata.setAttribute("kafka_key", key);
         }
-        Headers headers = consumerRecord.headers();
-        if (headers != null) {
-            Map<String, Object> headerData = new HashMap<>();
-            for (Header header: headers) {
-                byte[] headerValue = header.value();
-                if (headerValue == null) {
-                    headerData.put(header.key(), null);
-                    continue;
-                }
-                String strValue = new String(header.value(), StandardCharsets.UTF_8);
-                if (Arrays.equals(headerValue, strValue.getBytes())) {
-                    headerData.put(header.key(), strValue);
-                    continue;
-                }
-                headerData.put(header.key(), headerValue);
-            }
+        Map<String, Object> headerData = KafkaHeadersExtractor.extractMessageHeaders(consumerRecord.headers());
+        if (headerData != null) {
             eventMetadata.setAttribute("kafka_headers", headerData);
         }
         final long receivedTimeStamp = getRecordTimeStamp(consumerRecord, now.toEpochMilli());
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
@@ -51,6 +51,7 @@ import software.amazon.awssdk.services.glue.model.AccessDeniedException;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
@@ -506,9 +507,14 @@ public class KafkaCustomConsumer implements Runnable, ConsumerRebalanceListener 
         }
         Headers headers = consumerRecord.headers();
         if (headers != null) {
-            Map<String, byte[]> headerData = new HashMap<>();
+            Map<String, Object> headerData = new HashMap<>();
             for (Header header: headers) {
-                headerData.put(header.key(), header.value());
+                byte[] headerValue = header.value();
+                try {
+                    headerData.put(header.key(), ((headerValue != null) ? new String(header.value(), StandardCharsets.UTF_8) : null));
+                } catch (Exception e) {
+                    headerData.put(header.key(), headerValue);
+                }
             }
             eventMetadata.setAttribute("kafka_headers", headerData);
         }
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
@@ -27,8 +27,6 @@ import org.apache.kafka.common.errors.AuthenticationException;
 import org.apache.kafka.common.errors.RebalanceInProgressException;
 import org.apache.kafka.common.utils.ExponentialBackoff;
 import org.apache.kafka.common.errors.RecordDeserializationException;
-import org.apache.kafka.common.header.Header;
-import org.apache.kafka.common.header.Headers;
 import org.opensearch.dataprepper.model.acknowledgements.AcknowledgementSet;
 import org.opensearch.dataprepper.model.acknowledgements.AcknowledgementSetManager;
 import org.opensearch.dataprepper.model.buffer.Buffer;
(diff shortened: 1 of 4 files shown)


dlvenable:
@kkondaka , Why do we want to covert bytes to string? What is the issue with the current behavior? Is this a potential breaking change?

data-prepper-plugins/kafka-plugins/src/test/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumerTest.java
```diff
@@ -332,6 +340,54 @@ public void testPlainTextConsumeRecords() throws InterruptedException {
...
+        when(kafkaConsumer.poll(any(Duration.class))).thenReturn(consumerRecords);
+        consumer = createObjectUnderTest("plaintext", false);
+
+        try {
+            consumer.onPartitionsAssigned(List.of(new TopicPartition(topic, testPartition)));
+            consumer.consumeRecords();
+        } catch (Exception e){}
+        final Map.Entry<Collection<Record<Event>>, CheckpointState> bufferRecords = buffer.read(1000);
+        ArrayList<Record<Event>> bufferedRecords = new ArrayList<>(bufferRecords.getKey());
+        Assertions.assertEquals(consumerRecords.count(), bufferedRecords.size());
```
dlvenable: Use a static import.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -472,9 +473,14 @@ private <T> Record<Event> getRecord(ConsumerRecord<String, T> consumerRecord, in
         }
         Headers headers = consumerRecord.headers();
         if (headers != null) {
-            Map<String, byte[]> headerData = new HashMap<>();
+            Map<String, Object> headerData = new HashMap<>();
             for (Header header: headers) {
-                headerData.put(header.key(), header.value());
+                byte[] headerValue = header.value();
+                try {
+                    headerData.put(header.key(), ((headerValue != null) ? new String(header.value(), StandardCharsets.UTF_8) : null));
```
graytaylor0: Can't we get weird results converting byte to string like this? What do the headers look like? Just raw bytes or are they base64 encoded, etc?

kkondaka: Not sure I understand. Even if it is based64 encoded, it would still be a valid string. Right now, we are keeping them as bytes. This code converts valid strings as strings.

dlvenable: What are the possible data types that Kafka can produce here? This can work for a String. But, what if the data is an integer or float? 

Could this data also come as either a list or a map?

dlvenable (changes requested):

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -472,9 +473,19 @@ private <T> Record<Event> getRecord(ConsumerRecord<String, T> consumerRecord, in
         }
         Headers headers = consumerRecord.headers();
         if (headers != null) {
-            Map<String, byte[]> headerData = new HashMap<>();
+            Map<String, Object> headerData = new HashMap<>();
             for (Header header: headers) {
-                headerData.put(header.key(), header.value());
+                byte[] headerValue = header.value();
```
dlvenable: The code and the tests for this are already very complicated. We can simplify this by moving the header extraction into its own class. Then you can cover all the complexities in the unit tests for that class.

The unit tests for the customer consumer just need to mock the header extractor class and verify the mocked values are there.

data-prepper-plugins/kafka-plugins/src/test/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumerTest.java
```diff
@@ -303,30 +312,76 @@ public void testPlainTextConsumeRecords() throws InterruptedException {
...
         });
-        Assertions.assertEquals(consumer.getNumRecordsCommitted(), 2L);
+        assertEquals(consumer.getNumRecordsCommitted(), 2L);
 
+        for (Record<Event> record: bufferedRecords) {
+            Event event = record.getData();
+            String value1 = event.get(testKey1, String.class);
+            String value2 = event.get(testKey2, String.class);
+            assertTrue(value1 != null || value2 != null);
+            if (value1 != null) {
```
dlvenable: Let's not use conditionals in tests. The test become too complicated and we are not sure we are even testing what we want. Refactor the tests to have two different test cases instead. There are a few other situations like this in this PR.

Also, my suggestion to refactor the code can also simplify this.

data-prepper-plugins/kafka-plugins/src/test/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumerTest.java
```diff
@@ -697,6 +752,29 @@ private ConsumerRecords createPlainTextRecords(String topic, final long startOff
...
+        Map<TopicPartition, List<ConsumerRecord>> records = new HashMap<>();
+        RecordHeaders headers1 = new RecordHeaders();
+        RecordHeaders headers2 = new RecordHeaders();
+
+        // Adding a String header
+        headers1.add("test-string-header", testStringHeader.getBytes(StandardCharsets.UTF_8));
+        headers2.add("test-string-header", testStringHeader.getBytes(StandardCharsets.UTF_8));
+
+        // Adding an Integer header (stored as 4 bytes)
+        int testIntHeader = 5;
```
dlvenable, line 856: Use more random data. Or parameterized data. This will be easier with a dedicated header class and its own test suite.

dlvenable (changes requested):

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
kkondaka: I have updated the description with more information. Basically, most header values are strings. If a value is not a string, it is kept as byte array and a type conversion processor can be used to convert to desired type.

This is not a breaking change because the current implementation is not correct.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
kkondaka: Moved to a separate class.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaHeadersExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+        }
+        Map<String, Object> headerData = new HashMap<>();
+        for (Header header: headers) {
+            byte[] headerValue = header.value();
+            if (headerValue == null) {
+                headerData.put(header.key(), null);
+                continue;
+            }
+            String strValue = new String(header.value(), StandardCharsets.UTF_8);
+            System.out.println("____"+headerValue.length+"...."+strValue.getBytes().length);
```
dlvenable: Remove the `println` statements. These will probably fail the build as well.

data-prepper-plugins/kafka-plugins/src/test/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumerTest.java
```diff
@@ -316,30 +325,76 @@ public void testPlainTextConsumeRecords() throws InterruptedException {
...
+            assertEquals(topicPartition.partition(), testPartition);
+            assertEquals(topicPartition.topic(), topic);
+            assertEquals(offsetAndMetadata.offset(), 2L);
         });
-        Assertions.assertEquals(consumer.getNumRecordsCommitted(), 2L);
+        assertEquals(consumer.getNumRecordsCommitted(), 2L);
 
+        for (Record<Event> record: bufferedRecords) {
+            Event event = record.getData();
+            String value1 = event.get(testKey1, String.class);
```
dlvenable, line 340: This is not a clear test. I think you are testing these values in any order. You can put them in a `Map` in advance. Or you can use Hamcrest matchers to assert values in any order. Either way, let's make this test a bit clearer and less nested.

data-prepper-plugins/kafka-plugins/src/test/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumerTest.java
```diff
@@ -316,30 +325,76 @@ public void testPlainTextConsumeRecords() throws InterruptedException {
...
+        });
+        assertEquals(consumer.getNumRecordsCommitted(), 2L);
 
         for (Record<Event> record: bufferedRecords) {
             Event event = record.getData();
             String value1 = event.get(testKey1, String.class);
             String value2 = event.get(testKey2, String.class);
+            Map<String, Object> attributes = event.getMetadata().getAttributes();
+            Map<String, Object> kafkaHeaders = (Map<String, Object>) attributes.get("kafka_headers");
+            assertThat(kafkaHeaders.get("test-string-header"), equalTo(testStringHeader));
```
dlvenable, line 386: Similar comment as before. Simplify these tests.

dlvenable (changes requested):

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaHeadersExtractor.java
kkondaka: @dlvenable Looks like you are looking at older version? Not the latest.

data-prepper-plugins/kafka-plugins/src/test/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumerTest.java
kkondaka: Same here. Please check the latest revision.

dlvenable (approved):

graytaylor0 (approved):

PR #6508, opened 2026-02-13
Fix Sink Metrics Request Latency Metric units

kkondaka:
### Description
Make Sink Metrics Request Latency Metric units as milliseconds because all other latency metrics in the pipeline are in milliseconds
 
### Issues Resolved
 
### Check List


data-prepper-plugins/common/src/main/java/org/opensearch/dataprepper/common/sink/DefaultSinkOutputStrategy.java
```diff
@@ -40,7 +40,8 @@ public void flushBuffer() {
         try {      
             SinkFlushResult flushResult = flushableBuffer.flush();
             if (flushResult == null) { // success
-                sinkMetrics.recordRequestLatency((double)(System.nanoTime() - startTime));
+                double latency = (System.nanoTime() - startTime)/1000_000; // convert to ms
```
dlvenable, line 43: This would be clearer: `1_000_000`

dlvenable:
@kkondaka , I labeled this as a breaking change so that we can note it in the release notes. I think it is the right solution, but we just need to be clear on it.

PR #6512, opened 2026-02-13
Add additional metrics for otel logs source and kafka buffer producer

graytaylor0:
### Description
This change adds a couple of metrics for visibility into request latency with KafkaBuffer and OtelLogsSource. Enables the following kafka producer side metrics

```
"request-latency-avg", "requestLatencyAvg",
            "request-latency-max", "requestLatencyMax",
            "produce-throttle-time-avg", "produceThrottleTimeAvg",
            "produce-throttle-time-max", "produceThrottleTimeMax"
```

Also adds one metric to the otel logs source time to parse the ExportLogsServiceRequest (`requestParsingDuration`)

One more metric is added `produceDataPreparationTime` for Kafka producers which tracks time taken to prepare the data to send to Kafka (currently just the compression piece).
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/otel-logs-source/src/main/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsGrpcService.java
+++ b/data-prepper-plugins/otel-logs-source/src/main/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsGrpcService.java
@@ -27,6 +27,7 @@ import org.slf4j.LoggerFactory;
 
 import java.time.Instant;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 public class OTelLogsGrpcService extends LogsServiceGrpc.LogsServiceImplBase {
@@ -36,6 +37,7 @@ public class OTelLogsGrpcService extends LogsServiceGrpc.LogsServiceImplBase {
     public static final String SUCCESS_REQUESTS = "successRequests";
     public static final String PAYLOAD_SIZE = "payloadSize";
     public static final String REQUEST_PROCESS_DURATION = "requestProcessDuration";
+    public static final String REQUEST_PARSING_DURATION = "requestParsingDuration";
 
     private final int bufferWriteTimeoutInMillis;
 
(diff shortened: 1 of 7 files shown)


data-prepper-plugins/otel-logs-source/src/main/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsGrpcService.java
```diff
@@ -92,17 +98,19 @@ private void processRequest(final ExportLogsServiceRequest request, final Stream
         final List<OpenTelemetryLog> logs;
 
         try {
+            final long startParsingTime = System.currentTimeMillis();
             logs = oTelProtoDecoder.parseExportLogsServiceRequest(request, Instant.now());
+            requestParsingDuration.record(System.currentTimeMillis() - startParsingTime, TimeUnit.MILLISECONDS);
```
kkondaka, line 103: We should probably call it requestValidationDuration because we should replace this code to do just validation without creating Log events.

graytaylor0: Well this code is still used when there is no byteBuffer used right? And in that context it is actually used for parsing the request

dlvenable: I agree with @graytaylor0 that we should use the more generic one for both scenarios.

kkondaka: As discussed, in case of byte buffer, we do not need to create log events to save cpu and memory resources. So, we need to do slightly different thing for byte buffer.

kkondaka: Anyway, I can approve it for now. Was just indicating that it would be different "duration" for byte buffer vs not bytebuffer.

PR #6513, opened 2026-02-13
Use a Timer for the sqsSinkRequestLatency metric to report it with time units

dlvenable:
### Description

This uses a Micrometer `Timer` for the `sqsSinkRequestLatency` metrics. This will have it report using time units.

See a similar PR #6510 for more information.
 
### Issues Resolved

N/A
 
### Check List

### DIFF
--- a/data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkExecutor.java
+++ b/data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkExecutor.java
@@ -1,6 +1,10 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  */
 
 package org.opensearch.dataprepper.plugins.sink.sqs;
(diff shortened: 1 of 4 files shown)


data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkExecutor.java
```diff
@@ -82,7 +87,7 @@ public void flushBuffer() {
         if (failedStatus != null) {
             pushFailedObjectsToDlq(failedStatus);
         } else {
-            recordLatency((double)System.nanoTime() - startTime);
+            recordLatency(System.nanoTime() - startTime, TimeUnit.NANOSECONDS);
```
graytaylor0, line 90: Why do we use nanoseconds here instead of just milliseconds?

dlvenable: That is the current resolution that we report. So I didn't change it.

But, it shouldn't matter much. Micrometer will send these using units and CloudWatch can normalize them for operators.

graytaylor0: Ok if cloudwatch normalizes then that's good. We should just be seeing consistent numbers for latency on metrics across Data Prepper

PR #6539, opened 2026-02-19
otel_apm_service_map: Added support for deriving remote service and remote operation

kkondaka:
### Description
otel_apm_service_map: Added support for deriving remote service and remote operation.
Derives operation differently for server and client spans based on various attributes.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/otel-proto-common/src/test/java/org/opensearch/dataprepper/plugins/otel/common/AwsServiceMappingsProviderTest.java
+++ b/data-prepper-plugins/otel-proto-common/src/test/java/org/opensearch/dataprepper/plugins/otel/common/AwsServiceMappingsProviderTest.java
@@ -1,6 +1,11 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
  */
 
 package org.opensearch.dataprepper.plugins.otel.common;
--- a/data-prepper-plugins/otel-proto-common/src/test/java/org/opensearch/dataprepper/plugins/otel/common/ServiceAddressPortAttributesProviderTest.java
+++ b/data-prepper-plugins/otel-proto-common/src/test/java/org/opensearch/dataprepper/plugins/otel/common/ServiceAddressPortAttributesProviderTest.java
@@ -1,6 +1,11 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
  */
 
 package org.opensearch.dataprepper.plugins.otel.common;


data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -231,6 +266,140 @@ public static String computeOperationName(final String spanName, final Map<Strin
...
+            "dynamodb", "AWS::DynamoDB",
+            "DynamoDBv2", "AWS::DynamoDB",
+            "sns", "AWS::SNS",
+            "AmazonSNS", "AWS::SNS",
+            "SimpleNotificationService", "AWS::SNS",
+            "AmazonKinesis", "AWS::Kinesis",
+            "Amazon S3", "AWS::S3",
+            "s3", "AWS::S3");
+
+    public static Pair<String,String> computeRemoteOperationAndService(final Map<String, Object> spanAttributes) {
```
dlvenable: Does this need to be public? If so, lets return a model instead of a `Pair`.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -231,6 +266,140 @@ public static String computeOperationName(final String spanName, final Map<Strin
...
+            remoteOperation = httpMethod != null ? httpMethod + " " + extractFirstPathFromUrl(urlString) : urlString;
+        }
+
+        return Pair.of(
+                remoteOperation != null ? remoteOperation : "UnknownRemoteOperation",
+                remoteService != null ? remoteService : "UnknownRemoteService");
+    }
+
+    private static String deriveServiceFromNetwork(final Map<String, Object> spanAttributes, final String urlString) {
+        final String[][] addressPortPairs = {
```
dlvenable: This also can come from a configuration file.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -231,6 +266,140 @@ public static String computeOperationName(final String spanName, final Map<Strin
...
+            "AmazonKinesis", "AWS::Kinesis",
+            "Amazon S3", "AWS::S3",
+            "s3", "AWS::S3");
+
+    public static Pair<String,String> computeRemoteOperationAndService(final Map<String, Object> spanAttributes) {
+        String remoteOperation = null;
+        String remoteService = null;
+
+        // RPC attributes
+        final String rpcService = getStringAttribute(spanAttributes, "rpc.service");
```
dlvenable: There is a lot in here that can change or be added to. Let's find a way to avoid requiring code changes for these. Perhaps using a resources file that has the mappings. Then the code can read the file for these mappings.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -231,6 +266,140 @@ public static String computeOperationName(final String spanName, final Map<Strin
         return spanName != null ? spanName : "UnknownOperation";
     }
 
+    private static final Map<String, String> AWS_SERVICE_MAPPINGS = Map.of(
```
dlvenable: Let's put this into a file as well.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -85,14 +102,23 @@ public static void deriveAttributesForSpan(final Span span) {
 
             final ErrorFaultResult errorFault = computeErrorAndFault(span.getStatus(), spanAttributes);
 
-            final String operationName = computeOperationName(span.getName(), spanAttributes);
+            String operationName = null; 
+            if (isServerSpan(span)) {
+                operationName = computeOperationName(span.getName(), spanAttributes);
+            } else if (isClientSpan(span) || isProducerSpan(span)) {
+                final Pair<String, String> remoteOperationAndService = computeRemoteOperationAndService(spanAttributes);
```
dlvenable: Is this change going to affect `service_map` as well? Is that ok?

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -48,12 +54,23 @@ public static void deriveServerSpanAttributes(final List<Span> spans) {
...
 
         for (final Span span : spans) {
-            if (span != null && SPAN_KIND_SERVER.equals(span.getKind())) {
+            if (span != null && isServerSpan(span)) {
                 deriveAttributesForSpan(span);
             }
         }
     }
 
+    private static boolean isServerSpan(Span span) {
```
dlvenable, line 61: These would be better as default methods in the `Span` interface.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
kkondaka: It is just adding an attribute. Not going to change any thing in the service map

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
vamsimanohar: @kkondaka  can we make changes for environment detection as well. there are more clauses for cloud specific environemnt?

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
kkondaka: @dlvenable it's not just about the list of things but the logic to extract the derived fields is not same. We cannot capture the logic in the resource file, right?

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
kkondaka: Span is in data-prepper-api and it doesn't import io.opentelemetry.proto.trace.v1.Span. And this method should actually be. comparing with io.opentelemetry.proto.trace.v1.Span.SpanKind.SPAN_KIND_SERVER. I will keep it here for now. If needed we can refactor to a different file under otel-proto-common in future

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -48,12 +58,69 @@ public static void deriveServerSpanAttributes(final List<Span> spans) {
         }
 
         for (final Span span : spans) {
-            if (span != null && SPAN_KIND_SERVER.equals(span.getKind())) {
+                System.out.println("--checking-----SERVER SPAN---"+isServerSpan(span));
```
dlvenable: Remove `println`. This is probably failing the build.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
dlvenable: That reasoning makes sense and we probably should not add this to the package. Once we get the EventFactory fully working we can probably solve this because we won't need to have it in `data-prepper-api`.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/RemoteOperationAndService.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ */
+
+package org.opensearch.dataprepper.plugins.otel.common;
+
+public class RemoteOperationAndService {
```
dlvenable: Can these classes be package protected? If so, let's start there and open only as needed.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -54,22 +58,68 @@ public static void deriveServerSpanAttributes(final List<Span> spans) {
...
+                throw e;
+            }
+        } catch (final Exception e) {
+            LOG.error("An exception occurred while initializing hostport attribute list for Data Prepper", e);
+        }
+        return null;
+    }
+
+
+    public Map<String, String> getAwsServiceMappingsFromResources() {
```
dlvenable: This should have it's own class and it's own unit testing. This makes it more general purpose for extension.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -54,22 +58,68 @@ public static void deriveServerSpanAttributes(final List<Span> spans) {
...
+        try (
+            final InputStream inputStream = getClass().getClassLoader().getResourceAsStream(SERVICE_MAPPINGS_FILE);
+        ) {
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream))) {
+                return reader.lines()
+                .filter(line -> line.contains(","))
+                .collect(Collectors.toMap(
+                    line -> line.split(",", 2)[0].trim(),
```
dlvenable: Add some error handling if the line doesn't split as expected.

Also, only split once and re-use that array.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -54,22 +58,68 @@ public static void deriveServerSpanAttributes(final List<Span> spans) {
...
                 deriveAttributesForSpan(span);
             }
         }
     }
 
+    private AddressPortAttributeKeys getAddressPortAttributeKeys(final String str) {
+        return new AddressPortAttributeKeys(str);
+    }
+
+    public List<AddressPortAttributeKeys> getAddressPortAttributeKeysList() {
```
dlvenable: This also should have its own class and its own unit tests. This makes is ready for expansion.

data-prepper-plugins/otel-proto-common/src/main/resources/aws_service_mappings
```diff
@@ -0,0 +1,12 @@
+AmazonDynamoDBv2,AWS::DynamoDB
```
dlvenable, line 1: Rename to `aws_service_mappings`.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/OTelSpanDerivationUtil.java
```diff
@@ -460,6 +438,23 @@ static String getStringAttributeFromMap(final Map<String, Object> map, final Str
     }
 
 
+    public static class AddressPortAttributeKeys {
+        final String address;
+        final String port;
+        public AddressPortAttributeKeys(final String str) {
+            this.address = str.split(",")[0];
```
dlvenable: Split once and re-use the array. Also, add error handling with a useful error message if there are not exactly two elements in the array.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/common/RemoteOperationAndService.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public String getOperation() {
+        return remoteOperation;
+    }
+
+    public String getService() {
+        return remoteService;
+    }
+
+    public boolean isNull() {
```
dlvenable: This code would be clearer if you return `Optional<RemoteOperationAndService>` instead of having an `isNull` method.

An approach to make the call simple would be a static helper like this:

```
Optional<RemoteOperationAndService> fromNullableFields(final String operation, final String service) {
  if(operation == null || service == null)
    return Optional.empty();
  return Optional.of(new RemoteOperationAndService(operation, service);
}
```

kkondaka: We can't return empty if one of them is NULL. But we need to know that one of them is null so that we can set it to default value

kkondaka: Maybe the name is little misleading. I will rename it.

ps48 (approved):

Issue #6542, opened 2026-02-19
[RFC] New `otel_trace_genai` Processor


kylehounslow:
## 1. Context

OpenTelemetry recently stabilized [GenAI Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) (v1.39.0), defining standard span attributes for AI/LLM workloads: `gen_ai.system`, `gen_ai.operation.name`, `gen_ai.request.model`, `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`, etc.

Adoption is growing across instrumentation libraries. The official [OTel GenAI instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai) covers 7 libraries. Third-party libraries like [OpenLLMetry](https://github.com/traceloop/openllmetry) and [OpenInference](https://github.com/Arize-ai/openinference) cover 30+ each, but use their own attribute naming (e.g. `llm.usage.prompt_tokens` vs `llm.token_count.prompt` for the same concept). There is an ongoing community effort to converge on the OTel standard, including a [proposed OTel Collector processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46069) for attribute normalization.

As these traces flow into OpenSearch via Data Prepper, three problems emerge that can't be solved at the collector or SDK layer alone.

### Problem 1: Root spans lack GenAI attributes

Agent frameworks like [LangGraph](https://github.com/langchain-ai/langgraph), [CrewAI](https://github.com/crewAIInc/crewAI), and [OpenAI Agents SDK](https://github.com/openai/openai-agents-python) place `gen_ai.*` attributes on child spans (the actual LLM calls), not on the root span (the agent entry point):

```
Root: LangGraph.workflow          <-- no gen_ai.* attributes
  └── Child: ChatBedrock.chat     <-- gen_ai.system, gen_ai.usage.input_tokens, etc.
  └── Child: tool_call
  └── Child: ChatBedrock.chat     <-- gen_ai.system, gen_ai.usage.output_tokens, etc.
```

[...]

PR #6548, opened 2026-02-20
feat: Add GenAI agent trace enrichment to otel_traces processor

kylehounslow:
## Description

Adds GenAI agent trace enrichment to the existing `otel_traces` processor, as proposed in [RFC opensearch-project/data-prepper#6542](https://github.com/opensearch-project/data-prepper/issues/6542) with comments feedback applied (e.g. from [here](https://github.com/opensearch-project/data-prepper/issues/6542#issuecomment-3943071289) and [here](https://github.com/opensearch-project/data-prepper/issues/6542#issuecomment-3945992862))

### Some background context
Agent frameworks like [LangGraph](https://github.com/langchain-ai/langgraph) and [CrewAI](https://github.com/crewAIInc/crewAI) place `gen_ai.*` attributes on child spans (the LLM calls), not on the root span. This makes it impossible to identify GenAI traces at the root level in OpenSearch Dashboards. Token aggregation is technically possible at query time by traversing child spans, but becomes expensive at scale.

Separately, some instrumentation libraries (notably [OpenInference's Strands processor](https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-strands-agents/src/openinference/instrumentation/strands_agents/processor.py#L395-L400)) emit the same field as both a JSON string and flattened indexed sub-keys, which causes OpenSearch to reject the document due to field type conflicts. This is a known issue ([traceloop/openllmetry#1455](https://github.com/traceloop/openllmetry/issues/1455)).

See the RFC for full problem statement and design rationale.

### Summary of changes
The enrichment runs as a post-processing step in `doExecute` after all spans are collected. It does three things:

1. Propagate select GenAI attributes (`gen_ai.system`, `gen_ai.operation.name`, `gen_ai.provider.name`, `gen_ai.agent.name`, `gen_ai.request.model`) from child spans to root span (if root doesn't already have them)
2. Sum `gen_ai.usage.input_tokens` and `gen_ai.usage.output_tokens` across child spans and write totals to root (if root doesn't already have them)
3. Strip flattened sub-keys (e.g. `llm.input_messages.0.message.role`) when a parent string  value (e.g.  `llm.input_messages="..."`) exists on the same span, to prevent OpenSearch mapping conflicts. If only sub-keys exist (no parent), they are preserved to avoid data loss.

#### Files changed

- `OTelTraceRawProcessor.java` — 3 lines added: call `GenAiEnrichmentHelper.enrichBatch()` in `doExecute`
- `GenAiEnrichmentHelper.java` — New helper: groups batch by traceId, propagates select GenAI attributes + aggregates tokens from children to root, strips conflicting flattened sub-keys
- `GenAiEnrichmentTest.java` — 6 tests: system propagation, no-overwrite, token aggregation, token skip, sub-key stripping, non-genai passthrough
- `src/test/resources/` (13 JSON fixtures) — Span fixtures for root/child/token/flattened-subkey/plain-trace scenarios

### Why this approach
Per [feedback from @vamsimanohar](https://github.com/opensearch-project/data-prepper/issues/6542#issuecomment-3931039764), `otel_traces` already holds traces in memory and is built for enhancing the OSD UI. Adding enrichment here avoids a second processor in the pipeline. Per [feedback from @dlvenable](https://github.com/opensearch-project/data-prepper/issues/6542#issuecomment-3936791498), no flag is needed since the enrichment is a no-op for non-GenAI traces.

### Testing
1. Added unit test coverage. All existing + new tests passing.
2. Tested end-to-end with [LangGraph](https://github.com/langchain-ai/langgraph), [Strands](https://github.com/strands-agents/sdk-python), and [CrewAI](https://github.com/crewAIInc/crewAI) agent traces against a local OpenSearch cluster. CrewAI traces were previously silently dropped due to the mapping conflict and now index successfully.

### Known limitations
- In-memory batch only. Distributed deployments (OSIS) where spans land on different nodes would need an OpenSearch lookup fallback, similar to `otel_trace_group`. For typical agent traces (<3 min), same-batch delivery is the common case.
- No attribute normalization yet. Vendor-specific attributes (e.g. `llm.usage.prompt_tokens`) are not renamed to `gen_ai.*` semconv. Planned follow-up per the RFC.

### Issues Resolved
RFC: https://github.com/opensearch-project/data-prepper/issues/6542

### Check List
  - https://github.com/opensearch-project/documentation-website/issues/11976

### DIFF
--- a/data-prepper-plugins/otel-trace-raw-processor/src/main/resources/genai-attribute-mappings.yaml
+++ b/data-prepper-plugins/otel-trace-raw-processor/src/main/resources/genai-attribute-mappings.yaml
@@ -1,3 +1,12 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
+
 # GenAI attribute mappings from vendor-specific instrumentation libraries to OTel GenAI Semantic Conventions.
 # Each entry maps a source attribute key to a target gen_ai.* key.
 # Optional: wrap_as_array: true wraps a scalar string value into a single-element JSON array string.


data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiEnrichmentHelper.java
```diff
@@ -0,0 +1,222 @@
...
+                    children.add(span);
+                }
+            }
+            if (rootSpan != null && !children.isEmpty()) {
+                enrichRootSpan(rootSpan, children);
+            }
+        }
+    }
+
+    private static boolean isRootSpan(final Span span) {
```
dlvenable, line 82: I believe we have this logic elsewhere in the common OTel project.

data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiAttributeMappings.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /** Value mappings for gen_ai.operation.name (case-insensitive). */
+    static final Map<String, String> OPERATION_NAME_VALUES = buildOperationNameValues();
+
+    private static Map<String, MappingTarget> buildLookupTable() {
+        final Map<String, MappingTarget> table = new HashMap<>();
+
+        // --- OpenInference (Arize) ---
+        // https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md
+        table.put("llm.token_count.prompt", new MappingTarget("gen_ai.usage.input_tokens"));
```
dlvenable: These constants look like the kind of thing that can grow over time and may not be exhaustive. We should move this into a configuration file that is retained in the Java jar (resources). This class would have unit tests to verify the file reading aspect.

Additionally, we should eventually allow customers to add additional mapping files. We do something similar with Grok.

data-prepper-plugins/otel-trace-raw-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiEnrichmentTest.java
```diff
@@ -0,0 +1,244 @@
...
+        assertEquals(500, ((Number) root.getAttributes().get("gen_ai.usage.input_tokens")).intValue());
+        assertEquals(200, ((Number) root.getAttributes().get("gen_ai.usage.output_tokens")).intValue());
+    }
+
+    @Test
+    void testFlattenedSubkeysStripped() {
+        final Collection<Record<Span>> result = processor.doExecute(records(
+                "genai-flattened-subkeys-root-span.json", "genai-flattened-subkeys-child-span.json"));
+
+        for (final Record<Span> record : result) {
```
dlvenable, line 101: Assert the size before the loop or this assertion may have no affect.

data-prepper-plugins/otel-trace-raw-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiEnrichmentTest.java
```diff
@@ -0,0 +1,244 @@
...
+            final Map<String, Object> attrs = record.getData().getAttributes();
+            for (final String key : attrs.keySet()) {
```
dlvenable, line 104: Assert the size before the loop or this assertion may have no affect.

data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiAttributeMappings.java
```diff
@@ -0,0 +1,119 @@
...
+        MappingTarget(final String key, final boolean wrapSlice) {
+            this.key = key;
+            this.wrapSlice = wrapSlice;
+        }
+    }
+
+    private GenAiAttributeMappings() {}
+
+    /** Combined lookup table for all profiles. */
+    static final Map<String, MappingTarget> LOOKUP_TABLE = buildLookupTable();
```
dlvenable: Expose these as package-protected get methods.

data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiAttributeMappings.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Map;
+
+/**
+ * Attribute mappings from vendor-specific instrumentation libraries to OTel GenAI Semantic Conventions v1.39.0.
+ * Ported from <a href="https://github.com/kylehounslow/genainormalizer">genainormalizer</a>.
+ */
+final class GenAiAttributeMappings {
+
+    static final class MappingTarget {
+        final String key;
```
dlvenable: These should be private and exposed as package protected getters.

dlvenable:
@kylehounslow , Have you tested this with multiple Data Prepper nodes using peer forwarder?

data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/OTelTraceRawProcessor.java
```diff
@@ -99,6 +99,9 @@ public Collection<Record<Span>> doExecute(Collection<Record<Span>> records) {
 
         processedSpans.addAll(getTracesToFlushByGarbageCollection());
 
+        // Enrich GenAI agent traces (propagate select gen_ai attributes to root, aggregate tokens, strip conflicting sub-keys)
+        GenAiEnrichmentHelper.enrichBatch(processedSpans);
```
dlvenable, line 103: A unit test on this behavior would be good. It could use static mocking to verify that this helper was called. No need to get complicated on the results since you test that already.

kylehounslow:
> @kylehounslow , Have you tested this with multiple Data Prepper nodes using peer forwarder?

@dlvenable I have not! Will set this up and post the results here

data-prepper-plugins/otel-trace-raw-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiEnrichmentTest.java
kylehounslow: Good catch. Added assertions on `result` and `attrs` being non-empty before the loop so the test can't silently pass when the collection is empty.

data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiAttributeMappings.java
kylehounslow: Agreed, done. Made `MappingTarget.key` and `wrapSlice` private with `getKey()` and `isWrapSlice()` getters, and `LOOKUP_TABLE`/`OPERATION_NAME_VALUES` private with static getters. Added `GenAiAttributeMappingsTest` with direct coverage of the getters and key mappings.

data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiAttributeMappings.java
kylehounslow: Thanks for the Grok reference. I've applied the same approach and separated into config yaml.

data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiEnrichmentHelper.java
```diff
@@ -0,0 +1,248 @@
...
+        final String parentSpanId = span.getParentSpanId();
+        return parentSpanId == null || parentSpanId.isEmpty()
+                || "0000000000000000".equals(parentSpanId);
```
kkondaka, line 106: Maybe this can be done in a follow-up PR. Please make this a static constant like 
```
static String ZERO_SPAN_ID="0000000000000000"
```
and then
```
ZERO_SPAN_ID.equals(parentSpanId)
```

data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiEnrichmentHelper.java
```diff
@@ -0,0 +1,248 @@
...
+        for (final Span span : spans) {
+            spansByTrace.computeIfAbsent(span.getTraceId(), k -> new ArrayList<>()).add(span);
+        }
+
+        for (final List<Span> traceSpans : spansByTrace.values()) {
+            Span rootSpan = null;
+            final List<Span> children = new ArrayList<>();
+            for (final Span span : traceSpans) {
+                if (isRootSpan(span)) {
+                    rootSpan = span;
```
kkondaka, line 92: Is it expected that there will be only one root span?

data-prepper-plugins/otel-trace-raw-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiEnrichmentHelper.java
```diff
@@ -0,0 +1,248 @@
...
+            "gen_ai.prompt", "gen_ai.completion"
+    };
+
+    private GenAiEnrichmentHelper() {}
+
+    private static final String STORAGE_PREFIX = "span.attributes.";
+
+    /**
+     * Converts a logical dot-notation key (e.g. "gen_ai.system") to the storage format
+     * used by OTelProtoOpensearchCodec in the span attributes map (e.g. "span.attributes.gen_ai@system").
```
kkondaka, line 55: We have two modes - Opensearch  and OTEL standard (`OTelProtoOpensearchCodec` and `OTelProtoStandardCodec`) In the OTEL standard codec, we do not use "@". Does this feature only work with OpensearchCodec, if yes, should we check that the mode is `OTelProtoOpensearch`?

data-prepper-plugins/otel-trace-raw-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/oteltrace/GenAiEnrichmentTest.java
```diff
@@ -0,0 +1,288 @@
...
+    @BeforeEach
+    void setup() {
+        final OtelTraceRawProcessorConfig config = mock(OtelTraceRawProcessorConfig.class);
+        when(config.getTraceFlushIntervalSeconds()).thenReturn(180L);
+        when(config.getTraceGroupCacheMaxSize()).thenReturn(OtelTraceRawProcessorConfig.MAX_TRACE_ID_CACHE_SIZE);
+        when(config.getTraceGroupCacheTimeToLive()).thenReturn(OtelTraceRawProcessorConfig.DEFAULT_TRACE_ID_TTL);
+
+        final PipelineDescription pipelineDescription = mock(PipelineDescription.class);
+
+        processor = new OTelTraceRawProcessor(config, pipelineDescription, mock(PluginMetrics.class));
```
kkondaka, line 53: Usually, unit tests do not instantiate the processor and call doExecute. They should be just focused on GenAiEnrichmentHelper class level. These tests come under Integration tests.

kylehounslow:
FYI there's a README follow-up PR here: https://github.com/opensearch-project/data-prepper/pull/6570

PR #6554, opened 2026-02-23
Add Iceberg CDC source plugin

### LINKED ISSUE #6552: Add Iceberg source plugin for CDC ingestion
[Apache Iceberg](https://iceberg.apache.org/) is an open table format widely used for lakehouse on storage like S3, GCS, and HDFS. Iceberg tables are written by engines like Spark, Flink, and Trino, and are commonly used as the central store for analytical data.

There is currently no way to capture changes from Iceberg tables and ingest them into OpenSearch through Data Prepper. Users who want to keep OpenSearch indices synchronized with their Iceberg tables have to build custom pipelines involving Spark or Flink clusters, which adds significant operational complexity for what is essentially a data movement task.

A typical use case is when an organization uses an Iceberg-based lakehouse as the single source of truth for their data platform, and wants to continuously sync a subset of that data into OpenSearch for search and dashboarding.

An Iceberg source plugin for Data Prepper that detects changes in Iceberg tables and emits them as CDC events (INSERT, DELETE, UPDATE) into the pipeline.

Iceberg tracks all table changes as immutable snapshots. Each write operation (insert, update, delete) creates a new snapshot. Iceberg provides a core Java API called `IncrementalChangelogScan` that can compute the row-level diff between two snapshots, returning which rows were added, deleted, or updated. This API is part of the Iceberg core library and does not require Spark or any other compute engine.

The plugin would work as follows.

1. Periodically poll the Iceberg table for new snapshots
2. When a new snapshot is detected, run `IncrementalChangelogScan` from the last processed snapshot to the current one
3. Read the changed rows from the underlying data files (Parquet/ORC on S3 etc.)
4. Emit each changed row as a Data Prepper event with the operation type (INSERT/DELETE/UPDATE)
5. Checkpoint the last processed snapshot ID for resumability

Running a Spark or Flink job to read Iceberg's changelog and write to OpenSearch. While technically possible, this requires setting up and maintaining a separate compute cluster, and implementing the changelog reading, transformation, and OpenSearch writing logic from scratch. This level of complexity is a significant barrier for most users who simply want to keep an OpenSearch index in sync with an Iceberg table.

**Additional context**
I am willing to contribute this plugin. I plan to start with a PoC using the Iceberg core Java API to validate the approach.

One limitation to note is that Iceberg's changelog scan currently only supports copy-on-write (CoW) tables. Support for merge-on-read (MoR) tables is being actively worked on in the Iceberg project (apache/iceberg#14264). The plugin can initially support CoW tables and add MoR support once that work is merged.

lawofcycles:
### Description
Adds a new `iceberg` source plugin that captures row-level changes (CDC) from Apache Iceberg tables and ingests them into the Data Prepper pipeline. The plugin uses Iceberg's `IncrementalChangelogScan` API directly, without requiring Spark or Flink.

Key features
- Polls Iceberg tables for new snapshots and processes row-level changes (INSERT, UPDATE, DELETE) incrementally
- Supports initial load (full table scan) with automatic transition to CDC
- Handles CoW carryover removal for correct change detection
- Distributes work across Data Prepper nodes via `EnhancedSourceCoordinator`
- Works with any Iceberg catalog (Glue, REST, Hive, Hadoop, Nessie, JDBC, etc.)

Current limitations
- Only Copy-on-Write (CoW) tables are supported. MoR tables are rejected at startup.
- `identifier_columns` must be configured by the user for CDC correctness
- When multiple DELETED and ADDED files exist within the same Iceberg partition and bounds-based pairing fails, work for that partition cannot be distributed across nodes.

Catalog support: Any catalog supported by Iceberg's `CatalogUtil.buildIcebergCatalog()` (Glue, REST, Hive, Hadoop, Nessie, JDBC, etc.)
 
### Issues Resolved
Resolves #6552 
 
### Check List

### DIFF
--- a/data-prepper-plugins/iceberg-source/src/test/java/org/opensearch/dataprepper/plugins/source/iceberg/worker/ChangelogRecordConverterTest.java
+++ b/data-prepper-plugins/iceberg-source/src/test/java/org/opensearch/dataprepper/plugins/source/iceberg/worker/ChangelogRecordConverterTest.java
@@ -19,12 +19,10 @@ import org.apache.iceberg.variants.Variant;
 import org.apache.iceberg.variants.VariantArray;
 import org.apache.iceberg.variants.VariantObject;
 import org.apache.iceberg.variants.VariantPrimitive;
-import org.apache.iceberg.variants.VariantValue;
 import org.junit.jupiter.api.Test;
 import org.opensearch.dataprepper.model.event.Event;
 
 import java.math.BigDecimal;
-import java.nio.ByteBuffer;
 import java.time.LocalDate;
 import java.util.List;
 import java.util.Map;


lawofcycles:
### Discussion: PeerForwarder extension for Source-layer shuffle

The current implementation groups DELETED and ADDED files together at the Source layer so that carryover removal can happen locally on each node. This works well when bounds-based pairing succeeds, but falls back to single-node processing when it fails (e.g., when an UPDATE changes a column's min/max bounds).

A more scalable approach would be to shuffle rows between ChangelogWorkers by a hash of all data columns (similar to Spark's repartition). Data Prepper's PeerForwarder infrastructure provides useful building blocks for this (consistent hashing via HashRing, node discovery, Armeria HTTP transport). However, the PeerForwarder's orchestration layer is tightly coupled to the Processor execution model (synchronous send/receive within a batch processing cycle, ReceiveBuffer integration with the Pipeline, pluginId-based routing), so Source-layer shuffling would likely need to build its own send/receive flow on top of these lower-level components, or the PeerForwarder architecture would need to be generalized to support non-Processor use cases.

[RFC #700 (Core Peer Forwarding)](https://github.com/opensearch-project/data-prepper/issues/700) notes that the initial implementation targets Processor plugins only, and suggests opening a new issue to expand the functionality to Source or Sink plugins. I plan to open a separate issue proposing this extension, with Iceberg CDC as the first use case.

I'd like to hear the maintainers' thoughts on the best approach for inter-node data exchange at the Source layer.

**20260311 UPDATE: Additional benefit of Source-layer shuffle**

Source-layer shuffle would also enable processing multiple snapshots in a single `IncrementalChangelogScan` call, rather than the current one-snapshot-at-a-time serial approach.

Currently, the LeaderScheduler processes each snapshot sequentially: plan one snapshot, wait for all partitions to complete (including sink acknowledgement when `acknowledgments: true`), then move to the next. This is necessary because without shuffle, carryover removal relies on bounds-based file pairing within each snapshot. With shuffle (repartition by hash of all data columns), carryover removal works correctly across snapshot boundaries, just as Spark's `RemoveCarryoverIterator` does after repartition.

The performance benefits of batch processing multiple snapshots include following points.

1. Eliminating per-snapshot synchronization overhead. With acknowledgments enabled, the leader must wait for the sink to finish writing and return acknowledgments before planning the next snapshot. Batch processing reduces N sequential wait cycles to one.
2. Reducing scan planning overhead. One `IncrementalChangelogScan` call reads manifests once instead of N times.

This is particularly relevant for tables with frequent commits (e.g., Spark Streaming committing every few minutes), where dozens of snapshots can accumulate between polling intervals.

Note that batch processing also requires ordering writes by `changeOrdinal` (the snapshot sequence number that `IncrementalChangelogScan` assigns to each task) to ensure correct operation order in OpenSearch when the same document is modified across multiple snapshots.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/TableConfig.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Map;
+
+public class TableConfig {
+
+    @JsonProperty("table_name")
+    @NotBlank
+    private String tableName;
+
+    @JsonProperty("catalog")
+    private Map<String, String> catalog = Collections.emptyMap();
```
dlvenable, line 27: Is there a model that we can use here instead of a map? The models provide the best experience for the community because they promote configuration consistency and help with documentation.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/TableConfig.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @JsonProperty("identifier_columns")
+    private List<String> identifierColumns = Collections.emptyList();
+
+    @JsonProperty("initial_load")
+    private boolean initialLoad = true;
```
dlvenable:
```suggestion
    @JsonProperty("disable_export")
    private boolean disableExport = false;
```

Our other CDC sources have both `export` and `stream` configurations. And we allow the following combinations:

* Export and stream
* Export only
* Stream only

Unless "initial load" is a specialized term in Iceberg we should keep our existing conventions.

Also, we prefer to default boolean values to false.

We may eventually want to add `disable_stream` but that is not necessary now.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/ChangelogTaskPartition.java
```diff
@@ -0,0 +1,49 @@
+/*
```
dlvenable, line 1: Please use the updated license headers.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/state/ChangelogTaskProgressState.java
```diff
@@ -0,0 +1,79 @@
...
+
+package org.opensearch.dataprepper.plugins.source.iceberg.coordination.state;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+public class ChangelogTaskProgressState {
+
+    @JsonProperty("snapshot_id")
```
dlvenable: These save to DynamoDB for source coordination. We use camelCase for these. Please update accordingly.

https://github.com/opensearch-project/data-prepper/blob/97cd930be30b74730dffc122b3f45092d6a8c6c3/data-prepper-plugins/dynamodb-source/src/main/java/org/opensearch/dataprepper/plugins/source/dynamodb/coordination/state/ExportProgressState.java#L12-L28

data-prepper-plugins/iceberg-source/build.gradle
```diff
@@ -0,0 +1,28 @@
+plugins {
```
dlvenable: You don't need these three lines. You can remove them.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergSource.java
```diff
@@ -0,0 +1,83 @@
...
+import org.opensearch.dataprepper.model.source.coordinator.enhanced.UsesEnhancedSourceCoordination;
+import org.opensearch.dataprepper.plugins.source.iceberg.coordination.PartitionFactory;
+import org.opensearch.dataprepper.plugins.source.iceberg.coordination.partition.LeaderPartition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Objects;
+import java.util.function.Function;
+
+@DataPrepperPlugin(name = "iceberg", pluginType = Source.class, pluginConfigurationType = IcebergSourceConfig.class)
```
dlvenable, line 35: Let's add the `@Experimental` annotation here. I'd like to get this PR merged, but we do have a few other items to address before making this fully supported.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/ChangelogTaskPartition.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private final String partitionKey;
+    private final ChangelogTaskProgressState state;
+
+    public ChangelogTaskPartition(final String partitionKey, final ChangelogTaskProgressState state) {
+        this.partitionKey = partitionKey;
+        this.state = state;
+    }
+
+    public ChangelogTaskPartition(final SourcePartitionStoreItem sourcePartitionStoreItem) {
```
dlvenable, line 31: This code should have unit tests. Verify the getter results.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/GlobalState.java
```diff
@@ -0,0 +1,51 @@
...
+    private final String stateName;
+    private Map<String, Object> state;
+
+    public GlobalState(final String stateName, final Map<String, Object> state) {
+        this.stateName = stateName;
+        this.state = state;
+    }
+
+    public GlobalState(final SourcePartitionStoreItem sourcePartitionStoreItem) {
+        setSourcePartitionStoreItem(sourcePartitionStoreItem);
```
dlvenable, line 30: This code should have unit tests. Verify the getter results.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/InitialLoadTaskPartition.java
```diff
@@ -0,0 +1,50 @@
...
+    private final String partitionKey;
+    private final InitialLoadTaskProgressState state;
+
+    public InitialLoadTaskPartition(final String partitionKey, final InitialLoadTaskProgressState state) {
+        this.partitionKey = partitionKey;
+        this.state = state;
+    }
+
+    public InitialLoadTaskPartition(final SourcePartitionStoreItem sourcePartitionStoreItem) {
+        setSourcePartitionStoreItem(sourcePartitionStoreItem);
```
dlvenable, line 32: This code should have unit tests. Verify the getter results.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/LeaderPartition.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private final LeaderProgressState state;
+
+    public LeaderPartition() {
+        this.state = new LeaderProgressState();
+        this.state.setInitialized(false);
+    }
+
+    public LeaderPartition(final SourcePartitionStoreItem sourcePartitionStoreItem) {
+        setSourcePartitionStoreItem(sourcePartitionStoreItem);
```
dlvenable, line 32: This code should have unit tests. Verify the getter results.

dlvenable (changes requested):
Thank you @lawofcycles for this great contribution!

I left a few initial comments. I want to look deeper at the logic as well.

One thing we will need before making this official is a set of integration tests that run against an actual Iceberg Docker image. We do this for our Kafka buffer and our OpenSearch sink.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/TableConfig.java
lawofcycles: I looked into this. Iceberg does not provide a typed model for catalog configuration. All official implementations pass catalog properties as a string map

* Java (reference): [`CatalogUtil.buildIcebergCatalog(String name, Map<String, String> options, Object conf)`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/CatalogUtil.java#L301)
* Python (PyIceberg): [`Properties = dict[str, Any]`](https://github.com/apache/iceberg-python/blob/main/pyiceberg/typedef.py#L91)
* Rust: [`HashMap<String, String>`](https://github.com/apache/iceberg-rust/blob/main/crates/iceberg/src/catalog/mod.rs#L152)
* Go: [`Properties = map[string]string`](https://github.com/apache/iceberg-go/blob/main/types.go#L37)

Kafka Connect's Iceberg integration also uses [`Map<String, String>`](https://github.com/apache/iceberg/blob/main/kafka-connect/kafka-connect/src/main/java/org/apache/iceberg/connect/IcebergSinkConfig.java#L306) in the same way.

I think this is an intentional design choice because each catalog type (Glue, REST, Hive, Nessie, JDBC, etc.) accepts a completely different set of properties. A typed model would either couple the plugin to specific catalog types or still need a map fallback for catalog specific properties.

That said, the catalog properties will need thorough documentation with examples for each supported catalog type so that users have clear guidance on what to configure.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/TableConfig.java
lawofcycles: changed from `initialLoad ` to `disableExport`

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/ChangelogTaskPartition.java
lawofcycles: fixed!

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/LeaderPartition.java
lawofcycles: added unit test

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/ChangelogTaskPartition.java
lawofcycles: added unit test.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/InitialLoadTaskPartition.java
lawofcycles: added unit test

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/partition/GlobalState.java
lawofcycles: added unit test.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergSource.java
lawofcycles: added the @Experimental !

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/coordination/state/ChangelogTaskProgressState.java
lawofcycles: made them camel case.

data-prepper-plugins/iceberg-source/build.gradle
lawofcycles: removed.

lawofcycles:
@dlvenable Thank you for the review. Here is a summary of the updates.

**Review feedback**
All review comments have been addressed.

**Integration tests**
Added Docker based integration tests following the existing patterns. The tests use SeaweedFS (S3 compatible storage) and the Iceberg REST Catalog.

```sh
docker-compose -f data-prepper-plugins/iceberg-source/docker/docker-compose.yml up -d
./gradlew :data-prepper-plugins:iceberg-source:integrationTest
docker-compose -f data-prepper-plugins/iceberg-source/docker/docker-compose.yml down
```

**Code improvements**
Reviewed the code in detail and fixed several issues found during the personal review process.

**Manual verification**
Verified end to end with Iceberg tables on S3 (Glue Catalog + EMR Spark) writing CDC events to OpenSearch, covering multiple scenarios including INSERT, UPDATE, DELETE, and carryover removal.

I am changing this PR from Draft to Ready for Review.

dlvenable:
There are a couple of checkstyle errors:

```
Error: eckstyle] [ERROR] /home/runner/work/data-prepper/data-prepper/data-prepper-plugins/iceberg-source/src/test/java/org/opensearch/dataprepper/plugins/source/iceberg/worker/ChangelogRecordConverterTest.java:22:8: Unused import - org.apache.iceberg.variants.VariantValue. [UnusedImports]
Error: eckstyle] [ERROR] /home/runner/work/data-prepper/data-prepper/data-prepper-plugins/iceberg-source/src/test/java/org/opensearch/dataprepper/plugins/source/iceberg/worker/ChangelogRecordConverterTest.java:27:8: Unused import - java.nio.ByteBuffer. [UnusedImports]
```

Build results:
https://github.com/opensearch-project/data-prepper/actions/runs/22817460047/job/66184697244?pr=6554

lawofcycles:
Found and fixed a correctness issue with UPDATE event handling.
https://github.com/opensearch-project/data-prepper/pull/6554/changes/094e5a274ffff12076f3827222f1fa03bf7a6c2b

Previously, a CoW UPDATE produced two separate events after carryover removal: DELETE (old row) followed by INDEX (new row) for the same `document_id`. This relied on the DELETE arriving at OpenSearch before the INDEX. However, Data Prepper's pipeline uses multiple ProcessWorker threads that consume from the buffer in parallel, so these two events could end up in different bulk requests with no ordering guarantee. If the INDEX arrived first and the DELETE second, the document would be deleted and lost.

The fix merges UPDATE pairs into a single event. After carryover removal, when `identifier_columns` is configured, if a DELETE and INSERT produce the same `document_id`, only the INSERT is emitted as INDEX. Since OpenSearch INDEX is an upsert, the DELETE is unnecessary. Pure deletes (DELETE with no matching INSERT) are unaffected. When `identifier_columns` is not configured, no `document_id` is generated and the behavior is unchanged.

lawofcycles:
> There are a couple of checkstyle errors:

Fixed https://github.com/opensearch-project/data-prepper/pull/6554/commits/3dc4a7de9bd48b740a46fcc1c6bd37fd0808b46a

dlvenable (approved):

dlvenable:
> ### Discussion: PeerForwarder extension for Source-layer shuffle
> 
> The current implementation groups DELETED and ADDED files together at the Source layer so that carryover removal can happen locally on each node. This works well when bounds-based pairing succeeds, but falls back to single-node processing when it fails (e.g., when an UPDATE changes a column's min/max bounds).
> 
> A more scalable approach would be to shuffle rows between ChangelogWorkers by a hash of all data columns (similar to Spark's repartition). Data Prepper's PeerForwarder infrastructure provides useful building blocks for this (consistent hashing via HashRing, node discovery, Armeria HTTP transport). However, the PeerForwarder's orchestration layer is tightly coupled to the Processor execution model (synchronous send/receive within a batch processing cycle, ReceiveBuffer integration with the Pipeline, pluginId-based routing), so Source-layer shuffling would likely need to build its own send/receive flow on top of these lower-level components, or the PeerForwarder architecture would need to be generalized to support non-Processor use cases.
> 
> [RFC #700 (Core Peer Forwarding)](https://github.com/opensearch-project/data-prepper/issues/700) notes that the initial implementation targets Processor plugins only, and suggests opening a new issue to expand the functionality to Source or Sink plugins. I plan to open a separate issue proposing this extension, with Iceberg CDC as the first use case.
> 
> I'd like to hear the maintainers' thoughts on the best approach for inter-node data exchange at the Source layer.
> 
> **20260311 UPDATE: Additional benefit of Source-layer shuffle**
> 
> Source-layer shuffle would also enable processing multiple snapshots in a single `IncrementalChangelogScan` call, rather than the current one-snapshot-at-a-time serial approach.
> 
> Currently, the LeaderScheduler processes each snapshot sequentially: plan one snapshot, wait for all partitions to complete (including sink acknowledgement when `acknowledgments: true`), then move to the next. This is necessary because without shuffle, carryover removal relies on bounds-based file pairing within each snapshot. With shuffle (repartition by hash of all data columns), carryover removal works correctly across snapshot boundaries, just as Spark's `RemoveCarryoverIterator` does after repartition.
> 
> The performance benefits of batch processing multiple snapshots include following points.
> 
>     1. Eliminating per-snapshot synchronization overhead. With acknowledgments enabled, the leader must wait for the sink to finish writing and return acknowledgments before planning the next snapshot. Batch processing reduces N sequential wait cycles to one.
> 
>     2. Reducing scan planning overhead. One `IncrementalChangelogScan` call reads manifests once instead of N times.
> 
> 
> This is particularly relevant for tables with frequent commits (e.g., Spark Streaming committing every few minutes), where dozens of snapshots can accumulate between polling intervals.
> 
> Note that batch processing also requires ordering writes by `changeOrdinal` (the snapshot sequence number that `IncrementalChangelogScan` assigns to each task) to ensure correct operation order in OpenSearch when the same document is modified across multiple snapshots.

@lawofcycles , I agree that this would be a great approach. The processor approach was what we needed at the time.

We actually have some hard-coded, non peer-forwarder support for source forwarding. It is in our OTel traces pipeline only when used with the Kafka buffer. In that case, we put the traces on different partitions.

The scenarios you outlined as well as tracing are both good examples of how we could make use of source-based peer-forwarding. Note that we do not generally recommend using `kafka` buffer with pull-based pipelines. So peer-forwarding would probably be the best solution here.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/leader/LeaderScheduler.java
```diff
@@ -0,0 +1,307 @@
...
+
+            try (CloseableIterable<FileScanTask> tasks = scan.planFiles()) {
+                for (final FileScanTask task : tasks) {
+                    final InitialLoadTaskProgressState taskState = new InitialLoadTaskProgressState();
+                    taskState.setSnapshotId(snapshotId);
+                    taskState.setTableName(tableName);
+                    taskState.setDataFilePath(task.file().location());
+                    taskState.setTotalRecords(task.file().recordCount());
+
+                    final String partitionKey = tableName + "|initial|" + UUID.randomUUID();
```
dlvenable, line 146: Why do you use a UUID here? Could this cause duplicates?

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/leader/LeaderScheduler.java
```diff
@@ -0,0 +1,307 @@
...
+                }
+
+                // Create a completion tracker in GlobalState
+                final String completionKey = SNAPSHOT_COMPLETION_PREFIX + snapshot.snapshotId();
+                sourceCoordinator.createPartition(new GlobalState(completionKey,
+                        Map.of("total", taskGroups.size(), "completed", 0)));
+
+                for (final ChangelogTaskProgressState taskState : taskGroups) {
+                    final String partitionKey = tableName + "|" + snapshot.snapshotId()
+                            + "|" + UUID.randomUUID();
```
dlvenable, line 240: Why do you use a UUID here? Could this cause duplicates?

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/worker/ChangelogRecordConverter.java
```diff
@@ -0,0 +1,174 @@
...
+                         final org.apache.iceberg.Schema schema,
+                         final String operation,
+                         final long snapshotId) {
+        final Map<String, Object> data = new LinkedHashMap<>();
+        for (final Types.NestedField field : schema.columns()) {
+            final Object value = record.getField(field.name());
+            data.put(field.name(), convertValue(value, field.type()));
+        }
+
+        final Event event = JacksonEvent.builder()
```
dlvenable, line 64: We should move toward using the `EventFactory`.

Here is a simple example of usage:

https://github.com/opensearch-project/data-prepper/blob/ce57396b3998d753cbeb2fb576cf39c6d4bfd8ba/data-prepper-plugins/common/src/main/java/org/opensearch/dataprepper/plugins/source/file/FileSource.java#L139-L142

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/leader/TaskGrouper.java
```diff
@@ -0,0 +1,245 @@
...
+                state.setSnapshotId(toSnapshotId);
+                state.setTableName(tableName);
+                state.setDataFilePaths(group.stream().map(t -> t.filePath).collect(Collectors.toList()));
+                state.setTaskTypes(group.stream().map(t -> t.taskType).collect(Collectors.toList()));
+                state.setTotalRecords(group.stream().mapToLong(t -> t.recordCount).sum());
+                result.add(state);
+            }
+        }
+
+        LOG.info("Planned {} task group(s) for table {} (snapshot {} -> {})",
```
dlvenable, line 89: This could become noisy quickly. We should make this a `DEBUG` log.

Another option here is to use a DistributionSummary metric for this. This would allow you to analyze maximum task group sizes and average.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/worker/ChangelogWorker.java
```diff
@@ -0,0 +1,412 @@
...
+            sourceCoordinator.giveUpPartition(partition);
+            return;
+        }
+
+        final Schema schema = table.schema();
+        final ChangelogRecordConverter converter = new ChangelogRecordConverter(
+                tableName, tableConfig.getIdentifierColumns());
+        final CarryoverRemover carryoverRemover = new CarryoverRemover();
+
+        LOG.info("Processing partition for table {} snapshot {} with {} file(s)",
```
dlvenable, line 140: This should become a debug level log.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/worker/ChangelogWorker.java
```diff
@@ -0,0 +1,412 @@
...
+            for (int i = 0; i < allRows.size(); i++) {
+                final RowWithMeta row = allRows.get(i);
+                final List<Object> dataColumns = new ArrayList<>();
+                for (final Types.NestedField field : schema.columns()) {
+                    dataColumns.add(row.record.getField(field.name()));
+                }
+                changelogRows.add(new CarryoverRemover.ChangelogRow(dataColumns, row.operation, i));
+            }
+            survivingIndices = carryoverRemover.removeCarryover(changelogRows);
+            LOG.info("Carryover removal: {} rows -> {} rows", allRows.size(), survivingIndices.size());
```
dlvenable, line 175: This should be a debug log.

horovits:
Great work @lawofcycles, nice to see it merged

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/leader/LeaderScheduler.java
lawofcycles: The UUID was used for simplicity, but you are right that it can cause duplicate partitions if the leader crashes after creating partitions but before updating `lastProcessedSnapshotId`. Since `tryCreatePartitionItem` uses a conditional put that returns false for existing keys, switching to deterministic keys would prevent this.

For initial load: `tableName|initial|snapshotId|filePath` (one file per task, so the file path is unique).
For CDC: `tableName|snapshotId|sha256(sorted file paths)` (the hash of sorted file paths within each task group).

I will submit a follow up PR for this change.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/leader/LeaderScheduler.java
lawofcycles: Same as above. Will be addressed in the same follow up PR.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/worker/ChangelogRecordConverter.java
lawofcycles: Got it, raised follow up PR https://github.com/opensearch-project/data-prepper/pull/6641

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/leader/TaskGrouper.java
lawofcycles: I opened PR https://github.com/opensearch-project/data-prepper/pull/6648 to adress this point.

lawofcycles:
@dlvenable Thank you for the feedback and for sharing the OTel traces precedent. That is very helpful context.

I have been thinking about the implementation approach and would like to propose a two track plan.

**Track 1: Framework (RFC)**
I will open a separate RFC issue proposing a generalized Source layer peer forwarding extension to the existing PeerForwarder infrastructure. The design would extract the reusable building blocks (HashRing, PeerForwarderClient, PeerListProvider, Armeria transport) into a shared layer that both Processor and Source plugins can use, while keeping the orchestration logic (send/receive flow, buffering) pluggable per use case. This would also provide a path to replace the hard coded OTel traces source forwarding with the generalized mechanism.

**Track 2: Iceberg Source (plugin level implementation)**
In parallel, I will implement the shuffle logic directly within the iceberg source plugin using the same lower level components (HashRing for consistent hashing, Armeria HTTP for transport). This unblocks the Iceberg CDC improvements (fallback partition distribution, multi snapshot batch processing) without waiting for the framework RFC to be reviewed and merged.

**Convergence**
Once the framework lands, the Iceberg source plugin's shuffle implementation would be migrated to use the framework API. The plugin level code is designed with this migration in mind: the hashing and transport logic will be isolated behind an internal interface so that swapping to the framework API is straightforward.

This approach keeps the Iceberg CDC work moving forward while ensuring the framework design benefits from a concrete, working implementation as a reference.

Does this plan sound reasonable?

lawofcycles:
@dlvenable I need to correct my earlier proposal. After further analysis, I found that the PeerForwarder extension (Track 1) is not necessary.

I am now thinking the shuffle should use a pull based approach (similar to Spark's shuffle architecture) where writers partition data to local disk by hash key and readers pull their partitions from each node via HTTP. This would not require extending the PeerForwarder framework, which is push based.

I also identified a correctness bug in the current implementation: when a partition column is updated, the DELETE (old partition) and INSERT (new partition) for the same document end up in separate tasks. Since the UPDATE merge only operates within a single task, it cannot detect this cross partition update. The DELETE and INSERT are sent to the sink independently with no ordering guarantee, which can cause data loss if the DELETE arrives after the INSERT.

I will open a detailed issue with the revised design shortly.

lawofcycles:
I opened two follow up issues.

https://github.com/opensearch-project/data-prepper/issues/6666
https://github.com/opensearch-project/data-prepper/issues/6667

I believe #6666 is more critical.

PR #6572, opened 2026-02-27
Fix breaking changes in otel_logs_source

oeyh:
### Description
Commit 8d85609 (#6250) added HTTP service support to otel_logs_source but introduced regressions that break existing users:

  1. NPE on startup — configureHttpService() was called unconditionally, causing String.replace() to throw on a null http_path for any user who only configured a gRPC path.
  2. ACM certificate TLS silently broken — The SSL setup branch dropped the useAcmCertForSSL() check, so users relying on ACM-issued certificates would start without TLS.
  3. HTTP health check registered unconditionally — The old behavior gated the HTTP /health endpoint on enableUnframedRequests() being true; the new code registered it for all configs regardless.
  4. Auth provider split — The server-level HTTP decorator was loaded as an ArmeriaHttpAuthenticationProvider while the gRPC interceptor used a separate GrpcAuthenticationProvider, potentially
  resolving to different plugin implementations. The original code used a single GrpcAuthenticationProvider for both.
 
All changes are confined to OTelLogsSource.java:

  - Gate configureHttpService() on http_path being non-null — making HTTP service opt-in, consistent with how gRPC path works
  - Restore || useAcmCertForSSL() in the SSL condition
  - Restore the HTTP health check gate: registered only when enableUnframedRequests() or http_path is set (the latter ensures the new HTTP service also gets a health endpoint)
  - Hoist GrpcAuthenticationProvider creation into createServer, apply its getHttpAuthenticationService() as the server-level decorator, and pass the same instance into configureGrpcService —
  removing the split-provider pattern and the now-dead createHttpAuthentication() helper

### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/data-prepper-plugins/otel-logs-source/src/test/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSourceGrpcTest.java
+++ b/data-prepper-plugins/otel-logs-source/src/test/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSourceGrpcTest.java
@@ -113,6 +113,7 @@ import static org.opensearch.dataprepper.plugins.source.otellogs.OTelLogsSourceC
 import static org.opensearch.dataprepper.plugins.source.otellogs.OtelLogsSourceConfigFixture.createBuilderForConfigWithAcmeSsl;
 import static org.opensearch.dataprepper.plugins.source.otellogs.OtelLogsSourceConfigFixture.createConfigBuilderWithBasicAuth;
 import static org.opensearch.dataprepper.plugins.source.otellogs.OtelLogsSourceConfigFixture.createDefaultConfig;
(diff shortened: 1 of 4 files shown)


data-prepper-plugins/otel-logs-source/src/test/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSourceGrpcTest.java
```diff
@@ -340,6 +341,18 @@ void testStartWithEmptyBuffer() {
         assertThrows(IllegalStateException.class, () -> source.start(null));
     }
 
+    @Test
+    void start_withoutHttpPath_doesNotThrowNPE() {
+        final OTelLogsSourceConfig config = createDefaultConfigBuilder()
+                .httpPath(null)
```
dlvenable, line 347: I think we should invert this. Most tests should set `httpPath` to `null` and we only have one that sets it.

I'm fine with merging this fix now, but think we should update in a quick follow-on.

PR #6593, opened 2026-03-02
Adding support for pipeline DLQ to SQS sink

divbok:
### Description
Add support for pipeline DLQ to SQS sink. When "dlq_pipeline" is present in the configuration, SQS sink uses that for DLQ and ignores the sink level dlq config, if present.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkService.java
+++ b/data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkService.java
@@ -12,6 +12,9 @@ package org.opensearch.dataprepper.plugins.sink.sqs;
 import org.opensearch.dataprepper.metrics.PluginMetrics;
 import org.opensearch.dataprepper.expression.ExpressionEvaluator;
 import org.opensearch.dataprepper.model.codec.OutputCodec;
+import org.opensearch.dataprepper.model.configuration.PipelineDescription;
+import org.opensearch.dataprepper.model.configuration.PluginSetting;
+import org.opensearch.dataprepper.model.pipeline.HeadlessPipeline;
 import org.opensearch.dataprepper.model.sink.OutputCodecContext;
 import org.opensearch.dataprepper.plugins.accumulator.BufferFactory;
 import org.opensearch.dataprepper.plugins.accumulator.InMemoryBufferFactory;
@@ -62,6 +65,10 @@ public class SqsSinkService extends SqsSinkExecutor {
     private final SqsSinkMetrics sinkMetrics;
     private final DlqPushHandler dlqPushHandler;
     private final List<DlqObject> dlqObjects;
+    private HeadlessPipeline dlqPipeline;
+    private final List<Record<Event>> dlqPipelineRecords;
+    private final PluginSetting pluginSetting;
+    private final PipelineDescription pipelineDescription;
 
     public SqsSinkService(final SqsSinkConfig sqsSinkConfig,
                           final SqsClient sqsClient,
@@ -69,9 +76,13 @@ public class SqsSinkService extends SqsSinkExecutor {
                           final OutputCodec codec,
                           final SinkContext sinkContext,
                           final DlqPushHandler dlqPushHandler,
-                          final PluginMetrics pluginMetrics) {
+                          final PluginMetrics pluginMetrics,
+                          final PluginSetting pluginSetting,
+                          final PipelineDescription pipelineDescription
+                          ) {
         batchUrlMap = new HashMap<>();
         dlqObjects = new ArrayList<>();
+        dlqPipelineRecords = new ArrayList<>();
         inMemoryBufferFactory =new InMemoryBufferFactory();
         this.sqsClient = sqsClient;
         this.dlqPushHandler = dlqPushHandler;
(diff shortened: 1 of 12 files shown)


data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkService.java
```diff
@@ -286,10 +308,24 @@ public boolean exceedsFlushTimeInterval() {
...
+        addMessageToDLQ(batchEntry.getBody(), batchEntry.getEventHandles(), batchEntry.getEvents(), errorMessage);
     }
 
-    private void addMessageToDLQ(final String message, final List<EventHandle> eventHandles, final String errorMessage) {
+    private void addMessageToDLQ(final String message, final List<EventHandle> eventHandles, List<Event> events, final String errorMessage) {
+        if (dlqPipeline != null) {
+            for (Event event: events) {
+                if (event != null) {
+                   event.updateFailureMetadata()
+                       .with("pluginId", pluginSetting.getName())
```
dlvenable: Some of these look like they would be the same as the OpenSearch sink. We should have constants in the `FailureMetadata` class for them. Even better would be `withPluginName` methods.

data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkService.java
```diff
@@ -286,10 +308,24 @@ public boolean exceedsFlushTimeInterval() {
...
+                       .with("pluginName", pluginSetting.getName())
+                       .with("pipelineName", pipelineDescription.getPipelineName())
+                       .with("sqsSinkQueueUrl", queueUrl)
+                       .with("message", errorMessage);
```
dlvenable: Is `message` the same key we use in the OpenSearch sink?

It was probably ill advised. Now we can see it is confusing because `Message` is what SQS calls inputs!

data-prepper-plugins/sqs-sink/src/test/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkServiceTest.java
```diff
@@ -520,6 +532,83 @@ void TestWithOneBatch_RetryFlushes() throws Exception {
...
+        SqsSinkService sqsSinkService = createObjectUnderTest();
+        sqsSinkService.setDlqPipeline(dlqPipeline);
+        Record<Event> record = getLargeRecord(300 * 1024);
+        sqsSinkService.execute(List.of(record));
+
+        verify(dlqPipeline, times(1)).sendEvents(any());
+        assertThat(capturedRecords.size(), equalTo(1));
+
+        Event event = capturedRecords.get(0).getData();
+        assertThat(event.get("_failure_metadata/pluginId", String.class), equalTo("name"));
```
dlvenable: Please use random values for the values. `"name", `"pipeline"` may easily be magic values.

data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkService.java
divbok: Should we change this to errorMessage instead?

kkondaka: Yes, we should make it errorMessage in all places.

kkondaka: @dlvenable should have `withErrorMessage()` , `withPluginName()`, `withPipelineName()` to `EventFailureMetadata` class?

data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkService.java
```diff
@@ -286,10 +308,24 @@ public boolean exceedsFlushTimeInterval() {
...
+    private void addMessageToDLQ(final String message, final List<EventHandle> eventHandles, List<Event> events, final String errorMessage) {
+        if (dlqPipeline != null) {
+            for (Event event: events) {
+                if (event != null) {
+                   event.updateFailureMetadata()
+                       .withPluginId(pluginSetting.getName())
+                       .withPluginName(pluginSetting.getName())
+                       .withPipelineName(pipelineDescription.getPipelineName())
+                       .withErrorMessage(errorMessage)
+                       .with("sqsSinkQueueUrl", queueUrl);
```
graytaylor0, line 323: .with() is kind of weird, would be clearer with something like "withAttribute". I know that isn't added in this PR though.

One nit: You only use it once but would probably just make "sqsSinkQueueUrl" a static variable on the class.

graytaylor0 (approved):

data-prepper-plugins/sqs-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/sqs/SqsSinkService.java
divbok: Will makes these changes in a followup PR

PR #6596, opened 2026-03-02
Add support for invoking acknowledgmentSet callback on expiry

kkondaka:
### Description
Add support for invoking acknowledgmentSet callback on expiry
1. Supports creating acknowledgmentSet with option to invoke the callback upon expiry
2. Invokes acknowledgementSet callback function with "false" as the result
3. Provides config option in kafka source and kafka buffer to configure acknowledgements timeout
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/buffer/KafkaBufferConfig.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/buffer/KafkaBufferConfig.java
@@ -1,6 +1,10 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  */
 
 package org.opensearch.dataprepper.plugins.kafka.buffer;
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
@@ -1,7 +1,12 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  */
+
 package org.opensearch.dataprepper.plugins.kafka.consumer;
 
 import com.amazonaws.services.schemaregistry.serializers.json.JsonDataWithSchema;
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/buffer/BufferTopicConfig.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/buffer/BufferTopicConfig.java
@@ -1,6 +1,10 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  */
 
 package org.opensearch.dataprepper.plugins.kafka.buffer;
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/buffer/KafkaBuffer.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/buffer/KafkaBuffer.java
@@ -1,6 +1,10 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  */
 
 package org.opensearch.dataprepper.plugins.kafka.buffer;
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/KafkaConsumerConfig.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/configuration/KafkaConsumerConfig.java
@@ -1,6 +1,10 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  */
 
 package org.opensearch.dataprepper.plugins.kafka.configuration;
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumerFactory.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumerFactory.java
@@ -1,6 +1,10 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  */
 
 package org.opensearch.dataprepper.plugins.kafka.consumer;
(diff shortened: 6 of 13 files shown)


data-prepper-core/src/main/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSet.java
```diff
@@ -127,19 +138,26 @@ public boolean isDone() {
...
-            if (Instant.now().isAfter(expiryTime)) {
+            if (!expired && Instant.now().isAfter(expiryTime)) {
+                expired = true;
                 if (progressCheckFuture != null) {
                     progressCheckFuture.cancel(false);
                 }
-                if (callbackFuture != null) {
+                metrics.increment(DefaultAcknowledgementSetMetrics.EXPIRED_METRIC_NAME);
+                if (invokeCallbackOnExpiry) {
+                    result = false;
```
graytaylor0, line 153: So we turn the acknowledgment timeout into a negative acknowledgment because that will lead to the KafkaConsumer to notice the failure and give up the partition? Were you able to verify that behavior with this change?

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/buffer/KafkaBufferConfig.java
```diff
@@ -29,6 +29,7 @@
 
 class KafkaBufferConfig implements KafkaProducerConfig, KafkaConsumerConfig {
     private static final Duration DEFAULT_DRAIN_TIMEOUT = Duration.ofSeconds(30);
+    static final Duration DEFAULT_ACKNOWLEDGEMENTS_TIMEOUT = Duration.ofHours(2);
```
dlvenable: I think for backward compatibility, we should set this to Int.MAX_VALUE like you did for the source.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -129,7 +132,7 @@ public KafkaCustomConsumer(final KafkaConsumer consumer,
         this.ownedPartitionsEpoch = new HashMap<>();
         this.metricsUpdatedTime = Instant.now().getEpochSecond();
         this.acknowledgedOffsets = new ArrayList<>();
-        this.acknowledgementsTimeout = Duration.ofSeconds(Integer.MAX_VALUE);
+        this.acknowledgementsTimeout = consumerConfig.getAcknowledgementsTimeout();
```
dlvenable, line 140: Shouldn't the unit tests for this class change?

data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetManagerTests.java
```diff
@@ -270,7 +270,40 @@ void testWithProgressCheckCallbacks_AcksExpire() {
...
+    @Test
+    void testCreateWithInvokeCallbackOnExpiryFalse() {
+        AcknowledgementSet acknowledgementSet = acknowledgementSetManager.create((flag) -> { result = flag; }, TEST_TIMEOUT, false);
+        assertThat(acknowledgementSet, notNullValue());
+    }
+
+    @Test
+    void testExpirationWithInvokeCallbackOnExpiryTrue() throws InterruptedException {
+        AcknowledgementSet acknowledgementSet = acknowledgementSetManager.create((flag) -> { result = flag; }, TEST_TIMEOUT, true);
+        eventHandle3 = mock(DefaultEventHandle.class);
```
dlvenable: No need for a `3` here. Please clean up some.

data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetManagerTests.java
```diff
@@ -270,7 +270,40 @@ void testWithProgressCheckCallbacks_AcksExpire() {
...
+        lenient().doAnswer(a -> {
```
dlvenable: Do you actually need this? Why is it lenient? Perhaps remove this entirely.

data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetManagerTests.java
```diff
@@ -270,7 +270,40 @@ void testWithProgressCheckCallbacks_AcksExpire() {
...
+            Boolean res = (Boolean)a.getArgument(0);
+            acknowledgementSet.release(eventHandle3, res);
+            return null;
+        }).when(eventHandle3).release(any(Boolean.class));
+        event3 = mock(JacksonEvent.class);
```
dlvenable: Drop the `3`.

Relatedly, can we just have common mocking for the event in `@BeforeEach`?

data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetManagerTests.java
```diff
@@ -270,7 +270,40 @@ void testWithProgressCheckCallbacks_AcksExpire() {
...
+        lenient().when(event3.getEventHandle()).thenReturn(eventHandle3);
+        acknowledgementSet.add(event3);
+        lenient().when(eventHandle3.getAcknowledgementSet()).thenReturn(acknowledgementSet);
+        acknowledgementSet.complete();
+
+        Thread.sleep(TEST_TIMEOUT.multipliedBy(5).toMillis());
```
dlvenable: Our test suite already takes 30 minutes. How can we avoid any sleep? In one class I injected a Clock to avoid sleeps. Maybe we can follow that pattern here as well.

https://github.com/opensearch-project/data-prepper/blob/c1aa83d533e75406296d56bc53dbd9e840f8501c/data-prepper-core/src/test/java/org/opensearch/dataprepper/core/pipeline/PipelineShutdownTest.java#L125-L127

data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetTests.java
```diff
@@ -313,4 +314,40 @@ void shutdown_cancels_progress_check_and_callback_future() throws NoSuchFieldExc
...
+    void testDefaultAcknowledgementSetWithCallbackOnExpiryTrue() throws Exception {
+        setupMetrics();
+        defaultAcknowledgementSet = new DefaultAcknowledgementSet(executor, (flag) -> {
+            acknowledgementSetResult = flag;
+        }, TEST_TIMEOUT, metrics, true);
+
+        defaultAcknowledgementSet.add(event);
+        defaultAcknowledgementSet.complete();
+
+        Thread.sleep(TEST_TIMEOUT.multipliedBy(2).toMillis());
```
dlvenable, line 328: This test should definitely be able to support an injectable `Clock` to avoid sleep. It doesn't have any threads that run.

data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/InactiveAcknowledgementSetManagerTests.java
```diff
@@ -32,4 +32,11 @@ void testCreateAPI() {
         assertThrows(UnsupportedOperationException.class, () -> acknowledgementSetManager.create((a)->{}, Duration.ofMillis(10)));
     }
 
+    @Test
```
dlvenable: Use `@ParameterizedTest` with `@ValueSource(booleans = {true, false})` instead. These are two different test cases.

data-prepper-core/src/main/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSet.java
kkondaka: Negative acknowledgement code path in Kafka buffer has been tested before. But I will do manual testing before committing this.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
kkondaka: Good point. Modified the test.

data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetManagerTests.java
kkondaka: This test is about expiration. Not sure if we can avoid sleep.

data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetTests.java
kkondaka: How we expire the acknowledgement set without sleeping?

kkondaka: We can reduce the timeout if sleep time is a concern

PR #6598, opened 2026-03-03
Add custom Jackson deserializer to handle empty plugin configs and re…

Davidding4718:
### Description
After upgrading Jackson, empty YAML plugin configurations like `stdout:` were being deserialized as empty strings `""` instead of null/empty objects, causing `MismatchedInputException`. This change adds a custom Jackson deserializer to `PluginModel` that explicitly handles the valid formats (`null`, `empty value`, `{}`) and rejects empty strings `""` with a clear error message. The custom serializer is also updated to output `{}` for empty (non-null) plugin settings. 
### Issues Resolved
NULL
 
### Check List

### DIFF
--- a/data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/PluginModelTests.java
+++ b/data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/PluginModelTests.java
@@ -26,6 +26,7 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -80,7 +81,7 @@ class PluginModelTests {
         InputStream inputStream = PluginModelTests.class.getResourceAsStream("plugin_model_with_empty_object.yaml");
 
         assertThat(serialized, notNullValue());
-        assertThat(serialized, equalTo(convertInputStreamToString(inputStream)));
+        assertThat(serialized, equalTo(stripComments(convertInputStreamToString(inputStream))));
     }
 
     @ParameterizedTest
@@ -281,6 +282,14 @@ class PluginModelTests {
             }
         }
         return stringBuilder.toString();
+    }
+
+    static String stripComments(final String content) {
+        final String stripped = Arrays.stream(content.split("\n"))
+                .filter(line -> !line.startsWith("#"))
+                .collect(java.util.stream.Collectors.joining("\n"))
+                .replaceAll("^\n+", "");
+        return content.endsWith("\n") ? stripped + "\n" : stripped;
     }
 
 }
--- a/data-prepper-api/src/test/resources/list_of_plugins_multiple_fields.yaml
+++ b/data-prepper-api/src/test/resources/list_of_plugins_multiple_fields.yaml
@@ -1,3 +1,12 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
+
 listOfPlugins:
   - customPluginA:
       key1: value1
--- a/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_array_value.yaml
+++ b/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_array_value.yaml
@@ -1,2 +1,10 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
 ---
 customPlugin: [1, 2, 3]
--- a/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_boolean_value.yaml
+++ b/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_boolean_value.yaml
@@ -1,2 +1,10 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
 ---
 customPlugin: true
--- a/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_empty_string.yaml
+++ b/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_empty_string.yaml
@@ -1,2 +1,10 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
 ---
 customPlugin: ""
--- a/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_explicit_empty.yaml
+++ b/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_explicit_empty.yaml
@@ -1,2 +1,10 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
 ---
 customPlugin: {}
--- a/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_number_value.yaml
+++ b/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_number_value.yaml
@@ -1,2 +1,10 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
 ---
 customPlugin: 123
--- a/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_with_empty_object.yaml
+++ b/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_with_empty_object.yaml
@@ -1,2 +1,10 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
 ---
 customPlugin: {}
--- a/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_with_empty_value.yaml
+++ b/data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_with_empty_value.yaml
@@ -1,2 +1,10 @@
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
 ---
 customPlugin:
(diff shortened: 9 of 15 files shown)


data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/PluginModel.java
```diff
@@ -41,7 +38,13 @@
...
 public class PluginModel {
 
-    private static final ObjectMapper SERIALIZER_OBJECT_MAPPER = new ObjectMapper();
+    private static final ObjectMapper SERIALIZER_OBJECT_MAPPER;
+    
+    static {
+        SERIALIZER_OBJECT_MAPPER = new ObjectMapper();
+        // Note: We don't configure coercion here because our custom deserializer
+        // handles all the cases (null, empty, {}, and rejects empty strings)
+    }
```
oeyh: Does this static block change anything from the original:
```
private static final ObjectMapper SERIALIZER_OBJECT_MAPPER = new ObjectMapper();
```

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/PluginModelTests.java
```diff
@@ -141,6 +141,85 @@ final void testUsingCustomDeserializer_with_array() throws JsonParseException, J
...
+    @Test
+    final void testDeserialize_emptyString_throwsException() throws IOException {
+        final InputStream inputStream = PluginModelTests.class.getResourceAsStream("plugin_model_empty_string.yaml");
+        final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+
+        final JsonMappingException exception = org.junit.jupiter.api.Assertions.assertThrows(
+            JsonMappingException.class,
+            () -> mapper.readValue(inputStream, PluginModel.class)
+        );
+        assertThat(exception.getMessage(), org.hamcrest.Matchers.containsString("Empty string is not allowed"));
```
oeyh: Use static import instead.

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/PluginModelTests.java
```diff
@@ -141,6 +141,85 @@ final void testUsingCustomDeserializer_with_array() throws JsonParseException, J
...
+    @Test
+    final void testDeserialize_nonEmptyString_throwsException() throws IOException {
+        final String yaml = "customPlugin: someStringValue";
+        final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+
+        final JsonMappingException exception = org.junit.jupiter.api.Assertions.assertThrows(
+            JsonMappingException.class,
+            () -> mapper.readValue(yaml, PluginModel.class)
+        );
+        assertThat(exception.getMessage(), org.hamcrest.Matchers.containsString("String values not allowed"));
```
oeyh: Use static import instead.

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SamplePipelineConfigurationTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Test
+    void deserialize_pipeline_withEmptyStringPluginConfig_throwsException() {
+        final InputStream inputStream = getClass().getResourceAsStream("sample_pipelines/sample_pipeline_plugin_empty_string.yaml");
+
+        final JsonMappingException exception = assertThrows(
+            JsonMappingException.class,
+            () -> objectMapper.readValue(inputStream, PipelinesDataFlowModel.class)
+        );
+        assertThat(exception.getMessage(), org.hamcrest.Matchers.containsString("Empty string is not allowed"));
```
oeyh: Same here.

data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_not_present.yaml
```diff
@@ -1,2 +1,2 @@
 ---
-customPlugin:
+customPlugin: {}
```
oeyh: Should we add new tests with {}, instead of changing the previous test? Otherwise, null case is no longer tested.
Similar for other tests.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/PluginModel.java
dlvenable: Yea, revert this small section.

I think you are working on code that I originally experimented with. I once had some configurations on the `ObjectMapper` that are no longer relevant.

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/PluginModelTests.java
```diff
@@ -141,6 +141,85 @@ final void testUsingCustomDeserializer_with_array() throws JsonParseException, J
...
+        assertThat(pluginModel2.getPluginName(), equalTo("customPlugin"));
+        assertThat(pluginModel2.getPluginSettings().size(), equalTo(0));
+    }
+
+    @Test
+    final void testDeserialize_emptyString_throwsException() throws IOException {
+        final InputStream inputStream = PluginModelTests.class.getResourceAsStream("plugin_model_empty_string.yaml");
+        final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+
+        final JsonMappingException exception = org.junit.jupiter.api.Assertions.assertThrows(
```
dlvenable: Another place to use a static import.

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SamplePipelineConfigurationTest.java
```diff
@@ -0,0 +1,120 @@
...
+    private static final String PIPELINE_NAME = "test-pipeline";
+
+    private ObjectMapper objectMapper;
+
+    @BeforeEach
+    void setup() {
+        objectMapper = new ObjectMapper(new YAMLFactory());
+    }
+
+    // --- Valid pipeline scenarios ---
```
dlvenable: Please don't have sections like this. They don't remain over time. There are two approaches:

1. You can use `@Nested` to group them, but I tend to avoid `@Nested` unless there is common setup.
2. Just remove this comment.

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SamplePipelineConfigurationTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        final PipelinesDataFlowModel model = objectMapper.readValue(inputStream, PipelinesDataFlowModel.class);
+        final PipelineModel pipeline = model.getPipelines().get(PIPELINE_NAME);
+
+        assertThat(pipeline.getSource().getPluginSettings().size(), equalTo(0));
+        assertThat(pipeline.getProcessors().get(0).getPluginSettings().size(), equalTo(0));
+        assertThat(pipeline.getSinks().get(0).getPluginSettings().size(), equalTo(0));
+    }
+
+    // --- Invalid pipeline scenario ---
```
dlvenable: Again, handle this comment.

data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_not_present.yaml
dlvenable: Yea, I think we should keep this file the way it was.

data-prepper-api/src/test/resources/pipeline_with_extension.yaml
```diff
@@ -2,10 +2,10 @@ extension:
   test_extension:
 test-pipeline:
   source:
-    testSource: null
+    testSource: {}
```
dlvenable: Do we need to modify this file?

I think the only scenario where exiting inputs will no longer pass is when we do a deserialize followed by a serialize.

Maybe a better way to test this would be to perform another deserialize and then compare the maps?

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/PluginModelTests.java
```diff
@@ -141,6 +141,85 @@ final void testUsingCustomDeserializer_with_array() throws JsonParseException, J
...
+
+        final PluginModel pluginModel2 = mapper.readValue(serialized, PluginModel.class);
+        assertThat(pluginModel2.getPluginName(), equalTo("customPlugin"));
+        assertThat(pluginModel2.getPluginSettings().size(), equalTo(0));
+    }
+
+    @Test
+    final void testRoundTrip_withNullValue() throws IOException {
+        final InputStream inputStream = PluginModelTests.class.getResourceAsStream("plugin_model_with_null.yaml");
+        final ObjectMapper mapper = new ObjectMapper(new YAMLFactory().enable(YAMLGenerator.Feature.USE_PLATFORM_LINE_BREAKS));
```
dlvenable: Do we need this feature enabled? `enable(YAMLGenerator.Feature.USE_PLATFORM_LINE_BREAKS)`

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SamplePipelineConfigurationTest.java
Davidding4718: Thanks! Will remove it in new commit

data-prepper-api/src/test/java/org/opensearch/dataprepper/model/configuration/SamplePipelineConfigurationTest.java
Davidding4718: Thanks! Will remove it in new commit

data-prepper-api/src/test/resources/org/opensearch/dataprepper/model/configuration/plugin_model_null.yaml
```diff
@@ -1,2 +1,2 @@
 ---
-customPlugin: null
+customPlugin: {}
```
oeyh: This should be null. There's another file below for `{}`.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/PluginModel.java
```diff
@@ -156,33 +171,53 @@ abstract static class AbstractPluginModelDeserializer<T extends PluginModel, M e
...
+            } else if (jsonParser.currentToken() == JsonToken.VALUE_STRING) {
+                final String value = jsonParser.getValueAsString();
+                if (value.isEmpty()) {
+                    throw context.weirdStringException(value, Map.class,
+                            "Empty string is not allowed for plugin '" + pluginName + "'. Use null, empty (no value), or {} instead.");
+                } else {
+                    throw context.weirdStringException(value, Map.class,
+                            "String values not allowed for plugin '" + pluginName + "'");
+                }
+            }
```
oeyh: No else branch here. Can you check what will happen if `jsonParser.currentToken()` is not START_OBJECT or VALUE_NULL or VALUE_STRING. For example,:
```yaml
  test-pipeline:
    source:                                                                                                                                                                                                                                                                                                          
      stdin: 123 # or true/false, or array [1, 2]                                                                                                                                                                                                                                                                                                     
    processor:                                                                                                                                                                                                                                                                                                       
    - uppercase_string: null
    sink:
    - stdout: null
```

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/PluginModel.java
```diff
@@ -117,10 +119,23 @@ public PluginModelSerializer(final Class<PluginModel> valueClass) {
         public void serialize(
                 final PluginModel value, final JsonGenerator gen, final SerializerProvider provider) throws IOException {
             gen.writeStartObject();
-            Map<String, Object> serializedInner = SERIALIZER_OBJECT_MAPPER.convertValue(value.innerModel, Map.class);
-            if(serializedInner != null && serializedInner.isEmpty())
-                serializedInner = null;
-            gen.writeObjectField(value.getPluginName(), serializedInner);
+
+            final String jsonString = SERIALIZER_OBJECT_MAPPER.writeValueAsString(value.innerModel);
+            final Map<String, Object> serializedInner = SERIALIZER_OBJECT_MAPPER.readValue(jsonString, Map.class);
```
oeyh, line 130: Can you clarify why this serialization and immediate deserialization is necessary here and add a comment above the code?

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/PluginModel.java
Davidding4718: Good point! Will add a else condition with following unit test.

data-prepper-api/src/main/java/org/opensearch/dataprepper/model/configuration/PluginModel.java
Davidding4718: Added a comment in the code. The round-trip is needed because `innerModel` can be a subclass (e.g. SinkInternalJsonModel) with extra Jackson-annotated fields. Serializing to JSON and reading back as a Map lets Jackson collect all those fields together, so we can check if the result is empty and write it out correctly. Directly accessing `pluginSettings` would miss any extra fields defined on subclasses.

PR #6621, opened 2026-03-06
Add substring expression functions

### LINKED ISSUE #6612: [FEATURE] Add substring expression functions (substringAfter, substringBefore, substringAfterLast, substringBeforeLast)

The expression language has no way to extract a portion of a string by delimiter. Existing string processors (replace_string, split_string, trim_string, etc.) mutate fields in-place but cannot produce a value for assignment via value_expression in
add_entries.

For example, given a field containing "/app/src/main.py", there is no way to extract "main.py", "main", or "py" and assign them to new fields using expressions.

Four new expression functions that extract substrings relative to a delimiter:

| Function | Description | Example | Result |
|----------|-------------|---------|--------|
| substringAfter(s, d) | Everything after the first occurrence of d | substringAfter("key=a=b", "=") | "a=b" |
| substringBefore(s, d) | Everything before the first occurrence of d | substringBefore("key=a=b", "=") | "key" |
| substringAfterLast(s, d) | Everything after the last occurrence of d | substringAfterLast("/app/src/main.py", "/") | "main.py" |
| substringBeforeLast(s, d) | Everything before the last occurrence of d | substringBeforeLast("/app/src/main.py", "/") | "/app/src" |

Behavior:
- Both arguments can be a JSON Pointer (resolved from the event) or a string literal.
- If the delimiter is not found, the original string is returned.
- If either argument is null, null is returned.
- Usable anywhere expressions are accepted: value_expression, add_when, delete_when, conditional routing.

Example: Derive file name and extension

```yaml
- add_entries:
    entries:
      - key: "extension"
        value_expression: 'substringAfterLast(/path, ".")'
      - key: "_filename"
        value_expression: 'substringAfterLast(/path, "/")'
- add_entries:
    entries:
      - key: "name"
        value_expression: 'substringBeforeLast(/_filename, ".")'
```

Given /path = "/app/src/main.py", this produces extension = "py", name = "main".

- **Using split_string + index access:** split_string splits a field in-place into an array but provides no way to select a specific element from the result and assign it to a new field via an expression.
- **Using grok or dissect processors:** These can parse strings into named fields but are heavier-weight, require pattern definitions, and don't compose naturally inside value_expression for inline field derivation.
- **A single generic substring function with mode parameter:** Four named functions are more readable and self-documenting than substring(/path, "/", "after", "last"). They also follow the convention of Apache Commons Lang's StringUtils which uses the
same four function names.
- The expression grammar's FunctionArg rule only accepts JSON Pointers or string literals — not other function calls. Nested calls like substringBefore(substringAfterLast(/path, "/"), ".") are not supported. Intermediate fields must be used instead (as shown in the example).

**Additional context**
 N/A

bagmarnikhil:
### Description

The expression language has no way to extract a portion of a string by delimiter. Existing string processors mutate fields in-place but cannot produce a value for assignment via value_expression.

Add four new expression functions:

- substringAfter(s, d): text after the first occurrence of d
- substringBefore(s, d): text before the first occurrence of d
- substringAfterLast(s, d): text after the last occurrence of d
- substringBeforeLast(s, d): text before the last occurrence of d

Both arguments accept JSON Pointers or string literals. If the delimiter is not found, the original string is returned. If the source resolves to null, null is returned.
 
### Issues Resolved
Resolves #6612
 
### Check List
- [Y ] New functionality includes testing.
- [ N] New functionality has a documentation issue. Please link to it in this PR.
- [Y ] Commits are signed with a real name per the DCO

### DIFF
--- a/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/SubstringAfterExpressionFunction.java
+++ b/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/SubstringAfterExpressionFunction.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.dataprepper.expression;
+
+import javax.inject.Named;
+
+@Named
+public class SubstringAfterExpressionFunction extends AbstractSubstringExpressionFunction {
+    static final String FUNCTION_NAME = "substringAfter";
+
+    @Override
+    public String getFunctionName() {
+        return FUNCTION_NAME;
+    }
+
+    @Override
+    protected String extractSubstring(final String source, final String delimiter) {
+        final int index = source.indexOf(delimiter);
+        if (index == -1) {
+            return source;
+        }
+        return source.substring(index + delimiter.length());
+    }
+}
--- a/docs/expression_syntax.md
+++ b/docs/expression_syntax.md
@@ -200,6 +200,22 @@ Currently, the following functions are supported
    - If the IP address is in the range of any given CIDR blocks, the function evaluates to true; otherwise, the function evaluates to false.
    - The function supports both IPv4 and IPv6 addresses.
    For example, `cidrContains(/sourceIp,"192.0.2.0/24","10.0.1.0/16")` evaluates to true if the event has `sourceIp` field with value "192.0.2.5".
+ * `substringAfter()`
+   - takes two String arguments: a source and a delimiter. Both should be either string literals or Json Pointers with String values.
+   - returns the substring of the source that follows the first occurrence of the delimiter. If the delimiter is not found, the original source string is returned. If the source resolves to null, null is returned.
+   For example, `substringAfter("key=a=b", "=")` returns "a=b", and `substringAfter("hello-world", "xyz")` returns "hello-world".
+ * `substringBefore()`
+   - takes two String arguments: a source and a delimiter. Both should be either string literals or Json Pointers with String values.
+   - returns the substring of the source that precedes the first occurrence of the delimiter. If the delimiter is not found, the original source string is returned. If the source resolves to null, null is returned.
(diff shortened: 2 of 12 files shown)


dlvenable:
@bagmarnikhil , Please rebase this from the latest `main` and use the approach that we just added in #6626. This changes how the functions receive arguments. It directly impacts your PR.

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/SubstringAfterExpressionFunction.java
```diff
@@ -0,0 +1,72 @@
...
+        return FUNCTION_NAME;
+    }
+
+    @Override
+    public Object evaluate(final List<Object> args, final Event event, final Function<Object, Object> convertLiteralType) {
+        if (args.size() != NUMBER_OF_ARGS) {
+            throw new RuntimeException("substringAfter() takes exactly two arguments");
+        }
+
+        final String[] strArgs = new String[NUMBER_OF_ARGS];
```
dlvenable: We should have an abstract class for substring functions as they all share the same basic behavior.

```
abstract AbstractSubstringExpressionFunction implements ExpressionFunction
```

You can use `getFunctionName` in the user messages.

bagmarnikhil: Updated the PR based on the suggested refactor

docs/expression_syntax.md
```diff
@@ -200,6 +200,22 @@ Currently, the following functions are supported
    - If the IP address is in the range of any given CIDR blocks, the function evaluates to true; otherwise, the function evaluates to false.
    - The function supports both IPv4 and IPv6 addresses.
    For example, `cidrContains(/sourceIp,"192.0.2.0/24","10.0.1.0/16")` evaluates to true if the event has `sourceIp` field with value "192.0.2.5".
+ * `substringAfter()`
```
graytaylor0, line 203: Please create a follow up PR to the documentation website here (https://github.com/opensearch-project/documentation-website).

This documentation is what users will use (https://docs.opensearch.org/latest/data-prepper/pipelines/functions)

graytaylor0 (approved):

docs/expression_syntax.md
bagmarnikhil: Created a pull request for documentation: https://github.com/opensearch-project/documentation-website/pull/12094

oeyh (approved):

PR #6624, opened 2026-03-10
Add S3 Scan processing condition evaluator to ensure object completeness

Zhangxunmt:
### Description
  Problem

  When using S3 scan to read the output of asynchronous batch inference jobs (e.g. Amazon Bedrock batch jobs), output files appear in S3 before the job has fully completed. There was no way to delay processing of those files until the job was actually done, leading to partial reads.

  Solution

  Introduces a new processing_conditions list at the S3 scan bucket level. Before processing an S3 object, the worker evaluates each applicable condition by reading a co-located manifest/sentinel file and running a Data Prepper expression against its JSON content. Processing is deferred until all conditions are satisfied.

  Configuration example
```
  scan:
    buckets:
      - bucket:
          name: "my-batch-bucket"
          data_selection: data_only
          processing_conditions:
            - object_name: "manifest.json.out"
              when: "/processedRecordCount == /totalRecordCount"
              applicable_prefix:
                - "bedrock-multisource/output-multisource/"
              retry_delay: PT5M
              max_retry: 10
          filter:
            include_prefix:
              - "bedrock-multisource/output-multisource/"
            exclude_suffix:
              - manifest.json.out
```
  For an object at s3://my-bucket/bedrock-multisource/output-multisource/job-123/result.jsonl.out, the worker will:
  1. Download s3://my-bucket/bedrock-multisource/output-multisource/job-123/manifest.json.out
  2. Parse it as JSON and evaluate /processedRecordCount == /totalRecordCount
  3. If true → proceed with processing
  4. If false or manifest not found → release the partition back with a now + retry_delay priority timestamp, retrying up to max_retry times

  processing_conditions fields:

object_name         │ string      │ required    │ Manifest file name, resolved in the same S3 directory as the object
when                      │ string      │ required    │ Data Prepper expression evaluated against manifest JSON
applicable_prefix  │ list           │ all objects │ Prefixes this condition applies to; absent means all objects in the bucket
retry_delay            │ Duration │ PT5M         │ Wait time before re-queuing when condition is not met
max_retry              │ int           │ 10               │ Max retry attempts before the partition is abandoned

  Implementation details:

  New classes:
  - S3ScanProcessingCondition — config POJO for a single condition entry
  - S3ScanProcessingConditionEvaluator — encapsulates all condition-checking logic: S3 manifest download, JSON parsing into JacksonEvent, expression evaluation via ExpressionEvaluator

  Modified classes:
  - S3ScanBucketOption — adds processing_conditions field
  - ScanObjectWorker — checks conditions before processing each object; on failure calls sourceCoordinator.closePartition(objectToProcess.get().getPartitionKey(), retryDelay, maxRetry, false);
  - S3ScanService — constructs a single shared S3ScanProcessingConditionEvaluator (stateless) and passes it to all workers
  - S3Source — injects ExpressionEvaluator via @DataPrepperPluginConstructor and threads it through to S3ScanService

  Testing:

  - 23 unit tests added for S3ScanProcessingConditionEvaluator covering:
    - Null/empty condition guards
    - include_prefix matching (no match skips, empty/null applies to all, partial match)
    - Expression evaluation (true/false, JSON fields forwarded correctly to evaluator)
    - Manifest key resolution (nested path uses same directory; root-level object uses filename only)
    - Error handling (NoSuchKeyException → defer; unexpected exception → defer)
    - Multi-condition logic (all pass, first fails short-circuits, non-matching skipped)
    - findFirstMatching for retry setting lookup
  - Existing unit and integration tests updated to compile with the new constructor signatures
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
+++ b/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
@@ -12,7 +12,12 @@ package org.opensearch.dataprepper.plugins.source.s3;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.opensearch.dataprepper.expression.ExpressionEvaluator;
+import org.opensearch.dataprepper.model.codec.InputCodec;
+import org.opensearch.dataprepper.model.configuration.PluginModel;
+import org.opensearch.dataprepper.model.configuration.PluginSetting;
+import org.opensearch.dataprepper.model.event.Event;
 import org.opensearch.dataprepper.model.event.JacksonEvent;
+import org.opensearch.dataprepper.model.plugin.PluginFactory;
 import org.opensearch.dataprepper.plugins.source.s3.configuration.S3ScanProcessingCondition;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -24,14 +29,20 @@ import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Evaluates {@link S3ScanProcessingCondition} entries for a given S3 object before
  * the object is processed. For each applicable condition the evaluator downloads the
- * manifest file co-located with the object and evaluates the {@code when} expression
- * against its JSON content.
+ * condition object co-located with the S3 object and evaluates the {@code when} expression
+ * against its content. When a {@code codec} is configured on the condition the object is
+ * parsed with that codec; otherwise the object is parsed as a JSON document.
  */
 public class S3ScanProcessingConditionEvaluator {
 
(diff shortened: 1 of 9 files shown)


data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/configuration/S3ScanProcessingCondition.java
```diff
@@ -0,0 +1,83 @@
...
+    private String when;
+
+    @JsonProperty("include_prefix")
+    private List<String> includePrefix;
+
+    @JsonProperty("retry_delay")
+    private Duration retryDelay = Duration.ofMinutes(5);
+
+    @JsonProperty("max_retry")
+    private int maxRetry = 10;
```
graytaylor0, line 51: Do you think 50 minutes default is enough? What happens when max_retry is exhausted? Do we just drop the object and mark it complete, or process it as is?

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
```diff
@@ -0,0 +1,126 @@
...
+        if (conditions == null || conditions.isEmpty()) {
+            return true;
+        }
+        for (final S3ScanProcessingCondition condition : conditions) {
+            if (!isApplicable(objectKey, condition)) {
+                continue;
+            }
+            final String manifestKey = resolveManifestKey(objectKey, condition.getFileName());
+            try {
+                final String content = readS3ObjectAsString(bucket, manifestKey);
```
graytaylor0: How do you think we should handle the case where the manifest file is not json?

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/configuration/S3ScanProcessingCondition.java
Zhangxunmt: Yes when the max_retry is reached, the object will be marked as "complete" and will not be processed anymore. From the current case studies 30+ minutes is enough because bedrock has a limit of 50,000 records for the input file. Usually bedrock takes less than 15 minutes to finish 50,000 records in a job.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
Zhangxunmt: This is a good call out. I think we can add a codec section in the processing condition config. Thoughts?

graytaylor0:
Also one more thing. As a follow up please create a follow up PR to the documentation website here (https://github.com/opensearch-project/documentation-website).

This documentation is what users will use (https://docs.opensearch.org/latest/data-prepper/pipelines/configuration/sources/s3/#scan-bucket)

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
dlvenable: Yes, I think a codec is ideal here.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/configuration/S3ScanProcessingCondition.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @JsonProperty("file_name")
+    @NotEmpty
+    private String fileName;
+
+    @JsonProperty("when")
+    @NotEmpty
+    private String when;
+
+    @JsonProperty("include_prefix")
```
dlvenable: I think this name may be a little confusing. It initially appeared to me to relate to whether or not to include the _condition_ file. But instead it means that the condition is applicable for any object in that prefix.

It seems we may need a slightly better name to help clarify this. Perhaps "applicable" would be clearer - `applicable_for_prefix`.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/ScanObjectWorker.java
```diff
@@ -225,6 +243,17 @@ private void startProcessingObject(final long waitTimeMillis) {
         final String bucket = objectToProcess.get().getPartitionKey().split("\\|")[0];
         final String objectKey = objectToProcess.get().getPartitionKey().split("\\|")[1];
 
+        final List<S3ScanProcessingCondition> conditions = bucketProcessingConditionsMap.get(bucket);
+        if (!processingConditionEvaluator.allConditionsMet(bucket, objectKey, conditions)) {
+            final S3ScanProcessingCondition failedCondition = processingConditionEvaluator.findFirstMatching(objectKey, conditions);
```
dlvenable: This may not be the failed condition, right? It seems we are just arbitrarily picking the first one.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/configuration/S3ScanProcessingCondition.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Represents a single processing condition entry under a scan bucket's
+ * {@code processing_conditions} list. Before an S3 object is processed, the
+ * worker downloads {@link #fileName} from the same S3 directory and evaluates
+ * {@link #when} against its JSON content. Processing is deferred when the
+ * condition is not yet satisfied.
+ */
+public class S3ScanProcessingCondition {
+
+    @JsonProperty("file_name")
```
dlvenable: S3 does not call these files. They are objects.

So it may make more sense to call this an `object_name`.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
```diff
@@ -0,0 +1,126 @@
...
+                        .withEventType("event")
+                        .withData(data)
+                        .build();
+                if (!expressionEvaluator.evaluateConditional(condition.getWhen(), event)) {
+                    LOG.debug("Processing condition '{}' not satisfied for {}/{} using manifest {}",
+                            condition.getWhen(), bucket, objectKey, manifestKey);
+                    return false;
+                }
+            } catch (final NoSuchKeyException e) {
+                LOG.debug("Manifest file {}/{} not found yet, condition not met", bucket, manifestKey);
```
dlvenable: These need not be manifest objects/files.


```suggestion
                LOG.debug("Object for condition {}/{} not found yet, condition not met", bucket, manifestKey);
```

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
```diff
@@ -0,0 +1,126 @@
...
+                                    final String objectKey,
+                                    final List<S3ScanProcessingCondition> conditions) {
+        if (conditions == null || conditions.isEmpty()) {
+            return true;
+        }
+        for (final S3ScanProcessingCondition condition : conditions) {
+            if (!isApplicable(objectKey, condition)) {
+                continue;
+            }
+            final String manifestKey = resolveManifestKey(objectKey, condition.getFileName());
```
dlvenable: Let's remove reference to "manifest" since that is a Bedrock term.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/configuration/S3ScanProcessingCondition.java
Zhangxunmt: renamed to applicable_prefix

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/ScanObjectWorker.java
Zhangxunmt: yeah probably not accurate enough. refactored to use the exact failed condition to proceed.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/configuration/S3ScanProcessingCondition.java
Zhangxunmt: done

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
Zhangxunmt: renamed to conditionObjectKey

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
Zhangxunmt: Added codec to process different formats of the object for conditions. It'd be one codec per condition so I used a Map to load all the codec.

data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ScanProcessingConditionEvaluator.java
```diff
@@ -40,83 +51,98 @@ public class S3ScanProcessingConditionEvaluator {
...
         return directory + fileName;
     }
 
+    private Event parseFirstEventWithCodec(final String bucket,
+                                            final String key,
+                                            final InputCodec codec) throws IOException {
+        final AtomicReference<Event> firstEvent = new AtomicReference<>();
+        final GetObjectRequest request = GetObjectRequest.builder().bucket(bucket).key(key).build();
+        try (final ResponseInputStream<GetObjectResponse> response = s3Client.getObject(request)) {
+            codec.parse(response, record -> firstEvent.compareAndSet(null, record.getData()));
```
graytaylor0, line 141: One thing we should make sure of. Does this actually work with a single json object in a file? Last I checked the json codec only supported json arrays. Did you test again with a static json object (not ndjson)

Zhangxunmt: Yes I tested and verified it works with single json object. I uses the actual one of the real manifest file in the below format. 
```
{"totalRecordCount":50000,"processedRecordCount":50000,"successRecordCount":30000,"errorRecordCount":0,"inputTokenCount":5300000,"outputTokenCount":0,"inputAudioSecond":0,"inputVideoSecond":0,"inputStandardImageCount":0,"inputDocumentImageCount":0,"inputTextTokenCount":0,"inputImageTokenCount":0,"inputAudioTokenCount":0,"inputVideoTokenCount":0,"outputTextTokenCount":0,"outputImageTokenCount":0}
```

Zhangxunmt:
> Also one more thing. As a follow up please create a follow up PR to the documentation website here (https://github.com/opensearch-project/documentation-website).
> 
> This documentation is what users will use (https://docs.opensearch.org/latest/data-prepper/pipelines/configuration/sources/s3/#scan-bucket)

The document PR is created for this new config update: https://github.com/opensearch-project/documentation-website/pull/12104.

PR #6604, opened 2026-03-04
Otel metrics source http service

TomasLongo:
### Description
Introduce HTTP/protobuf and HTTP/JSON support for OTel Metrics source.
Adds endpoint to receive OTLP data over HTTP. Aligns with similar support
for OTelTraceSource.
 
### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has a documentation issue. Please link to it in this PR.
  - [ ] New functionality has javadoc added
- [ ] Commits are signed with a real name per the DCO

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/data-prepper/blob/main/CONTRIBUTING.md).

PR #6633, opened 2026-03-12
Fix requestsTooLarge metric reporting when decompression buffer overflows on armeria in otel logs source

graytaylor0:
### Description
This change handles an edge case where the otel logs source does not report the "requestsTooLarge" metric despite throwing a 413 exception. 

The edge case happens when the source is receiving compressed requests. If the decompressed request exceeds the max_request_length, the decompression buffer in armeria throws a 413 error, but it does not increment the metric "requestsTooLarge" because the exception handler does not run in this scenario.
 
### Issues Resolved
Related to #5594 
 
### Check List

### DIFF
--- a/data-prepper-plugins/otel-logs-source/src/test/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSourceHttpTest.java
+++ b/data-prepper-plugins/otel-logs-source/src/test/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSourceHttpTest.java
@@ -46,7 +46,6 @@ import static org.opensearch.dataprepper.plugins.source.otellogs.OtelLogsSourceC
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.UUID;
 import java.util.Base64;
 import java.util.List;
 import java.util.Map;
@@ -390,7 +389,7 @@ class OTelLogsSourceHttpTest {
     private void assertSecureResponseWithStatusCode(final AggregatedHttpResponse response,
                                                     final HttpStatus expectedStatus,
                                                     final Throwable throwable) {
-         assertThat("Http Status", response.status(), equalTo(expectedStatus));
+        assertThat("Http Status", response.status(), equalTo(expectedStatus));
         assertThat("Http Response Throwable", throwable, is(nullValue()));
 
         final List<String> headerKeys = response.headers()


data-prepper-plugins/otel-logs-source/src/test/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSourceHttpTest.java
```diff
@@ -389,7 +390,7 @@ private ExportLogsServiceRequest createExportLogsRequest() {
     private void assertSecureResponseWithStatusCode(final AggregatedHttpResponse response,
                                                     final HttpStatus expectedStatus,
                                                     final Throwable throwable) {
-        assertThat("Http Status", response.status(), equalTo(expectedStatus));
+         assertThat("Http Status", response.status(), equalTo(expectedStatus));
```
KarstenSchnitter: Unecessary whitespace.
```suggestion
        assertThat("Http Status", response.status(), equalTo(expectedStatus));
```

graytaylor0: @KarstenSchnitter The test I had did not actually cover the new code. I have not been able to find a good way to test this code path with log access writer. I think there needs to be some actual integration tests that don't mock the OtelLogsSource server to cover this. If you see a way to add a unit test for this let me know.

data-prepper-plugins/otel-logs-source/src/main/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSource.java
```diff
@@ -144,6 +146,12 @@ private Server createServer(ServerBuilder serverBuilder, Buffer<Record<Object>>
             serverBuilder.http(oTelLogsSourceConfig.getPort());
         }
 
+        serverBuilder.accessLogWriter(log -> {
```
kkondaka, line 149: Are you sure this catches all cases?
When `  buffer.writeBytes(request.toByteArray(), null, bufferWriteTimeoutInMillis);` throws an exception, we catch and throw  `new BufferWriteException(e.getMessage(), e);`

kkondaka:
@graytaylor0 we should probably add the siimilar metric for OTEL traces and metrics as well, right?

data-prepper-plugins/otel-logs-source/src/test/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSourceHttpTest.java
KarstenSchnitter: Unfortunately, I do not see an easy possibility. Furthermore, it would be more of a test of armeria.

data-prepper-plugins/otel-logs-source/src/main/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSource.java
KarstenSchnitter: Isn't this exception already handled? This would indicate to me, that this PR closes a hole in the existing error handling.

KarstenSchnitter:
> @graytaylor0 we should probably add the siimilar metric for OTEL traces and metrics as well, right?

I agree. In my experience the issue is most likely with metrics.Note, that this will create a conflict with #6604.

data-prepper-plugins/otel-logs-source/src/main/java/org/opensearch/dataprepper/plugins/source/otellogs/OTelLogsSource.java
graytaylor0: Yes this should already be handled. The problem was that armeria was swallowing the exception for decompression overflowing the max request size, and not running the exception handler code that increments this metric.

graytaylor0:
> @graytaylor0 we should probably add the siimilar metric for OTEL traces and metrics as well, right?

Yes we should. I will follow up with a similar change in those sources.

graytaylor0:
> > @graytaylor0 we should probably add the siimilar metric for OTEL traces and metrics as well, right?
> 
> I agree. In my experience the issue is most likely with metrics.Note, that this will create a conflict with #6604.

Yes I will make a similar change in the other sources as a follow up.

PR #6634, opened 2026-03-12
fix: mitigate STS assume role throttling in Kafka buffer

dinujoh:
### Description
Prevent excessive STS AssumeRole calls when customers delete their IAM role or misconfigure the trust policy. Previously, one customer could generate 12,000 STS calls in 4 minutes due to unbounded retries of non-retryable AccessDeniedException errors.

Changes:
- KafkaSecurityConfigurer: Fail fast on STS 403 (AccessDenied) in getBootStrapServersForMsk() instead of retrying 360 times
- KafkaSecurityConfigurer: Replace fixed 10s retry sleep with exponential backoff (10s to 10min max) for retryable STS and Kafka errors
- KafkaCustomConsumer: Replace fixed 10s retry with exponential backoff using Kafka's ExponentialBackoff (10s to 10min max) for AuthenticationException errors
- KafkaCustomConsumer: Use Duration constants for backoff readability
- KafkaCustomConsumer: Reset backoff counter on successful poll to handle transient errors gracefully
- Add exponential backoff to outer run() exception handler

 
### Check List

### DIFF
--- a/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
+++ b/data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
@@ -20,6 +20,7 @@ import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.AuthenticationException;
 import org.apache.kafka.common.errors.RebalanceInProgressException;
+import org.apache.kafka.common.utils.ExponentialBackoff;
 import org.apache.kafka.common.errors.RecordDeserializationException;
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
@@ -70,10 +71,8 @@ public class KafkaCustomConsumer implements Runnable, ConsumerRebalanceListener 
     private static final Logger LOG = LoggerFactory.getLogger(KafkaCustomConsumer.class);
     private static final Long COMMIT_OFFSET_INTERVAL_MS = 300000L;
     private static final int RETRY_ON_EXCEPTION_SLEEP_MS = 1000;
-    private static final long INITIAL_AUTH_BACKOFF_MS = 10000L;
-    private static final long MAX_AUTH_BACKOFF_MS = 300000L;
-    static final int MAX_CONSECUTIVE_AUTH_FAILURES = 50;
-    private static final double BACKOFF_MULTIPLIER = 2.0;
+    static final Duration INITIAL_BACKOFF = Duration.ofSeconds(10);
+    static final Duration MAX_BACKOFF = Duration.ofMinutes(10);
     private static final int BUFFER_WRITE_TIMEOUT = 2000;
     static final String DEFAULT_KEY = "message";
 
@@ -103,8 +102,8 @@ public class KafkaCustomConsumer implements Runnable, ConsumerRebalanceListener 
     private final LogRateLimiter errLogRateLimiter;
     private final ByteDecoder byteDecoder;
     private final long maxRetriesOnException;
-    private int consecutiveAuthFailures;
-    private long currentAuthBackoffMs;
+    private final ExponentialBackoff exponentialBackoff;
+    private long authFailureAttempts;
     private final Map<Integer, Long> partitionToLastReceivedTimestampMillis;
     private final CompressionOption compressionConfig;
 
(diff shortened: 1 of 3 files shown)


data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -238,9 +247,19 @@ <T> void consumeRecords() throws Exception {
...
-            LOG.warn("Authentication error while doing poll(). Will retry after 10 seconds", e);
+            consecutiveAuthFailures++;
             topicMetrics.getNumberOfPollAuthErrors().increment();
-            Thread.sleep(10000);
+            if (consecutiveAuthFailures >= MAX_CONSECUTIVE_AUTH_FAILURES) {
+                LOG.error("Reached maximum consecutive authentication failures ({}) for topic {}. " +
+                        "Stopping consumer. Verify that the IAM role exists and trust policy is correct.",
+                        MAX_CONSECUTIVE_AUTH_FAILURES, topicName, e);
+                shutdownInProgress.set(true);
+                return;
```
dlvenable: I'm not sure that this is the right behavior. I think this means that the thread will terminate. But the pipeline will continue to run. I don't think we currently have a way for a source or buffer to shutdown the pipeline. That would probably be ideal.

The proposed change here would create a silent failure that would confuse operators.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -70,6 +70,10 @@ public class KafkaCustomConsumer implements Runnable, ConsumerRebalanceListener
     private static final Logger LOG = LoggerFactory.getLogger(KafkaCustomConsumer.class);
     private static final Long COMMIT_OFFSET_INTERVAL_MS = 300000L;
     private static final int RETRY_ON_EXCEPTION_SLEEP_MS = 1000;
+    private static final long INITIAL_AUTH_BACKOFF_MS = 10000L;
```
dlvenable: Use `Duration` to make this more readable.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -70,6 +70,10 @@ public class KafkaCustomConsumer implements Runnable, ConsumerRebalanceListener
...
+    private static final long MAX_AUTH_BACKOFF_MS = 300000L;
```
dlvenable: This is only 5 minutes. Should we go up higher? Maybe 10 minutes.

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/util/KafkaSecurityConfigurer.java
```diff
@@ -294,8 +294,18 @@ public static String getBootStrapServersForMsk(final AwsConfig awsConfig,
...
-            } catch (KafkaException | StsException e) {
-                log.info("Failed to get bootstrap server information from MSK. Will try every 10 seconds for {} seconds", 10*MAX_KAFKA_CLIENT_RETRIES, e);
+            } catch (StsException e) {
+                if (e.statusCode() == 403) {
+                    throw new RuntimeException("Access denied when calling STS to get bootstrap server information from MSK. " +
+                            "Verify that the role exists and the trust policy is correctly configured.", e);
+                }
+                log.info("Failed to get bootstrap server information from MSK due to STS error. Will try every 10 seconds for {} seconds", 10*MAX_KAFKA_CLIENT_RETRIES, e);
+                try {
+                    Thread.sleep(10000);
```
dlvenable: Why is there no exponential backoff here?

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
```diff
@@ -238,9 +247,19 @@ <T> void consumeRecords() throws Exception {
...
+                LOG.error("Reached maximum consecutive authentication failures ({}) for topic {}. " +
+                        "Stopping consumer. Verify that the IAM role exists and trust policy is correct.",
+                        MAX_CONSECUTIVE_AUTH_FAILURES, topicName, e);
+                shutdownInProgress.set(true);
+                return;
+            }
+            LOG.warn("Authentication error while doing poll() for topic {} (consecutive failures: {}). " +
+                    "Will retry after {} ms", topicName, consecutiveAuthFailures, currentAuthBackoffMs, e);
+            sleepMillis(currentAuthBackoffMs);
+            currentAuthBackoffMs = Math.min((long) (currentAuthBackoffMs * BACKOFF_MULTIPLIER), MAX_AUTH_BACKOFF_MS);
```
dlvenable: As this is the Kafka buffer, why not use `org.apache.kafka.common.utils.ExponentialBackoff` instead of having this logic?

data-prepper-plugins/kafka-plugins/src/main/java/org/opensearch/dataprepper/plugins/kafka/consumer/KafkaCustomConsumer.java
dinujoh: Ack. Removed this change. Consumer will retry indefinitely with backoff

PR #6637, opened 2026-03-13
fix: add BackoffCredentialsProvider to mitigate STS throttling across all plugins

dinujoh:
### Description

Wrap StsAssumeRoleCredentialsProvider with BackoffCredentialsProvider in CredentialsProviderFactory. When credential resolution fails (e.g. role deleted or trust policy misconfigured), the wrapper caches the failure and applies exponential backoff (10s to 10min) before retrying STS, preventing excessive AssumeRole calls that cause STS throttling.

This protects all plugins that use CredentialsProviderFactory including S3, OpenSearch, Lambda, SQS, and most AWS-integrated sources and sinks.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-plugins/aws-plugin/src/main/java/org/opensearch/dataprepper/plugins/aws/BackoffCredentialsProvider.java
+++ b/data-prepper-plugins/aws-plugin/src/main/java/org/opensearch/dataprepper/plugins/aws/BackoffCredentialsProvider.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.dataprepper.plugins.aws;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.core.exception.SdkException;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+
+/**
+ * A credentials provider wrapper that applies exponential backoff when the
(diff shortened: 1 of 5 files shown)


data-prepper-plugins/aws-plugin/src/main/java/org/opensearch/dataprepper/plugins/aws/BackoffCredentialsProvider.java
```diff
@@ -0,0 +1,96 @@
...
+ * this provider throws the cached exception without calling the delegate,
+ * preventing excessive STS AssumeRole calls when a role is deleted or
+ * misconfigured.
+ *
+ * <p>On successful credential resolution, the backoff state is reset.</p>
+ */
+class BackoffCredentialsProvider implements AwsCredentialsProvider {
+    private static final Logger LOG = LoggerFactory.getLogger(BackoffCredentialsProvider.class);
+
+    static final Duration INITIAL_BACKOFF = Duration.ofSeconds(10);
```
dlvenable, line 34: I think we should make these configurable in the `aws` plugin. Not configured per credential, but at the high level.

```
  aws:
    max_backoff: 10m
    configurations:
      default:
        sts_role_arn: arn:aws:iam::123456789012:role/MyRole
        region: us-east-2
```

dinujoh: Should max_backoff be under  aws->configurations ?

```
aws:
  configurations:
    max_backoff: 10m
    default:
      sts_role_arn: ...

```

dlvenable: The way I view `configurations` is to be essentially named configurations, so it would be a map. But, `default` is a special name. See #2570 for more detail.

So it should be directly under `aws`.

dinujoh: `aws` can have other plugins correct ? for example `AwsSecretPlugin` ?

PR #6653, opened 2026-03-18
add UUID as new entry value in the add_entries processor

Zhangxunmt:
### Description

Added a new entry in the add_entries processor that generates a unique ID for each record. This is needed for cases where the original source data does not contain a unique identifier. The unique ID is essential for running asynchronous batch inference jobs, as it is used to match and merge the inference results back with the source data. 

UUID.randomUUID() (UUID v4) uses Java's SecureRandom, and collision probability is so low it has never been observed in practice in any production system, which is what all major distributed systems use for this exact problem.

Usage
```
  processor:
    - add_entries:
        entries:
          - key: recordId
            value_expression: 'generateUuid()'
```

### Issues Resolved
 
### Check List

### DIFF
--- a/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/GenerateUuidExpressionFunction.java
+++ b/data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/GenerateUuidExpressionFunction.java
@@ -18,12 +18,12 @@ import java.util.function.Function;
 
 /**
  * Expression function that generates a random UUID (version 4) string.
- * Usage: {@code generate_uuid()}
+ * Usage: {@code generateUuid()}
  */
 @Named
 public class GenerateUuidExpressionFunction implements ExpressionFunction {
 
-    static final String FUNCTION_NAME = "generate_uuid";
+    static final String FUNCTION_NAME = "generateUuid";
 
     @Override
     public String getFunctionName() {
--- a/data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/GenerateUuidExpressionFunctionTest.java
+++ b/data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/GenerateUuidExpressionFunctionTest.java
@@ -21,6 +21,7 @@ import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.Mockito.mock;
 
@@ -31,8 +32,8 @@ class GenerateUuidExpressionFunctionTest {
     private final Function<Object, Object> convertLiteralType = v -> v;
 
     @Test
-    void getFunctionName_returns_generate_uuid() {
-        assertThat(function.getFunctionName(), equalTo("generate_uuid"));
+    void getFunctionName_returns_generateUuid() {
+        assertThat(function.getFunctionName(), equalTo("generateUuid"));
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


graytaylor0:
This may be better implemented as an expression function rather than a new parameter

```
  processor:
    - add_entries:
        entries:
          - key: recordId
            value_expression: 'generate_uuid()'
```

Zhangxunmt:
> This may be better implemented as an expression function rather than a new parameter
> 
> ```
>   processor:
>     - add_entries:
>         entries:
>           - key: recordId
>             value_expression: 'generate_uuid()'
> ```

Thanks for the comment! @graytaylor0 Make sense to use a function. BTW, I think the data prepper documents needs update to list common available functions. Probably I can do that in a separate PR.

graytaylor0:
Looks good! Please follow up with a documentation change (https://docs.opensearch.org/latest/data-prepper/pipelines/functions/). Thanks!

Zhangxunmt:
> Looks good! Please follow up with a documentation change (https://docs.opensearch.org/latest/data-prepper/pipelines/functions/). Thanks!

Docs update PR -> https://github.com/opensearch-project/documentation-website/pull/12130.

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/GenerateUuidExpressionFunction.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.function.Function;
+
+/**
+ * Expression function that generates a random UUID (version 4) string.
+ * Usage: {@code generate_uuid()}
+ */
+@Named
+public class GenerateUuidExpressionFunction implements ExpressionFunction {
+
+    static final String FUNCTION_NAME = "generate_uuid";
```
dlvenable: Data Prepper functions use `lowerCamelCase`. Please update to `generateUuid`.


```suggestion
    static final String FUNCTION_NAME = "generateUuid";
```

data-prepper-expression/src/test/java/org/opensearch/dataprepper/expression/GenerateUuidExpressionFunctionTest.java
```diff
@@ -0,0 +1,59 @@
...
+    void getFunctionName_returns_generate_uuid() {
+        assertThat(function.getFunctionName(), equalTo("generate_uuid"));
+    }
+
+    @Test
+    void evaluate_returns_valid_uuid_string() {
+        final Object result = function.evaluate(Collections.emptyList(), event, convertLiteralType);
+        assertThat(result, instanceOf(String.class));
+        final String uuidStr = (String) result;
+        // UUID.fromString throws if invalid
```
dlvenable: This would be clearer and be retained in code rather than as a comment.

```
UUID regeneratedUuid = 
  assertDoesNotThrow(() -> UUID.fromString("hello"));
assertThat(regeneratedUuid.toString(), equalTo(uuidStr));
```

data-prepper-plugins/mutate-event-processors/src/test/java/org/opensearch/dataprepper/plugins/processor/mutateevent/AddEntryProcessorTests.java
```diff
@@ -1088,6 +1088,96 @@ public void testAddFlattenedNestedEntryIterateOn() {
                 equalTo(List.of(Map.of("key", 5, "nested/newMessage", 3)))); // [{"key": 5, "nested/newMessage": 3}}]
     }
 
+    @Test
+    void test_generate_uuid_expression_adds_uuid_string_to_event() {
+        final String uuidExpr = "generate_uuid()";
```
dlvenable: Be sure to update these as well.

data-prepper-expression/src/main/java/org/opensearch/dataprepper/expression/GenerateUuidExpressionFunction.java
Zhangxunmt: Good catch!

PR #6657, opened 2026-03-19
Fix duplicate _seconds suffix in APM latency metric name

vamsimanohar:
## Summary
- The Prometheus sink automatically appends unit suffixes to metric names (e.g., unit `"s"` → `_seconds`)
- The APM service map processor was naming the histogram metric `latency_seconds` with unit `"s"`, resulting in `latency_seconds_seconds` when exported to Prometheus
- Renamed the metric to `latency` so the final Prometheus metric name is correctly `latency_seconds`

## Test plan
- [x] Unit tests updated and passing for `ApmServiceMapMetricsUtilTest` and `OTelApmServiceMapProcessorTest`


ps48:
Built this change locally and checked the naming change works! Thanks @vamsimanohar @graytaylor0 

Before:
Screenshot 2026-03-19 at 9 56 48 AM

After:
Screenshot 2026-03-19 at 4 11 43 PM

Issue #6610, opened 2026-03-04
[FEATURE] Add filter_entries processor for array element filtering


bagmarnikhil:
**Is your feature request related to a problem? Please describe.**

There is no way to remove elements from an array based on a condition. While existing processors can modify keys within array elements or drop entire events, none can filter the array itself — keeping only elements that match a condition and discarding
the rest.

For example, given an array of objects at /items, there is no way to produce a new array containing only the elements where /type == "a".

**Describe the solution you'd like**

A new filter_entries processor that evaluates a condition against each element of an array and keeps only those elements where the condition is true.

| Option | Required | Description |
|--------|----------|-------------|
| source | Yes | Key of the array field to filter |
| target | No | Key to write the filtered array to. Defaults to source (in-place) |
| condition | Yes | Expression evaluated per element. Elements where this is true are kept |
| filter_entries_when | No | Expression evaluated against the root event. If false, the processor is skipped entirely |

Follows the mutate-event naming convention: add_entries, delete_entries, select_entries

Example

Input:
```json
{
  "items": [
    {"type": "a", "id": "1"},
    {"type": "b", "id": "2"},
    {"type": "a", "id": "3"}
  ]
}
```

Pipeline:
```yaml
- filter_entries:
    source: "/items"
    target: "/filtered"
    condition: '/type == "a"'
```

Expected output:
```json
{
  "filtered": [
    {"type": "a", "id": "1"},
    {"type": "a", "id": "3"}
  ]
}
```

**Describe alternatives you've considered (Optional)**

- **Extending select_entries with iterate_on:** select_entries filters keys (columns) within an event, while the needed operation is filtering elements (rows) within an array. These are semantically distinct — adding both behaviors to one processor would
make configuration confusing. Each mutate-event processor today does one thing: add_entries adds keys, delete_entries removes keys, select_entries keeps keys. filter_entries keeping elements is a natural extension of this pattern.

- **Using delete_entries with iterate_on and delete_from_element_when:** This can conditionally delete keys from array elements but cannot remove the element itself from the array. An element with all its keys deleted still remains in the array as an
empty object {}.

- **Using drop_events with a condition:** drop_events operates at the event level. It can drop an entire event but has no way to target individual elements within an array field inside the event.

**Additional context**
N/A

dlvenable:
I think this should be named `filter_list` since it operates on lists.

The value `condition` should probably be named `keep_when`.

[...]

Issue #6609, opened 2026-03-04
[FEATURE] Support parent event field access during iterate_on processing


bagmarnikhil:
**Is your feature request related to a problem? Please describe.**

When iterate_on is used in add_entries or delete_entries, all expressions (value_expression, format, add_to_element_when, delete_from_element_when) are evaluated against a temporary context built from the current array element only. Fields from the parent event are not accessible.

For example, given this input event:

```json
{"vendor": "Acme", "items": [{"id": "1"}, {"id": "2"}]}
```
And this pipeline:

```yaml
- add_entries:
    iterate_on: "/items"
    entries:
      - key: "vendor"
        value_expression: "/vendor"
```

The user expects each element to get "vendor": "Acme", but /vendor resolves against the element ({"id": "1"}), finds nothing, and produces null.

This prevents two common patterns:
- Injecting parent-level fields into each element of a nested array 
- Reading sibling fields during nested array iteration

**Describe the solution you'd like**
Element-first, parent-fallback resolution — no configuration changes required.

When resolving a key during iterate_on processing:

1. Resolve against the element context (current behavior).
2. If not found, fall back to the parent event.

Element keys always take priority, so existing pipelines are unaffected. The fallback is automatic and transparent.

This applies to all expression evaluation points during iterate_on:

| Processor | Expression Field | Current Context | Proposed Fallback |
|-----------|-----------------|-----------------|-------------------|
| add_entries | value_expression | element only | parent event |
| add_entries | format (${/key} references) | element only | parent event |
| add_entries | add_to_element_when | element only | parent event |
| delete_entries | delete_from_element_when | element only | parent event |

**Describe alternatives you've considered (Optional)**

N/A

**Additional context**

- No new configuration options — the fallback is implicit and backward-compatible.
- Element keys always win over parent keys, so no existing behavior changes.

dlvenable:
@bagmarnikhil ,

I agree that the current behavior is incorrect. I think ideally we'd differentiate between JSON pointer at the root `/` and JSON pointer relative (no `/`). However that is not the current behavior.

We need to be careful not to break existing functionality and how existing users expect these configurations to work. So this may require an additional configuration. Perhaps in the future we remove it, but for now I think we will need something configuration like `disable_root_keys: true`.

yavmanis:
Here is a simplified explanation of the problem and the implementation approaches to fix it:

[...]

PR #6659, opened 2026-03-20
Add filter_list processor to filter list elements

yavmanis:
## filter_list Processor

The `filter_list` processor filters elements within an array field by evaluating a condition against each element and retaining only those that satisfy the condition.

It supports:

* Arrays of objects
* Arrays of primitives (strings, numbers, booleans)

---

## Description

This processor iterates over a specified array field and evaluates an expression for each element.

* Element-level filtering is controlled by `keep_element_when`
* Processor execution can be conditionally controlled using `filter_list_when`

This behavior is aligned with existing processors like `delete_entries` for consistency.

---

## Configuration Updates

To promote consistency with other processors that iterate over elements:

* **`source` → `iterate_on`**
  Better reflects that the processor iterates over a collection

* **`keep_when` → `keep_element_when`**
  Clearly indicates evaluation happens at the element level

Other configuration options remain unchanged.

---

## Configuration Options

| Option            | Required | Type         | Default    | Description                                                                             |
| ----------------- | -------- | ------------ | ---------- | --------------------------------------------------------------------------------------- |
| iterate_on        | Yes      | String       | —          | Key of the array field to iterate and filter (supports nested paths like `outer/inner`) |
| target            | No       | String       | iterate_on | Key to write filtered output (defaults to in-place filtering)                           |
| keep_element_when | Yes      | String       | —          | Expression evaluated per element; elements where true are retained                      |
| filter_list_when  | No       | String       | null       | Expression evaluated at root event level; processor runs only if true                   |
| tags_on_failure   | No       | List<String> | null       | Tags added to event metadata on failure                                                 |

---

## Execution Semantics

* `keep_element_when` → evaluated in **element scope**
* `filter_list_when` → evaluated in **root event scope**
* If `filter_list_when` is false → processor is skipped
* If `target` is not provided → filtering is **in-place**
* If `target` is provided → filtered result is written to target field

---

## Failure Handling

If evaluation fails for one or more elements:

* Event is marked using `tags_on_failure`
* Metadata fields are added:

  * `filter_list_processor_failed_elements_count`
  * `filter_list_processor_failed_elements`

This avoids silent data loss and improves observability.

---

## Example Pipeline Configuration

```yaml
filter-list-pipeline:
  source:
    file:
      path: "pipelines/input.json"
      format: "json"
      record_type: "event"

  processor:
    - filter_list:
        iterate_on: "items"
        keep_element_when: '/status == "active"'
        filter_list_when: '/env == "production"'

    - filter_list:
        iterate_on: "tags"
        keep_element_when: '/value != ""'
        filter_list_when: '/type == "tagged"'

    - filter_list:
        iterate_on: "scores"
        target: "passing_scores"
        keep_element_when: '/value > 50'
        filter_list_when: '/type == "grades"'

    - filter_list:
        iterate_on: "items"
        target: "critical_items"
        keep_element_when: '/priority == "critical"'
        filter_list_when: '/type == "tickets"'

  sink:
    - stdout:
```

---

## Input

```json
{"items": [{"name": "item1", "status": "active"}, {"name": "item2", "status": "inactive"}, {"name": "item3", "status": "active"}], "env": "production", "type": "inventory"}
{"tags": ["important", "", "urgent", ""], "type": "tagged"}
{"scores": [95, 30, 75, 10, 88], "subject": "math", "type": "grades"}
{"items": [{"name": "bug1", "priority": "critical"}, {"name": "bug2", "priority": "low"}, {"name": "bug3", "priority": "critical"}], "type": "tickets"}
{"items": [{"name": "item1", "status": "active"}], "env": "staging", "type": "inventory"}
```

---

## Output

```json
{"items":[{"name":"item1","status":"active"},{"name":"item3","status":"active"}],"env":"production","type":"inventory"}
{"tags":["important","urgent"],"type":"tagged"}
{"scores":[95,30,75,10,88],"subject":"math","type":"grades","passing_scores":[95,75,88]}
{"items":[{"name":"bug1","priority":"critical"},{"name":"bug2","priority":"low"},{"name":"bug3","priority":"critical"}],"type":"tickets","critical_items":[{"name":"bug1","priority":"critical"},{"name":"bug3","priority":"critical"}]}
{"items":[{"name":"item1","status":"active"}],"env":"staging","type":"inventory"}
```

---

## End-to-End Scenarios

| # | Scenario                                                            | Result |
| - | ------------------------------------------------------------------- | ------ |
| 1 | Object array filtered in-place (`status == active`, env=production) | PASS   |
| 2 | Primitive array removes empty values                                | PASS   |
| 3 | Number array filtered into new target (`> 50`)                      | PASS   |
| 4 | Object array filtered into new target (`priority == critical`)      | PASS   |
| 5 | Condition not met → no mutation                                     | PASS   |

---

## Summary

* Introduces `filter_list` processor for array element filtering
* Aligns naming with existing processors:

  * `iterate_on`
  * `keep_element_when`
* Supports:

  * Object and primitive arrays
  * In-place and target-based filtering
* Adds strong failure visibility with metadata


data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
```diff
@@ -0,0 +1,149 @@
...
+                if (sourceList == null) {
+                    LOG.debug("Source list at path [{}] is null, skipping event", config.getSource());
+                    continue;
+                }
+
+                final List<Object> filteredList = new ArrayList<>();
+                final JacksonEvent.Builder contextBuilder = JacksonEvent.builder()
+                        .withEventType("event");
+
+                for (final Object element : sourceList) {
```
bagmarnikhil, line 88: Each array element gets wrapped in a full JacksonEvent via builder.withData(contextMap).build(), which internally runs mapper.valueToTree(). For large arrays this creates
significant GC pressure and serialization overhead — the same concern @dlvenable raised in [the issue discussion](https://github.com/opensearch-project/data-prepper/issues/6610#issuecomment-4013763109). This matches the pattern used by the translate processor, which has reportedly caused performance issues.

For now this is probably acceptable since the ExpressionEvaluator requires an Event, but worth noting as a known limitation.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
```diff
@@ -0,0 +1,149 @@
...
+                    @SuppressWarnings("unchecked")
+                    final Map<String, Object> contextMap = element instanceof Map
+                            ? (Map<String, Object>) element
+                            : Collections.singletonMap("value", element);
+
+                    try {
+                        final Event elementEvent = contextBuilder
+                                .withData(contextMap)
```
bagmarnikhil, line 99: When evaluateConditional throws for a single element, that element is silently dropped from the result. This is a data loss scenario with no signal to the user beyond WARN-level logs

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
```diff
@@ -0,0 +1,149 @@
...
+                    continue;
+                }
+
+                if (sourceList == null) {
+                    LOG.debug("Source list at path [{}] is null, skipping event", config.getSource());
+                    continue;
+                }
+
+                final List<Object> filteredList = new ArrayList<>();
+                final JacksonEvent.Builder contextBuilder = JacksonEvent.builder()
```
bagmarnikhil: The JacksonEvent.Builder is created once outside the loop and reused via .withData() on each iteration. This works today because build() consumes the current data reference, but it's fragile — if the builder ever accumulates state across calls, this breaks silently. Safer to create the builder inside the loop

bagmarnikhil (commented):
Nice work — clean implementation with thorough test coverage. Two items to address: 
- when evaluateConditional throws for a single element in the loop (FilterListProcessor.java:103-110), that element is silently dropped from the output — this is a data loss risk, consider failing open or tagging the event; 

- JacksonEvent.Builder at line 93 is reused across loop iterations which is fragile — safer to create it inside the loop. Rest looks good.

kkondaka:
Looks good too. But  it looks like we are creating new processors each of the "list" operations. Maybe we should have a list processor that can filter and that manipulate list items and one of the manipulations could be converting simple list items to map. That way the other PR (https://github.com/opensearch-project/data-prepper/pull/6665) can be functionally merged into one processor that operates on lists?

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
nishantKadivar: For now, we have acknowledged the limitation and mark it as TODO.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
nishantKadivar: Thanks for calling this out.
To improve visibility for this case, I updated the processor to mark the event as failed (tags_on_failure) whenever evaluateConditional throws for one or more elements, and to attach diagnostic metadata:

- filter_list_processor_failed_elements_count

- filter_list_processor_failed_elements

This keeps current filtering behavior consistent with similar processors, while ensuring downstream processors and users can clearly detect and inspect element-level evaluation failures instead of relying only on WARN logs.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
```diff
@@ -0,0 +1,162 @@
...
+    private final String target;
+
+    @DataPrepperPluginConstructor
+    public FilterListProcessor(final PluginMetrics pluginMetrics, final FilterListProcessorConfig config, final ExpressionEvaluator expressionEvaluator) {
+        super(pluginMetrics);
+        this.config = config;
+        this.expressionEvaluator = expressionEvaluator;
+        this.target = config.getTarget() != null ? config.getTarget() : config.getSource();
+
+        if (config.getFilterListWhen() != null
```
dlvenable: Put validations in the `Config` classes themselves.

You can use an `@AssertTrue` from the Jakarta package. Other configs follow this pattern.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
```diff
@@ -0,0 +1,162 @@
...
+                && !expressionEvaluator.isValidExpressionStatement(config.getFilterListWhen())) {
+            throw new InvalidPluginConfigurationException(
+                    String.format("filter_list_when %s is not a valid expression statement. " +
+                                    "See https://opensearch.org/docs/latest/data-prepper/pipelines/expression-syntax/ for valid expression syntax",
+                            config.getFilterListWhen()));
+        }
+
+        if (!expressionEvaluator.isValidExpressionStatement(config.getKeepWhen())) {
```
dlvenable: Put this in the `Config` class.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
dlvenable: There are two problems with the `translate` processor approach. One is the performance. But the other issue is how we handle relative paths. There is a good discussion in #6609 about this same issue.

Basically, we need a way to decide relative versus absolute paths. Some processors are all absolute and some have relative paths. I propose in #6609 that we should use `/` to state absolute. And lack of that can be relative. Your `elementEvent` below will be a relative path with an absolute marker. This is a larger problem because it will lock us into the current behavior until a breaking change.

data-prepper-plugins/mutate-event-processors/src/test/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessorTest.java
```diff
@@ -0,0 +1,661 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.lenient;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class FilterListProcessorTest {
```
dlvenable, line 45: We should have at least one test that uses `@DataPrepperPluginTest`. You can refactor this test to use it. However it doesn't support mocked `ExpressionEvaluator` - you will get a real one.

Here are two examples in the project:

https://github.com/opensearch-project/data-prepper/blob/666c7269d87df7849d4f3d761c95a3507c7a0239/data-prepper-plugins/grok-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/grok/GrokProcessorIT.java#L48

https://github.com/opensearch-project/data-prepper/blob/7d5e8dea3f4d7e0c4700dc9bc8ef65ed1572fae5/data-prepper-plugins/drop-events-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/drop/DropEventsProcessorIT.java#L35

dlvenable (changes requested):
Thank you @yavmanis for this contribution!

data-prepper-plugins/mutate-event-processors/src/test/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessorTest.java
dlvenable: You would also be able to create configs using YAML this way. This ensures that the YAML can be consistent across versions.

yavmanis: Added `FilterListProcessorIT` using `@DataPrepperPluginTest` with a real ExpressionEvaluator, following the pattern from `DropEventsProcessorIT` and `GrokProcessorIT`.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
yavmanis: Thanks for the feedback! We've moved the expression validation into a validateExpressions(ExpressionEvaluator) method in `FilterListProcessorConfig`, following the pattern used by `ObfuscationProcessorConfig.validateObfuscateWhen(ExpressionEvaluator)`.

For context, our initial implementation had the validation inline in the processor constructor because that's the pattern followed by every other processor in the mutate-event-processors module — `ListToMapProcessor`, `MapToListProcessor`, `AddEntryProcessor`, `DeleteEntryProcessor`, `RenameKeyProcessor`, and `CopyValueProcessor `all validate expressions via `isValidExpressionStatement `directly in their constructors. `ObfuscationProcessorConfig `(in a different module) was the only config class we found that delegates this to a config method. That said, we agree the config-based approach is cleaner, so we've updated accordingly.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
yavmanis: `Target` defaulting has been moved into `FilterListProcessorConfig.getTarget()`, which now returns source when target is not set. 
Regarding` @AssertTrue` — the existing usages in this module are all for pure cross-field validations that don't require external dependencies. Our `getTarget()` never rejects anything — it always succeeds. It's a getter with a fallback default, not a validation. So `@AssertTrue` doesn't apply here.
Expression validation requires `ExpressionEvaluator`, which is a runtime dependency not available to config POJOs during Jakarta validation. No config class in the project uses `@AssertTrue` for expression validation.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
yavmanis: Thanks for flagging this. We understand  this could conflict with a future convention where `/` explicitly means absolute (as discussed in #6609). We'll keep this in mind and align with whatever convention is adopted at the framework level. For now, this is the only viable approach given the current `ExpressionEvaluator `API.

yavmanis:
> ## ⚠️ License Header Violations Found
> The following newly added files are missing required license headers:
> 
> * `data-prepper-plugins/mutate-event-processors/src/test/resources/org/opensearch/dataprepper/plugins/processor/mutateevent/filter_objects_by_status.yaml`
> * `data-prepper-plugins/mutate-event-processors/src/test/resources/org/opensearch/dataprepper/plugins/processor/mutateevent/filter_objects_to_target.yaml`
> * `data-prepper-plugins/mutate-event-processors/src/test/resources/org/opensearch/dataprepper/plugins/processor/mutateevent/filter_primitives_with_condition.yaml`
> 
> Please add the appropriate license header to each file and push your changes.
> 
> **See the license header requirements:** https://github.com/opensearch-project/data-prepper/blob/main/CONTRIBUTING.md#license-headers

We do not have headers specifically defined in these YAML files, as we observed that they were not included in any of the existing processor files. This was done to maintain consistency across processors.

Please let us know if it is mandatory to add headers in these files as well.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
dlvenable: @yavmanis , We should avoid pushing this bug further into processors. Using `/` is clearly indicating a JSON pointer at a root object.

I think these are our options at this point:

1. Resolve this correctly through JacksonEvent.
2. Create a throw-away configuration like `use_relative_pointer_on_keep_when`. The only allowable value would be `true` and this is not a default value so users must provide it.

```
- filter_list:
        source: "items"
        keep_when: '/status == "active"'
        filter_list_when: '/env == "production"'
        use_relative_pointer_on_keep_when: true
```

We would eventually remove it.

This name helps clarify that `filter_list_when` isn't using the relative pointer.

dlvenable:
@yavmanis , Be sure to also correct the headers.

The last thing I want to do before merging this is to resolve the issue with the relative pointers.

yavmanis:
> @yavmanis , Be sure to also correct the headers.
> 
> The last thing I want to do before merging this is to resolve the issue with the relative pointers.

I have added the license headers in the yaml files

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
yavmanis: @dlvenable Since we implemented this processor from the beginning, we have carefully handled all the relevant conditions, particularly those related to relative JSON pointers. Specifically, the **filter_list_when** condition resolves against the root-level fields, while the **keep_when** condition resolves against the element-level fields for each object in the list.

Based on our testing, both conditions are working as expected within their respective contexts. Therefore, we believe that introducing the flag `use_relative_pointer_on_keep_when: true` may not be necessary for resolving JSON pointers, as this behavior is already managed within the processor implementation.

Additionally, the behavior of both conditions is consistent with the **delete_entries** processor.

However, if you have a specific use case in mind that is not adequately handled by the current implementation and would require this flag, please let us know. We would be happy to review and address it.

yavmanis:
As discussed on the call, David will revisit the **filter_list** processor requirements and the current implementation, and will advise on how he would like to handle the processor’s behavior for condition evaluation at the root versus element level, as well as how this processor should align with the other processors.

We will make the necessary changes in the PR once we receive David’s final guidance on this.

cc: @nishantKadivar @bagmarnikhil

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+
+                final List<Object> filteredList = new ArrayList<>();
+                final List<Object> failedElements = new ArrayList<>();
+                int failedElementCount = 0;
+
+                for (final Object element : sourceList) {
+                    @SuppressWarnings("unchecked")
+                    final Map<String, Object> contextMap = element instanceof Map
+                            ? (Map<String, Object>) element
+                            : Collections.singletonMap("value", element);
```
graytaylor0, line 92: From this and the examples in the description it looks like we are using "/value" in the `keep_when` expression as a special name for primitive lists. Are we sure this is the right approach? At first I was worried this would conflict with actual fields named "value", but this checking for Map condition does handle that conflict. We just need to be very clear in the documentation for `keep_when` that "/value" is a special keyword for primitive lists.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+            try {
+                if (Objects.nonNull(config.getFilterListWhen()) && !expressionEvaluator.evaluateConditional(config.getFilterListWhen(), recordEvent)) {
+                    continue;
+                }
+
+                final List<Object> sourceList;
+                try {
+                    sourceList = recordEvent.get(config.getSource(), List.class);
+                } catch (final Exception e) {
+                    LOG.warn(EVENT, "Given source path [{}] is not valid on record [{}]",
```
graytaylor0: Can you SENSITIVE marker to this log? This is needed if we are logging the entire recordEvent. Also, may be useful having a metric on this error case as well, but not required.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+                if (failedElementCount > 0) {
+                    addTagsOnFailure(recordEvent);
+                    recordEvent.getMetadata().setAttribute(FAILED_ELEMENTS_COUNT_METADATA_KEY, failedElementCount);
+                    recordEvent.getMetadata().setAttribute(FAILED_ELEMENTS_METADATA_KEY, failedElements);
+                }
+
+                recordEvent.put(config.getTarget(), filteredList);
+
+            } catch (final Exception e) {
+                LOG.atError()
```
graytaylor0, line 128: Same here logging the full event requires SENSITIVE marker

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessorConfig.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @NotNull
+    @NotEmpty
+    @JsonProperty("keep_when")
+    @JsonPropertyDescription("An expression evaluated per element. Elements where this expression evaluates to true are kept. " +
+            "The expression is evaluated against each element of the array as if it were a standalone event.")
+    @ExampleValues({
+            @Example(value = "/status == \"active\"", description = "Keeps only elements where 'status' equals 'active'."),
+            @Example(value = "/score > 50", description = "Keeps only elements where 'score' is greater than 50.")
+    })
```
graytaylor0, line 60: Let's add examples for primitive lists where "/value" is used as keyword, and explain that in the description.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
graytaylor0: @dlvenable if `keep_when` syntax stays the same regardless, I'm not sure if adding a flag like this `use_relative_pointer_on_keep_when: true` is correct yet. Right now, as long as we are consistently saying "keep_when uses local element scope", and "filter_list_when" uses scope from the root of the Event", I think that is pretty clear.

dlvenable:
To promote consistency with other processors that iterate over elements, we can rename `source` to `iterate_on`. Data Prepper already has some precedent for using `element_when` to evaluate against the element and not the whole event. So we could also rename `keep_when` to `keep_element_when`. The other configuration options can stay the same.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
dlvenable: We had a discussion last week about renaming the fields to promote consistency. I suggested in another comment that we use `iterate_on` and `keep_element_when` to signal that these have similar behavior to the delete entries processor.

The add entries processor is unfortunately not handling this in a great way currently. But this PR would match delete entries with those name changes.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessorConfig.java
nishantKadivar: Updated FilterListProcessorConfig @JsonPropertyDescription and @ExampleValues for keep_element_when:
- Added explanation: "For primitive elements (strings, numbers, booleans), the element value is accessible via the key /value"
- Added examples:
  - /value > 50 (numeric primitives)
  - /value != "" (string primitives)

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/FilterListProcessor.java
nishantKadivar: Added SENSITIVE marker to warning logs that output full event context:
- Line 66: LOG.warn(EVENT, SENSITIVE, ...) — when source path is invalid
- Line 101: LOG.warn(EVENT, SENSITIVE, ...) — when keep_element_when evaluation fails

nishantKadivar:
> To promote consistency with other processors that iterate over elements, we can rename `source` to `iterate_on`. Data Prepper already has some precedent for using `element_when` to evaluate against the element and not the whole event. So we could also rename `keep_when` to `keep_element_when`. The other configuration options can stay the same.

Updated as suggested — renamed source to iterate_on and keep_when to keep_element_when to align with existing Data Prepper conventions. Other configuration options remain unchanged.

dlvenable (approved):
Thank you @yavmanis and @nishantKadivar !

bagmarnikhil (approved):
Thank you for the change!

Issue #6611, opened 2026-03-04
[FEATURE] Add map_entries processor for array element projection


bagmarnikhil:
**Is your feature request related to a problem? Please describe.**

There is no way to convert between primitive arrays and object arrays. Processors like add_entries and delete_entries with iterate_on require List<Map<String, Object>> and cannot operate on primitive arrays (strings, numbers, booleans). Conversely, there
is no way to extract a single field from each object in an array into a flat primitive array.

For example, ["alpha", "beta"] cannot be transformed into [{"name": "alpha"}, {"name": "beta"}], and [{"url": "http://a1"}, {"url": "http://a2"}] cannot be reduced to ["http://a1", "http://a2"].

**Describe the solution you'd like**

A new map_entries processor that projects each element of an array through a key mapping. The direction is inferred from the input type:

- **Primitive array in** → wraps each value into an object (e.g., "alpha" → {"name": "alpha"})
- **Object array in** → extracts the specified field into a primitive (e.g., {"url": "http://a1"} → "http://a1")

| Option | Required | Description |
|--------|----------|-------------|
| source | Yes | Key of the array to transform |
| target | No | Key to write the resulting array to. Defaults to source (in-place) |
| key | Yes | For primitive→object: the key name in the resulting object. For object→primitive: the key to extract |
| append_if_target_exists | No | When true, appends results to the existing target array instead of overwriting. Default false |
| map_entries_when | No | Expression evaluated against the root event. If false, the processor is skipped |

Example: Primitive → Object

Input: {"names": ["alpha", "beta"]}

```yaml
- map_entries:
    source: "/names"
    target: "/agents"
    key: "name"
```

Output: {"agents": [{"name": "alpha"}, {"name": "beta"}]}

Example: Object → Primitive (extract and merge)

Input: {"list_a": [{"url": "http://a1"}, {"url": "http://a2"}], "list_b": [{"url": "http://b1"}]}

```yaml
- map_entries:
    source: "/list_a"
    target: "/all_urls"
    key: "url"
- map_entries:
    source: "/list_b"
    target: "/all_urls"
    key: "url"
    append_if_target_exists: true
```

Output: {"all_urls": ["http://a1", "http://a2", "http://b1"]}

**Describe alternatives you've considered (Optional)**

- **Using copy_values with from_list/to_list:** This copies fields between lists of objects but produces a list of objects, not a flat primitive array. It also cannot handle the reverse direction (primitive → object).
[...]

PR #6665, opened 2026-03-22
Add wrap_entries processor to mutate-event-processors

nishantKadivar:
# Add `wrap_entries` processor to mutate-event-processors

## Description

Adds a new `wrap_entries` processor that wraps each element of a primitive array into an object using a configured key name. This enables downstream processors like `add_entries` and `delete_entries` with `iterate_on`, which require `List<Map<String, Object>>` and cannot operate on primitive arrays (strings, numbers, booleans).

For example, `["alice", "bob"]` becomes `[{"name": "alice"}, {"name": "bob"}]`.

## Issues Resolved

https://github.com/opensearch-project/data-prepper/issues/6611

## Configuration Options

| Option | Required | Type | Default | Description |
|--------|----------|------|---------|-------------|
| `source` | Yes | String | — | Key of the primitive array to transform (JSON Pointer) |
| `target` | No | String | `source` | Key to write the result to (defaults to in-place) |
| `key` | Yes | String | — | Key name in each resulting object |
| `exclude_null_empty_values` | No | Boolean | `false` | Filter out null/empty elements before wrapping |
| `append_if_target_exists` | No | Boolean | `false` | Append to existing target array instead of overwriting |
| `map_entries_when` | No | String | `null` | Conditional expression evaluated at root event level |
| `tags_on_failure` | No | List | `null` | Tags to add on processing failure |

## Check List
- [x] New functionality includes testing
- [x] New functionality has been documented
- [x] Commits are signed per the DCO using `--signoff`

## End-to-End Test Results

Tested manually using Data Prepper's `file` source and `file` sink.

### Pipeline Configuration

```yaml
map-entries-pipeline:
  source:
    file:
      path: "pipelines/input.json"
      format: "json"
      record_type: "event"
  processor:
    - wrap_entries:
        source: "/names"
        key: "name"
        map_entries_when: '/type == "users"'
    - wrap_entries:
        source: "/tags"
        key: "value"
        map_entries_when: '/type == "tagged"'
    - wrap_entries:
        source: "/items"
        target: "/inventory_items"
        key: "product"
        map_entries_when: '/type == "inventory"'
    - wrap_entries:
        source: "/scores"
        key: "score"
        map_entries_when: '/type == "grades"'
  sink:
    - file:
        path: "pipelines/output.json"
```

### Input (5 records)

```json
{"names": ["alice", "bob", "charlie"], "type": "users"}
{"tags": ["critical", "network", "security"], "source": "firewall", "type": "tagged"}
{"list_a": [{"url": "http://api.example.com/v1"}, {"url": "http://api.example.com/v2"}], "list_b": [{"url": "http://cdn.example.com/assets"}], "type": "urls"}
{"items": ["laptop", "monitor", "keyboard", "mouse"], "department": "IT", "type": "inventory"}
{"scores": [95, 82, 78, 91, 88], "subject": "math", "type": "grades"}
```

### Output (actual results)

```json
{"names":[{"name":"alice"},{"name":"bob"},{"name":"charlie"}],"type":"users"}
{"tags":[{"value":"critical"},{"value":"network"},{"value":"security"}],"source":"firewall","type":"tagged"}
{"list_a":[{"url":"http://api.example.com/v1"},{"url":"http://api.example.com/v2"}],"list_b":[{"url":"http://cdn.example.com/assets"}],"type":"urls"}
{"items":["laptop","monitor","keyboard","mouse"],"department":"IT","type":"inventory","inventory_items":[{"product":"laptop"},{"product":"monitor"},{"product":"keyboard"},{"product":"mouse"}]}
{"scores":[{"score":95},{"score":82},{"score":78},{"score":91},{"score":88}],"subject":"math","type":"grades"}
```

### E2E Scenarios

| # | Scenario | Result |
|---|----------|--------|
| 1 | String array in-place wrap (`type=users`) | PASS |
| 2 | String array wrap with different key (`type=tagged`) | PASS |
| 3 | Condition not met — no mutation (`type=urls`) | PASS |
| 4 | Separate target preserves original (`type=inventory`) | PASS |
| 5 | Integer array wrap (`type=grades`) | PASS |


data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/MapEntriesProcessor.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+        return records;
+    }
+
+    private void processEvent(final Event event) {
+        final String source = config.getSource();
+
+        if (!event.containsKey(source)) {
+            LOG.warn(EVENT, "Source key [{}] does not exist in event [{}], skipping.", source, event);
+            addTagsOnFailure(event);
```
bagmarnikhil: Is a missing source key really a failure? In a pipeline processing mixed event types, lots of events won't have this key — tagging all of them as failed seems noisy. The other mutate-event processors (SelectEntriesProcessor, FilterListProcessor) treat missing keys as a silent no-op. I'd drop the addTagsOnFailure here and make it a debug-level log instead, reserving failure tags for actual errors like "source exists but isn't a list."

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/MapEntriesProcessor.java
```diff
@@ -0,0 +1,172 @@
...
+        for (final Object element : sourceList) {
+            if (config.getExcludeNullEmptyValues()) {
+                if (element == null) {
+                    continue;
+                }
+                if (element instanceof String && ((String) element).isEmpty()) {
+                    continue;
+                }
+            }
+            final Map<String, Object> wrapped = new HashMap<>(1);
```
bagmarnikhil: Since this map is never modified after the put, Collections.singletonMap(key, element) would be a better fit — immutable, less memory, and communicates that it's a single-entry map.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/MapEntriesProcessor.java
```diff
@@ -0,0 +1,172 @@
...
+
+        if (config.getMapEntriesWhen() != null
+                && !expressionEvaluator.isValidExpressionStatement(config.getMapEntriesWhen())) {
+            throw new InvalidPluginConfigurationException(
+                    String.format("map_entries_when \"%s\" is not a valid expression statement. " +
+                            "See https://opensearch.org/docs/latest/data-prepper/pipelines/expression-syntax/ " +
+                            "for valid expression syntax", config.getMapEntriesWhen()));
+        }
+
+        if (config.getTagsOnFailure() != null) {
```
bagmarnikhil: Nit: none of the other mutate-event processors validate tag contents like this (AddEntryProcessor, SelectEntriesProcessor, etc.). Not wrong, but it's inconsistent

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/MapEntriesProcessor.java
```diff
@@ -0,0 +1,172 @@
...
+                if (element instanceof String && ((String) element).isEmpty()) {
+                    continue;
+                }
+            }
+            final Map<String, Object> wrapped = new HashMap<>(1);
+            wrapped.put(key, element);
+            result.add(wrapped);
+        }
+
+        if (result.isEmpty()) {
```
bagmarnikhil: When exclude_null_empty_values: true and all elements get filtered out, this early return leaves the original primitive array untouched. But if even one element
survives, the target gets overwritten with an object array. This means the same processor produces different output types depending on the data:
- Input [null, ""] → output stays [null, ""] (primitives)
- Input [null, "ok"] → output becomes [{"key": "ok"}] (objects)

Downstream processors like add_entries with iterate_on expect a consistent List<Map> — they'll break on the primitive case. I'd remove this early return and let it fall through to event.put(effectiveTarget, result) so the output is always [] when nothing survives filtering.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/MapEntriesProcessor.java
nishantKadivar: To be consistent with the other processor, I have removed tag content validation

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/MapEntriesProcessor.java
nishantKadivar: I have removed the if (result.isEmpty()) { return; } check and let it fall through to event.put(effectiveTarget, result). That way the output is always List<Map<String, Object>> — either [] or [{"key": "val"}, ...] — regardless of whether filtering removed everything.

kkondaka:
Looks good. Not sure about the name of the processor. Looks like it's doing list to list with each list item becoming a map?

nishantKadivar:
@kkondaka - The name map_entries was chosen because the processor maps each entry in the list into a key-value structure, aligning with the naming of other mutate-event processors. 
Open for suggestion

nishantKadivar:
> The following newly added files are missing required license headers:
> 
> data-prepper-plugins/mutate-event-processors/src/test/resources/org/opensearch/dataprepper/plugins/processor/mutateevent/map_entries_when_filters_records.yaml
> Please add the appropriate license header to each file and push your changes.
> 
> See the license header requirements: https://github.com/opensearch-project/data-prepper/blob/main/CONTRIBUTING.md#license-headers
> 

Regarding this violation, We do not have headers specifically defined in these YAML files, as we observed that they were not included in any of the existing processor files. This was done to maintain consistency across processors. Let me know if that needed to be included.

data-prepper-plugins/mutate-event-processors/src/test/java/org/opensearch/dataprepper/plugins/processor/mutateevent/MapEntriesProcessorTest.java
```diff
@@ -0,0 +1,295 @@
...
+
+    private MapEntriesProcessor createObjectUnderTest() {
+        return new MapEntriesProcessor(pluginMetrics, config, expressionEvaluator);
+    }
+
+    private Record<Event> createEvent(final Map<String, Object> data) {
+        return new Record<>(JacksonEvent.builder().withEventType("event").withData(data).build());
+    }
+
+    // --- Constructor validation delegation ---
```
dlvenable: Let's avoid these comments for sections. Remove them or use `@Nested` classes.

data-prepper-plugins/mutate-event-processors/src/test/resources/org/opensearch/dataprepper/plugins/processor/mutateevent/map_entries_when_filters_records.yaml
```diff
@@ -0,0 +1,10 @@
+test-pipeline:
```
dlvenable: This also needs a header.

dlvenable:
> Looks good. Not sure about the name of the processor. Looks like it's doing list to list with each list item becoming a map?

I think this is a good point. "Map entries" sounds like it is running a map function. The word map can be confused with a verb.

data-prepper-plugins/mutate-event-processors/src/main/java/org/opensearch/dataprepper/plugins/processor/mutateevent/MapEntriesProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+        for (final Object element : sourceList) {
+            if (config.getExcludeNullEmptyValues()) {
+                if (element == null) {
+                    continue;
+                }
+                if (element instanceof String && ((String) element).isEmpty()) {
+                    continue;
+                }
+            }
+            result.add(Collections.singletonMap(key, element));
```
dlvenable: I suspect that `JacksonEvent` will take this as-is and not remap. The `singletonMap` is immutable and thus would fail to mutate in downstream processors. Let's create a new map and put that.

dlvenable (changes requested):
Thank you @nishantKadivar for this contribution. I have a few comments. We should change the name as well.

dlvenable:
> > Looks good. Not sure about the name of the processor. Looks like it's doing list to list with each list item becoming a map?
> 
> I think this is a good point. "Map entries" sounds like it is running a map function. The word map can be confused with a verb.

The key verb you use in the README is "wrap." This processor name should start with `wrap_` since that is the action it is taking.

data-prepper-plugins/mutate-event-processors/src/test/java/org/opensearch/dataprepper/plugins/processor/mutateevent/WrapEntriesProcessorTest.java
```diff
@@ -0,0 +1,295 @@
...
+
+        verify(config).validateExpressions(expressionEvaluator);
+    }
+
+
+    // --- Basic wrapping ---
+
+    @Test
+    void doExecute_with_string_array_wraps_each_element_into_object_in_place() {
+        final Record<Event> record = createEvent(Map.of("names", Arrays.asList("alpha", "beta")));
```
dlvenable, line 78: Let's add some test cases when the names are not primitives.

e.g.

```
List.of(Map.of("name": "alpha"), Map.of("name": "beta"))
```

and

```
List.of(List.of("alpha1": "beta1"), List.of("alpha2": "beta2"))
```

nishantKadivar: I have added test cases for both the scenarios.

data-prepper-plugins/mutate-event-processors/src/test/java/org/opensearch/dataprepper/plugins/processor/mutateevent/WrapEntriesProcessorTest.java
```diff
@@ -292,4 +278,50 @@ void doExecute_with_multiple_records_wraps_only_matching_and_leaves_non_matching
         final List<?> unmatchedOutput = result.get(1).getData().get("/names", List.class);
         assertThat(unmatchedOutput, equalTo(Arrays.asList("gamma")));
     }
+
+    @Test
+    void doExecute_with_map_elements_wraps_each_map_into_object() {
```
dlvenable, line 283: Is this the desirable behavior? I'm ok to approve now, but we should follow up and discuss if we want to have this before we release 2.16.

dlvenable (approved):

data-prepper-plugins/mutate-event-processors/src/test/java/org/opensearch/dataprepper/plugins/processor/mutateevent/WrapEntriesProcessorTest.java
nishantKadivar: Hello @dlvenable , 
I have raised an issue for the discussion about the wrap_entries behaviour #6753.

kkondaka (approved):

PR #6671, opened 2026-03-24
Passing http request headers in the event metadata for http source

### LINKED ISSUE #6239: Support ingesting http headers with http source
Current http source ignores the header infomation

Should provide an option to ingest header information along with the data and put it in the data or metadata

**Additional context**

divbok:
### Description
Adds a new `metadata_headers` configuration option to the HTTP source plugin that allows users to extract HTTP request headers and attach them as event metadata. This enables downstream processors and sinks to route, filter, or enrich events based on HTTP headers.
#### How it works
- **Whitelist mode**: Configure specific header names to extract:
  ```yaml
  metadata_headers: ["X-Tenant-Id", "X-Region"]
- Disabled (default): When metadata_headers is not set or null, no headers are extracted.
 #### Header key normalization
- All header keys are lowercased in event metadata regardless of how they were sent by the client or configured in the YAML. This is because Armeria (the underlying HTTP server) normalizes all header names to lowercase per the HTTP/2 specification. For example, a header sent as `X-Tenant-Id` is stored in metadata as `x-tenant-id`.
- Sensitive header filtering
A hardcoded blocklist of sensitive headers is always filtered out, regardless of configuration. This includes authorization, cookie, x-api-key, x-amz-security-token, x-amz-credential, and others.
- Multi-value headers
Headers with multiple values (e.g., X-Forwarded-For appearing twice) are stored as a List<String>. Single-value headers are stored as a plain String.
- Current limitations
This feature is currently only supported for the in-memory buffer (BlockingBuffer). The byte buffer / Kafka buffer path does not propagate metadata headers yet.
#### Example pipeline
  ```yaml

log-pipeline:
  source:
    http:
      port: 2021
      path: "/log/ingest"
      metadata_headers: ["X-Tenant-Id", "X-Region"]
  route:
    - tenant_alpha: 'getMetadata("headers/x-tenant-id") == "team-alpha"'
    - us_west: 'getMetadata("headers/x-region") == "us-west-2"'
  sink:
    - file:
        path: "/tmp/alpha.log"
        routes: ["tenant_alpha"]
    - file:
        path: "/tmp/us_west.log"
        routes: ["us_west"]
```
```bash
curl -X POST http://localhost:2021/log/ingest \
  -H "Content-Type: application/json" \
  -H "X-Tenant-Id: team-alpha" \
  -H "X-Region: us-west-2" \
  -d '[{"message": "hello"}]'
  ```
### Issues Resolved
Resolves #6239 
 
### Check List

### DIFF
--- a/data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/HTTPSourceConfig.java
+++ b/data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/HTTPSourceConfig.java
@@ -9,6 +9,9 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import org.opensearch.dataprepper.http.BaseHttpServerConfig;
 import org.opensearch.dataprepper.model.configuration.PluginModel;
 
+import java.util.Collections;
+import java.util.List;
+
 public class HTTPSourceConfig extends BaseHttpServerConfig {
 
     static final String DEFAULT_LOG_INGEST_URI = "/log/ingest";
@@ -27,7 +30,15 @@ public class HTTPSourceConfig extends BaseHttpServerConfig {
     @JsonProperty("codec")
     private PluginModel codec;
 
+    @JsonProperty("metadata_headers")
+    private List<String> metadataHeaders = Collections.emptyList();
+
     public PluginModel getCodec() {
         return codec;
     }
+
+    public List<String> getMetadataHeaders() {
+        return metadataHeaders;
+    }
+
 }
--- a/data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/HttpHeaderExtractor.java
+++ b/data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/HttpHeaderExtractor.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ */
+
+package org.opensearch.dataprepper.plugins.source.loghttp;
+
+import com.linecorp.armeria.common.AggregatedHttpRequest;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class HttpHeaderExtractor {
+
+    static final Set<String> SENSITIVE_HEADERS = Set.of(
+            "authorization",
+            "proxy-authorization",
+            "cookie",
+            "set-cookie",
+            "www-authenticate",
+            "proxy-authenticate",
+            "x-api-key",
+            "x-csrf-token",
+            "x-xsrf-token",
+            "x-auth-token",
+            "x-amz-security-token",
+            "x-amz-credential"
+    );
+
+    private final Collection<String> metadataHeaders;
+
+    public HttpHeaderExtractor(@Nonnull final Collection<String> metadataHeaders) {
+        this.metadataHeaders = metadataHeaders;
+    }
+
+    public Map<String, Object> extractHeaders(final AggregatedHttpRequest aggregatedHttpRequest) {
+        if (metadataHeaders.isEmpty()) {
+            return Collections.emptyMap();
+        }
+
+        final Set<String> headerNames = metadataHeaders.stream()
+                .map(String::toLowerCase)
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+
+        final Map<String, Object> headers = new HashMap<>();
+        for (String headerName : headerNames) {
+            if (isSensitiveHeader(headerName)) {
+                continue;
+            }
+            List<String> values = aggregatedHttpRequest.headers().getAll(headerName);
+            if (!values.isEmpty()) {
+                headers.put(headerName, values.size() == 1 ? values.get(0) : Collections.unmodifiableList(values));
+            }
+        }
+
+        return headers;
+    }
+
+    static boolean isSensitiveHeader(final String headerName) {
+        return SENSITIVE_HEADERS.contains(headerName.toLowerCase());
+    }
+}
--- a/data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
+++ b/data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
@@ -28,13 +28,15 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Collections;
+import java.util.Map;
 import java.util.UUID;
 import java.util.stream.Collectors;
-
 
 /*
 * A HTTP service for log ingestion to be executed by BlockingTaskExecutor.
 */
+
 @Blocking
 public class LogHTTPService {
     private static final int SERIALIZATION_OVERHEAD = 1024;
@@ -60,22 +62,32 @@ public class LogHTTPService {
     private final Timer requestProcessDuration;
     private Integer bufferMaxRequestLength;
     private Integer bufferOptimalRequestLength;
+    private final HttpHeaderExtractor httpHeaderExtractor;
 
     public LogHTTPService(final int bufferWriteTimeoutInMillis,
                           final Buffer<Record<Log>> buffer,
                           final PluginMetrics pluginMetrics,
-                          final InputCodec codec) {
+                          final InputCodec codec,
+                          final HttpHeaderExtractor httpHeaderExtractor) {
         this.buffer = buffer;
         this.bufferWriteTimeoutInMillis = bufferWriteTimeoutInMillis;
         this.bufferMaxRequestLength = buffer.getMaxRequestSize().isPresent() ? buffer.getMaxRequestSize().get(): null;
         this.bufferOptimalRequestLength = buffer.getOptimalRequestSize().isPresent() ? buffer.getOptimalRequestSize().get(): null;
         this.codec = codec;
+        this.httpHeaderExtractor = httpHeaderExtractor;
         requestsReceivedCounter = pluginMetrics.counter(REQUESTS_RECEIVED);
         successRequestsCounter = pluginMetrics.counter(SUCCESS_REQUESTS);
         requestsOverOptimalSizeCounter = pluginMetrics.counter(REQUESTS_OVER_OPTIMAL_SIZE);
         requestsOverMaximumSizeCounter = pluginMetrics.counter(REQUESTS_OVER_MAXIMUM_SIZE);
         payloadSizeSummary = pluginMetrics.summary(PAYLOAD_SIZE);
         requestProcessDuration = pluginMetrics.timer(REQUEST_PROCESS_DURATION);
+    }
+
+    public LogHTTPService(final int bufferWriteTimeoutInMillis,
+                          final Buffer<Record<Log>> buffer,
+                          final PluginMetrics pluginMetrics,
+                          final InputCodec codec) {
+        this(bufferWriteTimeoutInMillis, buffer, pluginMetrics, codec, new HttpHeaderExtractor(Collections.emptySet()));
     }
 
     @Post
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

File: data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java [hunk 1]
```diff
@@ -152,6 +186,35 @@ HttpResponse processRequest(final AggregatedHttpRequest aggregatedHttpRequest) t
         return HttpResponse.of(HttpStatus.OK);
     }
 
+    private Map<String, Object> extractHeaders(final AggregatedHttpRequest aggregatedHttpRequest) {
+        if (metadataHeaders == null || metadataHeaders.isEmpty()) {
+            return Collections.emptyMap();
+        }
+
+        final boolean includeAll = metadataHeaders.size() == 1 && "*".equals(metadataHeaders.get(0));
```
dlvenable: I'm not sure about supporting `*` without supporting regex in general. I think if we wanted this, we should add `metadata_headers_regex` instead. This would be consistent with other processors that support literals and regex.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
```diff
@@ -152,6 +186,35 @@ HttpResponse processRequest(final AggregatedHttpRequest aggregatedHttpRequest) t
...
+        final Set<String> headerNames = includeAll
+                ? aggregatedHttpRequest.headers().names().stream().map(Object::toString).collect(Collectors.toSet())
+                : metadataHeaders.stream().map(String::toLowerCase).collect(Collectors.toCollection(LinkedHashSet::new));
+
+        final Map<String, Object> headers = new HashMap<>();
+        for (String headerName : headerNames) {
+            if (isSensitiveHeader(headerName)) {
+                LOG.warn("Skipping sensitive header '{}' from metadata_headers configuration", headerName);
```
dlvenable: I think we don't need any logging here. Maybe a TRACE if you want.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
```diff
@@ -152,6 +186,35 @@ HttpResponse processRequest(final AggregatedHttpRequest aggregatedHttpRequest) t
...
+            List<String> values = aggregatedHttpRequest.headers().getAll(headerName);
+            if (!values.isEmpty()) {
+                headers.put(headerName, values.size() == 1 ? values.get(0) : new ArrayList<>(values));
+            }
+        }
+
+        return headers;
+    }
+
+    static boolean isSensitiveHeader(final String headerName) {
```
dlvenable: Put this in its own class and test it there.

data-prepper-plugins/http-source/src/test/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPServiceTest.java
```diff
@@ -391,6 +400,154 @@ void chunking_with_4mb() throws Exception {
...
+    })
+    public void testIsSensitiveHeader(String headerName, boolean expected) {
+        assertThat(LogHTTPService.isSensitiveHeader(headerName), equalTo(expected));
+    }
+
+    @Test
+    public void processRequestAttachesHeadersToEventMetadata() throws Exception {
+        final String tenantId = UUID.randomUUID().toString();
+        final String region = UUID.randomUUID().toString();
+        BlockingBuffer<Record<Log>> blockingBuffer = new BlockingBuffer<>(TEST_BUFFER_CAPACITY, 8, "test-pipeline");
```
dlvenable: Use a mocked `Buffer` instead. Then you don't need to use `read`. Verify that it is called and use an `ArgumentCaptor`.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
```diff
@@ -152,6 +186,35 @@ HttpResponse processRequest(final AggregatedHttpRequest aggregatedHttpRequest) t
         return HttpResponse.of(HttpStatus.OK);
     }
 
+    private Map<String, Object> extractHeaders(final AggregatedHttpRequest aggregatedHttpRequest) {
```
dlvenable: I tend to think that this should go into it's own class. You can unit test this more easily. Then verify an interaction with the mock in the unit tests for `LogHTTPService`.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
```diff
@@ -171,13 +234,13 @@ private void writeChunkedBody(final String chunk) {
...
+    private Record<Log> buildRecordLog(final String json, final Map<String, Object> headerAttributes) {
+        final JacksonLog.Builder builder = JacksonLog.builder()
                 .withData(json)
-                .getThis()
-                .build();
-
-        return new Record<>(log);
+                .getThis();
+        if (!headerAttributes.isEmpty()) {
+            builder.withEventMetadataAttributes(headerAttributes);
```
dlvenable: This approach will put all the headers as metadata in the root of the metadata. I wonder if it would make more sense to have these under `headers` in the metadata. Do we support nested metadata? If not, then I guess this is the approach we should take.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
divbok: I will add the regex config in a follow up PR

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/HTTPSourceConfig.java
```diff
@@ -27,7 +29,15 @@ public String getDefaultPath() {
     @JsonProperty("codec")
     private PluginModel codec;
 
+    @JsonProperty("metadata_headers")
+    private List<String> metadataHeaders;
```
oeyh: Suggest using empty list as default. Otherwise it's null by default and will need null check wherever it's used.
```suggestion
    private List<String> metadataHeaders  = Collections.emptyList();
```

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
```diff
@@ -92,6 +99,7 @@ public HttpResponse doPost(final ServiceRequestContext serviceRequestContext, fi
 
     HttpResponse processRequest(final AggregatedHttpRequest aggregatedHttpRequest) throws Exception {
         final HttpData content = aggregatedHttpRequest.content();
+        final Map<String, Object> extractedHeaders = httpHeaderExtractor.extractHeaders(aggregatedHttpRequest);
```
oeyh: All records in a batch share the exact same `extractedHeaders` map instance in their metadata. If a processor downstream somehow mutates the map, it will affect other records in the batch, might cause a surprise. We can wrap extractedHeaders in Collections.unmodifiableMap() to prevents accidental in-place mutation.

data-prepper-plugins/http-source/src/test/java/org/opensearch/dataprepper/plugins/source/loghttp/HTTPSourceConfigTest.java
```diff
@@ -21,5 +27,14 @@ void testDefault() {
         assertEquals(HTTPSourceConfig.DEFAULT_LOG_INGEST_URI, sourceConfig.getPath());
         assertEquals(HTTPSourceConfig.DEFAULT_PORT, sourceConfig.getDefaultPort());
         assertEquals(HTTPSourceConfig.DEFAULT_LOG_INGEST_URI, sourceConfig.getDefaultPath());
+        assertNull(sourceConfig.getMetadataHeaders());
```
dlvenable: I believe this is causing the failure:

```
HTTPSourceConfigTest > testDefault() FAILED
    org.opentest4j.AssertionFailedError: expected: <null> but was: <[]>
```

This is now an empty collection by default which is good.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
```diff
@@ -171,13 +185,13 @@ private void writeChunkedBody(final String chunk) {
...
-        final JacksonLog log = JacksonLog.builder()
+    private Record<Log> buildRecordLog(final String json, final Map<String, Object> headerAttributes) {
+        final JacksonLog.Builder builder = JacksonLog.builder()
                 .withData(json)
-                .getThis()
-                .build();
-
-        return new Record<>(log);
+                .getThis();
+        if (!headerAttributes.isEmpty()) {
```
dlvenable: How can we consolidate this block with line 138? We seem to have a split in how we create records that could result in discrepancies.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
```diff
@@ -28,13 +28,11 @@
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Collections;
+import java.util.Map;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
-
-/*
```
dlvenable, line 35: Did you mean to remove this?

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
```diff
@@ -78,6 +79,13 @@ public LogHTTPService(final int bufferWriteTimeoutInMillis,
         requestProcessDuration = pluginMetrics.timer(REQUEST_PROCESS_DURATION);
     }
 
+    public LogHTTPService(final int bufferWriteTimeoutInMillis,
+                          final Buffer<Record<Log>> buffer,
+                          final PluginMetrics pluginMetrics,
+                          final InputCodec codec) {
+        this(bufferWriteTimeoutInMillis, buffer, pluginMetrics, codec, new HttpHeaderExtractor(null));
```
dlvenable: You should use an empty collection for consistency. `new HttpHeaderExtractor()` or `new HttpHeaderExtractor(Collections.emptySet())`

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/HttpHeaderExtractor.java
```diff
@@ -0,0 +1,73 @@
...
+            "x-xsrf-token",
+            "x-auth-token",
+            "x-amz-security-token",
+            "x-amz-credential"
+    );
+
+    private final List<String> metadataHeaders;
+
+    public HttpHeaderExtractor(final List<String> metadataHeaders) {
+        this.metadataHeaders = metadataHeaders;
```
dlvenable, line 45: Require this to be non-null.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/HttpHeaderExtractor.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public Map<String, Object> extractHeaders(final AggregatedHttpRequest aggregatedHttpRequest) {
+        if (metadataHeaders == null || metadataHeaders.isEmpty()) {
```
dlvenable: Remove `metadataHeaders == null` after other changes I suggested.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/HttpHeaderExtractor.java
```diff
@@ -0,0 +1,73 @@
...
+            "x-csrf-token",
+            "x-xsrf-token",
+            "x-auth-token",
+            "x-amz-security-token",
+            "x-amz-credential"
+    );
+
+    private final List<String> metadataHeaders;
+
+    public HttpHeaderExtractor(final List<String> metadataHeaders) {
```
dlvenable: This can be `Collection<String>` You don't require ordering.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/LogHTTPService.java
divbok: Added it back.

data-prepper-plugins/http-source/src/main/java/org/opensearch/dataprepper/plugins/source/loghttp/HttpHeaderExtractor.java
divbok: Done

oeyh (approved):

PR #6672, opened 2026-03-24
Add configurable metric_timestamp_source, granularity, stable host ID, and NodeOperationDetail dedup for APM service map processor

### LINKED ISSUE #6710: APM service map metrics: cardinality explosion, Prometheus compatibility, and late-span data loss
## Summary

The `otel_apm_service_map` processor has several issues with metric generation when writing to Prometheus-compatible sinks (e.g., Amazon Managed Prometheus).

## Background: How Prometheus expects metrics

- **Remote write is type-agnostic.** Every sample is just labels + `(timestamp, value)`. Prometheus TSDB does not know if a sample is gauge, counter, or histogram.
- **Counters must be cumulative.** `rate()` works by diffing consecutive values. If the value drops (e.g., 15 → 5), Prometheus treats it as a counter reset and applies incorrect compensation.
- **Same-timestamp writes are rejected.** Two samples for the same series with the same timestamp but different values → second write is rejected with `ErrDuplicateSampleForTimestamp` ([source](https://github.com/prometheus/prometheus/blob/main/tsdb/head_append.go#L658)). Identical values are silently accepted as a no-op.
- **`isMonotonic` only exists on Sum metrics.** Gauge, Histogram, Summary do not have this flag. It distinguishes counters (only go up) from non-monotonic sums (can go up and down).

## Issue 1: `randomKey` UUID causes unbounded cardinality explosion

Every metric includes a random UUID label (`ApmServiceMapMetricsUtil.java:277`):

```java
labelsWithRandomKey.put("randomKey", UUID.randomUUID().toString());
```

This creates ~120,000 new throwaway series per hour (5 hosts × 20 services × 10 operations × 4 metrics × 30 windows/hr). TSDB costs grow without bound, and `rate()`/`increase()` cannot work since each series has only one data point.

**`randomKey` is currently masking Issues 2, 3, and 4.** Since every write has unique labels, Prometheus never sees consecutive values or conflicting writes. These issues will surface once `randomKey` is removed.

**Related PR:** #6672 (replaces `randomKey` with stable `trace_processor_host_id`)

## Issue 2: Delta temporality is not supported by Prometheus

All metrics use `AGGREGATION_TEMPORALITY_DELTA` with `isMonotonic(true)` (`ApmServiceMapMetricsUtil.java:284-286`). This says "I am a counter that only reports increments per window" — but Prometheus expects counters to be cumulative running totals.

Once series become stable (after removing `randomKey`), Prometheus will see values like `15 → 5` between windows and treat them as counter resets, producing incorrect rates.

If the intent is per-window counts, consider using `Gauge` type or non-monotonic Sum instead.

## Issue 3: Late-arriving spans cause silent data loss

**Only occurs once `randomKey` is removed.** When a late span produces a metric with the same labels + timestamp as one already written, Prometheus rejects the second write ([`head_append.go:658-669`](https://github.com/prometheus/prometheus/blob/main/tsdb/head_append.go#L658)):

- **Same timestamp, same value:** silently accepted (no-op)
- **Same timestamp, different value:** rejected (`ErrDuplicateSampleForTimestamp`)

At low late-span rates (0.01–0.1%), data loss is under 0.1%. At 5%+ late rates (backpressure), high-traffic operations could see ~4.5% undercount.

## Issue 4: No multi-host protection by default

Without `randomKey`, multiple Data Prepper hosts processing the same service + operation produce identical metric series. Prometheus sees interleaved values from different hosts, causing counter resets and incorrect aggregations.

**Related PR:** #6672 (adds `trace_processor_host_id`)

## Suggested approaches

1. **Replace `randomKey` with a stable host identifier** (addressed by #6672)

2. **Evaluate temporality and metric type** — if per-window counts are the intent, consider Gauge or non-monotonic Sum. If true counters are desired, switch to cumulative temporality with running totals.

3. **To make these true Prometheus counters**, the processor would need to maintain a running total in memory across flush windows instead of resetting each window. For example, instead of emitting delta values `10, 15, 5` across three windows, emit cumulative values `10, 25, 30`. This requires:
   - A `Map<MetricKey, Long>` running total that persists across window rotations
   - Accumulating each window's count: `runningTotal.merge(key, windowCount, Long::sum)`
   - Emitting the running total as the value with `AGGREGATION_TEMPORALITY_CUMULATIVE`
   - Setting `startTimestamp` to when the key was first created (not equal to `timestamp`)
   
   The tradeoff is memory — the running total map grows with cardinality and never shrinks unless an expiration mechanism is added (similar to `metrics_expiration` in the [OTel Collector spanmetrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector)). On restart, all totals are lost and Prometheus sees a counter reset — this is expected and handled correctly by `rate()`.

4. **Consider using processor timestamp instead of span `endTime`** — in healthy pipelines where 95%+ of spans arrive on time, the processor timestamp is within seconds of the span's `endTime`, so time attribution is nearly identical. Late spans would get attributed to processing time instead of their actual time, but these are the spans that would otherwise be rejected and lost entirely. This is the same approach used by the [OTel Collector spanmetrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector), which uses `clock.Now()` at flush time for all metric timestamps.

5. **Consolidate error/fault into a label on the request metric** instead of emitting separate `error` and `fault` metrics. Currently the processor emits three separate Sum metrics per key: `request`, `error`, and `fault`. This triples the number of time series. Instead, add a `status` label to a single `request` metric:

   Current approach (3 separate series per key):
   ```
   request{service="orders", operation="POST /checkout"} 100
   error{service="orders", operation="POST /checkout"} 3
   fault{service="orders", operation="POST /checkout"} 1
   ```

   Suggested approach (1 metric with status label):
   ```
   request{service="orders", operation="POST /checkout", status="ok"} 96
   request{service="orders", operation="POST /checkout", status="error"} 3
   request{service="orders", operation="POST /checkout", status="fault"} 1
   ```

   This reduces series count by ~3x and allows simpler queries:
   - Total requests: `sum by(service, operation)(request)`
   - Error rate: `request{status="error"} / ignoring(status) sum by(service, operation)(request)`
   - All non-ok: `request{status!="ok"}`

6. **Document the late-span rejection behavior** and expected data-loss characteristics.

## Environment

- Data Prepper version: latest main branch
- Sink: Amazon Managed Prometheus / any Prometheus-compatible TSDB

vamsimanohar:
## Summary

Enhances the `otel_apm_service_map` processor with configurable timestamp handling, stable host identification, and reduced storage overhead for service map events.

### Changes

- **`metric_timestamp_source`** config option (default: `arrival_time`)
  - `arrival_time`: Uses `clock.instant()` at window evaluation — all spans in a window share the same timestamp, eliminating late-span collision risk in Prometheus/AMP
  - `span_end_time`: Uses the span's `endTime` field — provides span-aligned timestamps but risks `ErrDuplicateSampleForTimestamp` for late-arriving spans

- **`metric_timestamp_granularity`** config option (default: `seconds`) — applies to metrics only
  - `seconds`: Truncates metric timestamps to second boundaries (1s collision window in `span_end_time` mode)
  - `minutes`: Truncates to minute boundaries (60s collision window — use only if coarser granularity is acceptable)

- **Service map timestamps** — independent of metric config
  - NodeOperationDetail events always use **span end time truncated to minutes**
  - Reflects actual interaction time (not processing time)
  - Minutes truncation reduces OpenSearch document count while providing sufficient granularity for dashboard time-range queries
  - Late-arriving spans may produce duplicate documents, but `dedup nodeConnectionHash` handles this at query time

- **Stable `service_map_processor_host_id`** label replaces random UUID
  - SHA-256 hash of hostname (via shared `HostContext` in `data-prepper-api`), truncated to 16 hex chars
  - Stable across restarts, consistent across all components

- **NodeOperationDetail dedup** across all traces in a window
  - Same service relationship from different traces now emits one document per window instead of one per span
  - Dedup by full object equality including timestamp (same minute = deduped)

- **Shared `HostContext`** utility in `data-prepper-api`
  - `HostContext.getHostname()` returns hostname or `"unknown"` gracefully (no crash on resolution failure)

- **Test constructor defaults** match production config (`ARRIVAL_TIME`)

### Configuration Example

```yaml
processor:
  - otel_apm_service_map:
      window_duration: 60s
      metric_timestamp_source: arrival_time
      metric_timestamp_granularity: seconds
      group_by_attributes:
        - "deployment.environment"
```

## Test Plan

- [x] All existing unit tests pass
- [x] New tests for `MetricTimestampSource`, `MetricTimestampGranularity`, `HostContext`, and config defaults
- [x] Test: service map events always use span end time truncated to minutes (even in `arrival_time` mode)
- [x] Test: `arrival_time` mode uses clock.instant() for metrics
- [x] Test: `span_end_time` mode uses span endTime for metrics
- [x] Verified NodeOperationDetail dedup: each window emits exactly one document per unique relationship hash per minute
- [x] Unified anchor timestamp parameter (addressed review feedback — single `anchorTimestamp` instead of separate sum/histogram params)

## Issues Resolved

Closes #6710

### DIFF
--- a/data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
+++ b/data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
@@ -45,9 +45,14 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import org.opensearch.dataprepper.model.host.HostContext;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -92,7 +97,10 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
     private static Clock clock;
 
     private final int thisProcessorId;
+    private final String hostId;
     private final List<String> groupByAttributes;
+    private final MetricTimestampSource metricTimestampSource;
+    private final MetricTimestampGranularity metricTimestampGranularity;
     private final EventFactory eventFactory;
 
     @DataPrepperPluginConstructor
@@ -107,7 +115,9 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
                 pipelineDescription.getNumberOfProcessWorkers(),
                 eventFactory,
                 pluginMetrics,
-                config.getGroupByAttributes());
+                config.getGroupByAttributes(),
+                config.getMetricTimestampSource(),
+                config.getMetricTimestampGranularity());
     }
 
     OTelApmServiceMapProcessor(final Duration windowDuration,
@@ -116,7 +126,8 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
                                final int processWorkers,
                                final EventFactory eventFactory,
                                final PluginMetrics pluginMetrics) {
-        this(windowDuration, databasePath, clock, processWorkers, eventFactory, pluginMetrics, Collections.emptyList());
+        this(windowDuration, databasePath, clock, processWorkers, eventFactory, pluginMetrics,
+                Collections.emptyList(), MetricTimestampSource.SPAN_END_TIME, MetricTimestampGranularity.SECONDS);
     }
 
     OTelApmServiceMapProcessor(final Duration windowDuration,
@@ -126,9 +137,37 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
                                final EventFactory eventFactory,
                                final PluginMetrics pluginMetrics,
                                final List<String> groupByAttributes) {
+        this(windowDuration, databasePath, clock, processWorkers, eventFactory, pluginMetrics,
+                groupByAttributes, MetricTimestampSource.SPAN_END_TIME, MetricTimestampGranularity.SECONDS);
+    }
+
+    OTelApmServiceMapProcessor(final Duration windowDuration,
+                               final File databasePath,
+                               final Clock clock,
+                               final int processWorkers,
+                               final EventFactory eventFactory,
+                               final PluginMetrics pluginMetrics,
+                               final List<String> groupByAttributes,
+                               final MetricTimestampSource metricTimestampSource) {
+        this(windowDuration, databasePath, clock, processWorkers, eventFactory, pluginMetrics,
+                groupByAttributes, metricTimestampSource, MetricTimestampGranularity.SECONDS);
+    }
+
+    OTelApmServiceMapProcessor(final Duration windowDuration,
+                               final File databasePath,
+                               final Clock clock,
+                               final int processWorkers,
+                               final EventFactory eventFactory,
+                               final PluginMetrics pluginMetrics,
+                               final List<String> groupByAttributes,
+                               final MetricTimestampSource metricTimestampSource,
+                               final MetricTimestampGranularity metricTimestampGranularity) {
         super(pluginMetrics);
 
+        this.hostId = resolveHostId();
         this.groupByAttributes = groupByAttributes != null ? Collections.unmodifiableList(groupByAttributes) : Collections.emptyList();
+        this.metricTimestampSource = metricTimestampSource != null ? metricTimestampSource : MetricTimestampSource.ARRIVAL_TIME;
+        this.metricTimestampGranularity = metricTimestampGranularity != null ? metricTimestampGranularity : MetricTimestampGranularity.SECONDS;
 
         this.eventFactory = eventFactory;
         OTelApmServiceMapProcessor.clock = clock;
@@ -375,7 +414,9 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
 
         final EphemeralSpanDecorations ephemeralDecorations = new EphemeralSpanDecorations();
 
-        final Map<MetricKey, MetricAggregationState> metricsStateByKey = new HashMap<>();
+        final Map<MetricKey, MetricAggregationState> sumStateByKey = new HashMap<>();
+        final Map<MetricKey, MetricAggregationState> histogramStateByKey = new HashMap<>();
+        final Set<NodeOperationDetail> dedupedNodeDetails = new HashSet<>();
 
         final Map<String, Collection<SpanStateData>> previousSpansByTraceId = buildSpansByTraceIdMap(previousWindow);
         final Map<String, Collection<SpanStateData>> currentSpansByTraceId = buildSpansByTraceIdMap(currentWindow);
@@ -388,11 +429,24 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
             if (!traceData.getProcessingSpans().isEmpty()) {
                 decorateSpansInTraceWithEphemeralStorage(traceData);
 
-                apmEvents.addAll(generateNodeOperationDetailEvents(traceData, currentTime, metricsStateByKey));
+                generateNodeOperationDetailEvents(traceData, currentTime, sumStateByKey, histogramStateByKey, dedupedNodeDetails);
             }
         }
 
-        final List<JacksonMetric> metrics = ApmServiceMapMetricsUtil.createMetricsFromAggregatedState(metricsStateByKey);
+        // Convert deduped NodeOperationDetails to events
+        for (NodeOperationDetail detail : dedupedNodeDetails) {
+            final EventMetadata eventMetadata = new DefaultEventMetadata.Builder()
+                    .withEventType(EVENT_TYPE_OTEL_APM_SERVICE_MAP).build();
+
+            final Event event = eventFactory.eventBuilder(EventBuilder.class)
+                    .withEventMetadata(eventMetadata)
+                    .withData(detail)
+                    .build();
+
+            apmEvents.add(new Record<>(event));
+        }
+
+        final List<JacksonMetric> metrics = ApmServiceMapMetricsUtil.createMetricsFromAggregatedState(sumStateByKey, histogramStateByKey);
         metrics.sort(Comparator.comparing(JacksonMetric::getTime));
 
         final List<Record<Event>> apmEventsSorted = new ArrayList<>();
@@ -444,15 +498,23 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
     }
 
     /**
-     * Get anchor timestamp from span's endTime, truncated to minute boundary
+     * Get anchor timestamp for metrics, truncated to the specified unit.
+     * When metric_timestamp_source is ARRIVAL_TIME, uses fallbackTime (clock.instant()).
+     * When metric_timestamp_source is SPAN_END_TIME, uses the span's endTime field.
      *
      * @param spanStateData The span to extract timestamp from
-     * @param fallbackTime Current system time to use if span endTime is null
-     * @return Instant truncated to the lower 1-minute boundary
+     * @param fallbackTime Current system time to use as arrival time or if span endTime is null
+     * @param truncationUnit The ChronoUnit to truncate the timestamp to
+     * @return Instant truncated to the specified boundary
      */
-    private Instant getAnchorTimestampFromSpan(final SpanStateData spanStateData, final Instant fallbackTime) {
-        Instant timestamp = fallbackTime; // Default to current system time
+    private Instant getAnchorTimestampFromSpan(final SpanStateData spanStateData, final Instant fallbackTime,
+                                               final ChronoUnit truncationUnit) {
+        if (metricTimestampSource == MetricTimestampSource.ARRIVAL_TIME) {
+            return fallbackTime.truncatedTo(truncationUnit);
+        }
 
+        // SPAN_END_TIME mode: parse span's endTime, fall back to system time
+        Instant timestamp = fallbackTime;
         final String endTime = spanStateData.getEndTime();
         try {
             if (endTime != null && !endTime.isEmpty()) {
@@ -463,7 +525,23 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
                      endTime, e.getMessage());
         }
 
-        return timestamp.truncatedTo(java.time.temporal.ChronoUnit.MINUTES);
+        return timestamp.truncatedTo(truncationUnit);
+    }
+
+    /**
+     * Resolve a stable host identifier for this Data Prepper instance.
+     * Uses a truncated SHA-256 hash of the hostname (from {@link HostContext})
+     * to ensure uniqueness without revealing the actual hostname in emitted metrics.
+     */
+    private String resolveHostId() {
+        try {
+            final String hostname = HostContext.getHostname();
+            final MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            final byte[] hash = digest.digest(hostname.getBytes(StandardCharsets.UTF_8));
+            return Hex.encodeHexString(hash).substring(0, 16);
+        } catch (final java.security.NoSuchAlgorithmException e) {
+            throw new RuntimeException("SHA-256 algorithm not available", e);
+        }
     }
 
     /**
@@ -718,11 +796,11 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
      * @param metricsStateByKey Shared map for metric aggregation across all traces
      * @return Collection of NodeOperationDetail events
      */
-    private Collection<Record<Event>> generateNodeOperationDetailEvents(final ThreeWindowTraceDataWithDecorations traceData,
-                                                                        final Instant currentTime,
-                                                                        final Map<MetricKey, MetricAggregationState> metricsStateByKey) {
-        final Collection<Record<Event>> events = new HashSet<>();
-
+    private void generateNodeOperationDetailEvents(final ThreeWindowTraceDataWithDecorations traceData,
+                                                    final Instant currentTime,
+                                                    final Map<MetricKey, MetricAggregationState> sumStateByKey,
+                                                    final Map<MetricKey, MetricAggregationState> histogramStateByKey,
+                                                    final Set<NodeOperationDetail> dedupedNodeDetails) {
         // Step 1: CLIENT spans — primary emission path
         for (SpanStateData clientSpan : traceData.getProcessingSpans()) {
             if (SPAN_KIND_CLIENT.equals(clientSpan.getSpanKind())) {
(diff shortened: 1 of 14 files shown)


data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -71,6 +73,7 @@ public class OTelApmServiceMapProcessor extends AbstractProcessor<Record<Event>,
 
     private static final String SPANS_DB_SIZE = "spansDbSize";
     private static final String SPANS_DB_COUNT = "spansDbCount";
+    private static final String HOST_ID = resolveHostId();
```
dlvenable: This should probably be non-static and performed during the construction of the processor.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtil.java
```diff
@@ -45,134 +48,140 @@ public final class ApmServiceMapMetricsUtil {
...
+                                                   final Instant sumAnchorTimestamp,
+                                                   final Instant histogramAnchorTimestamp,
+                                                   final String hostId) {
+        // Build metric labels using span's groupByAttributes
         final Map<String, Object> labels = new HashMap<>();
         labels.put("namespace", "span_derived");
         labels.put("environment", serverSpan.getEnvironment());
         labels.put("service", serverSpan.getServiceName());
         labels.put("operation", serverSpan.getOperationName());
+        labels.put("trace_processor_host_id", hostId);
```
dlvenable: We call this processor "service map", not "trace processor." Is this an existing label name? Might this name cause confusion?

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtilTest.java
```diff
@@ -45,25 +45,30 @@
 @ExtendWith(MockitoExtension.class)
 class ApmServiceMapMetricsUtilTest {
 
+    private static final String TEST_HOST_ID = "test-host-1";
```
dlvenable: Make this a non-static member. Generate it to a random value in `@BeforeEach`.

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtilTest.java
```diff
@@ -575,62 +597,86 @@ void testMetricsLabelsCorrectness_ClientSpan() {
...
+                sumStateByKey, histogramStateByKey);
 
         // Then
         assertFalse(metrics.isEmpty());
-        // Verify metrics are sorted by timestamp - compare the first few metrics
         if (metrics.size() >= 2) {
             String firstTimestamp = metrics.get(0).getTime();
             String secondTimestamp = metrics.get(1).getTime();
-            assertTrue(firstTimestamp.compareTo(secondTimestamp) <= 0, 
+            assertTrue(firstTimestamp.compareTo(secondTimestamp) <= 0,
```
dlvenable: Use Hamcrest `lessThanOrEqualTo`:

```
assertThat("Metrics should be sorted by timestamp"
  firstTimestamp, lessThanOrEqualTo(secondTimestamp));
```

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtil.java
```diff
@@ -45,134 +48,140 @@ public final class ApmServiceMapMetricsUtil {
...
-        if (durationInNanos != null && durationInNanos > 0) {
-            final double durationInSeconds = durationInNanos / 1_000_000_000.0;
-            state.addLatencyDuration(durationInSeconds);
+        labels.put("environment", clientSpan.getEnvironment());
+        labels.put("service", clientSpan.getServiceName());
+        labels.put("operation", decoration.getParentServerOperationName());
+        labels.put("remoteEnvironment", decoration.getRemoteEnvironment());
+        labels.put("remoteService", decoration.getRemoteService());
+        labels.put("remoteOperation", decoration.getRemoteOperation());
+        labels.put("trace_processor_host_id", hostId);
```
dlvenable: This should be a private static final. I see it used three times in this file.

Also, maybe we should have small method named `putUniqueLabels()` that this and line 133 use. It is small, but this will help us keep a consistent view if things change over time.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -463,7 +471,25 @@ private Instant getAnchorTimestampFromSpan(final SpanStateData spanStateData, fi
...
+        return timestamp.truncatedTo(truncationUnit);
+    }
+
+    /**
+     * Resolve a stable host identifier for this Data Prepper instance.
+     * Uses a truncated SHA-256 hash of the hostname to ensure uniqueness
+     * without revealing the actual hostname in emitted metrics.
+     * Falls back to a random UUID if hostname resolution fails.
+     */
+    private static String resolveHostId() {
```
dlvenable: I wonder if we should perform this in Data Prepper core and expose it somehow in `data-prepper-api`. We get the hostname (or maybe IP address) in Data Prepper core for peer-forwarding. Maybe we want to consistency here. I'm not sure on the effort to value yet though.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -463,7 +471,25 @@ private Instant getAnchorTimestampFromSpan(final SpanStateData spanStateData, fi
...
+        try {
+            final String hostname = InetAddress.getLocalHost().getHostName();
+            final java.security.MessageDigest digest = java.security.MessageDigest.getInstance("SHA-256");
+            final byte[] hash = digest.digest(hostname.getBytes(java.nio.charset.StandardCharsets.UTF_8));
+            return Hex.encodeHexString(hash).substring(0, 16);
+        } catch (Exception e) {
+            LOG.warn("Failed to resolve hostname for trace_processor_host_id, using random UUID: {}", e.getMessage());
```
dlvenable: Can we have a better fallback here?

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -463,7 +471,25 @@ private Instant getAnchorTimestampFromSpan(final SpanStateData spanStateData, fi
...
+    /**
+     * Resolve a stable host identifier for this Data Prepper instance.
+     * Uses a truncated SHA-256 hash of the hostname to ensure uniqueness
+     * without revealing the actual hostname in emitted metrics.
+     * Falls back to a random UUID if hostname resolution fails.
+     */
+    private static String resolveHostId() {
+        try {
+            final String hostname = InetAddress.getLocalHost().getHostName();
+            final java.security.MessageDigest digest = java.security.MessageDigest.getInstance("SHA-256");
```
dlvenable: Use imports instead of fully qualified names. There are a few places where this needs to be corrected.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
vamsimanohar: Done. Changed `HOST_ID` from a static field to a non-static `hostId` instance field, resolved during construction via `resolveHostId()`.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtil.java
vamsimanohar: Good catch — renamed from `trace_processor_host_id` to `service_map_processor_host_id` to match the processor name.

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtilTest.java
vamsimanohar: Done. Changed `TEST_HOST_ID` to a non-static `testHostId` field, generated as a random UUID in `@BeforeEach`.

data-prepper-plugins/otel-apm-service-map-processor/src/test/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtilTest.java
vamsimanohar: Done. Switched to Hamcrest `lessThanOrEqualTo` for timestamp comparison.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/utils/ApmServiceMapMetricsUtil.java
vamsimanohar: Done. Extracted `HOST_ID_LABEL` as a `private static final` constant, and added a `putCommonLabels()` helper method used by both `generateMetricsForClientSpan` and `generateMetricsForServerSpan`.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
vamsimanohar: Done in this PR. Created a shared `HostContext` utility class in `data-prepper-api` (`org.opensearch.dataprepper.model.host.HostContext`) that uses `InetAddress.getLocalHost().getHostName()` — the same pattern used by `LeaseBasedSourceCoordinator` and `EnhancedLeaseBasedSourceCoordinator` in core. The processor now calls `HostContext.getHostname()` instead of resolving inline. This gives us a single shared hostname source that other components can also use.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
vamsimanohar: Done. The hostname resolution is now handled by `HostContext` in `data-prepper-api`, which follows the same pattern as `LeaseBasedSourceCoordinator` — using `InetAddress.getLocalHost().getHostName()` with a clear `IllegalStateException` if resolution fails. The processor hashes the hostname via SHA-256 (truncated to 16 hex chars) for the label value.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
vamsimanohar: Done. Replaced all fully qualified names with proper imports.

data-prepper-plugins/otel-apm-service-map-processor/src/main/java/org/opensearch/dataprepper/plugins/processor/otel_apm_service_map/OTelApmServiceMapProcessor.java
```diff
@@ -746,24 +824,18 @@ private Collection<Record<Event>> generateNodeOperationDetailEvents(final ThreeW
...
-                            .build();
-
-                    events.add(new Record<>(event));
+                    dedupedNodeDetails.add(nodeOperationDetail);
 
                     if (decoration.getParentServerOperationName() != null) {
                         ApmServiceMapMetricsUtil.generateMetricsForClientSpan(
-                                clientSpan, decoration, currentTime, metricsStateByKey, anchorTimestamp);
+                                clientSpan, decoration, currentTime, sumStateByKey, histogramStateByKey,
+                                anchor, anchor, hostId);
```
dlvenable: You added a different anchor for sum and histogram, but you only pass in one. Is this correct? Should you have different ones with different units?

vamsimanohar: Good catch. The two separate parameters (`sumAnchorTimestamp` and `histogramAnchorTimestamp`) were leftover from an earlier design where sum metrics used seconds truncation and histogram metrics used minutes truncation. After adding the configurable `metric_timestamp_granularity`, both use the same granularity, making the two parameters redundant.

Fixed by unifying into a single `anchorTimestamp` parameter in both `generateMetricsForClientSpan` and `generateMetricsForServerSpan`.

Additionally, separated service map and metric timestamp handling:
- **Metrics**: use the configured `metric_timestamp_source` + `metric_timestamp_granularity` (the `anchorTimestamp` parameter)
- **Service map (NodeOperationDetail)**: always use span end time truncated to minutes — this reflects actual interaction time and reduces OpenSearch document count, since dashboards query service maps with `dedup nodeConnectionHash` over time ranges

dlvenable (approved):
Thank you @vamsimanohar !

PR #2692, opened 2023-05-13
Fix float point number grammar in DataPrepperExpression

kkondaka:
### Description
Fixes float point number  grammar in DataPrepperExpression
The new grammar also supports exponent notation of float point numbers.

Resolves #2691 
 
### Issues Resolved
#2691 
 
### Check List
- [ X] New functionality includes testing.
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [X ] Commits are signed with a real name per the DCO

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/data-prepper/blob/main/CONTRIBUTING.md).

PR #6673, opened 2026-03-25
Fixes the typeof operator along with data types in expressions.

dlvenable:
### Description

The function composition change (#6628) converted function from a single lexer rule (`Function`) that matched the entire `name(args)` pattern into a parser rule that referenced a new standalone lexer rule `FunctionName`. This `FunctionName` rule was defined as `JsonPointerCharacters`, which matches `[A-Za-z0-9_.@]+`.

This caused the `typeof` operator to stop working. This is because ANTLR was resolving our data type keywords (e.g. `integer`, `map`) as `FunctionName`. ANTLR did this because `FunctionName` came first in the grammar and is very broad. ANTLR resolves conflicts by choosing the rules defined first the grammar. So ANTLR was trying to resolve the data types as functions.

The fix is to move the `FunctionName` below the `DataTypes`. Additionally, since `FunctionName` is so broad, I renamed it to `Identifier`. This means it can be a catch all for non-keywords that may be expanded in the expression language. This name should be clearer and allow it to be re-used so we don't hit this issue again with other identifiers.

Also there were no tests for the `typeof` operator. I have added some of these tests, first to verify the failure and fix, and second to ensure future changes do not break them.

This PR also adds additional test cases to ensure that existing features are working as expected. It corrects incorrect assertions related to tokens from #2692 that was attempting to verify that invalid floating point values are not parsed. It also adds new tests on the grammar for functions.
 
### Issues Resolved

N/A
 
### Check List

### DIFF
--- a/data-prepper-expression/src/main/antlr/DataPrepperExpression.g4
+++ b/data-prepper-expression/src/main/antlr/DataPrepperExpression.g4
@@ -155,7 +155,7 @@ jsonPointer
     ;
 
 function
-    : FunctionName LPAREN functionArgs? RPAREN
+    : Identifier LPAREN functionArgs? RPAREN
     ;
 
 functionArgs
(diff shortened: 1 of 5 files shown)


dlvenable:
Looking at the example Java grammar for ANTLR, they clearly call out that the the `identifer` must show up after all keywords.

https://github.com/antlr/grammars-v4/blob/8c7185f34a6958da71c6faa2e07abe9710f3e236/java/java20/Java20Lexer.g4#L273-L275

PR #6687, opened 2026-03-30
Fix NPE in sanitizeMetricName when unit or aggregationTemporality is …

### LINKED ISSUE #6683: Prometheus sink NPE in sanitizeMetricName when unit or aggregationTemporality is null.
**Describe the bug**
The prometheus sink throws NullPointerException in PrometheusTimeSeries.sanitizeMetricName() when processing metric events where the unit field or the aggregationTemporality field is null. The method calls .equals() on getAggregationTemporality() and .startsWith() on unit without null checks. These fields are optional in the OTEL metric model and are null when metrics come from sources that do not carry this metadata, such as Prometheus Remote Write v1.

**To Reproduce**
 1. Configure a Data Prepper pipeline with the prometheus source receiving Remote Write v1 data
 2. Configure the prometheus sink as the output of that pipeline
 3. Start a Prometheus instance with remote_write pointing to the Data Prepper source
 4. Observe repeated NullPointerException warnings in Data Prepper logs for every metric event

**Expected behavior**
The prometheus sink should handle null unit and null aggregationTemporality gracefully instead of throwing NullPointerException. A simple nullcheck before calling .equals() and .startsWith() on these fields would resolve both issues.

**Exceptions**
Stack trace for aggregationTemporality:
  NullPointerException: Cannot invoke "String.equals(Object)" because the return value of
  "org.opensearch.dataprepper.model.metric.Sum.getAggregationTemporality()" is null
      at PrometheusTimeSeries.sanitizeMetricName(PrometheusTimeSeries.java:305)
      at PrometheusTimeSeries.<init>(PrometheusTimeSeries.java:96)
      at PrometheusSinkBufferEntry.getTimeSeriesForEvent(PrometheusSinkBufferEntry.java:49)

  Stack trace for unit:
  NullPointerException: Cannot invoke "String.startsWith(String)" because "unit" is null
      at PrometheusTimeSeries.sanitizeMetricName(PrometheusTimeSeries.java:310)
      at PrometheusTimeSeries.<init>(PrometheusTimeSeries.java:96)
      at PrometheusSinkBufferEntry.getTimeSeriesForEvent(PrometheusSinkBufferEntry.java:49)

- OS: macOS (Darwin 25.3.0)
- Version: Data Prepper 2.15.0-SNAPSHOT (main branch)

**Additional context**
Found during end-to-end testing of the prometheus source plugin (PR #6627). The prometheus source produces valid OTEL metric events but Prometheus Remote Write v1 protocol does not carry unit or aggregation temporality metadata, so these fields are null. 

The sink was built assuming these fields are always populated, which is only true for metrics coming from the OTel Collector. Current workaround is to set sanitize_names: false in the prometheus sink configuration to bypass the sanitizeMetricName() method entirely.

srikanthpadakanti:
### Description
  Fix NPE in `PrometheusTimeSeries.sanitizeMetricName()` when `unit` or `aggregationTemporality` is null.
                                                                                                                                                    
  ### Issues Resolved
  Resolves #6683                                                                                                                                    
                                                            
  ### Check List

### DIFF
--- a/data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusTimeSeries.java
+++ b/data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusTimeSeries.java
@@ -307,35 +307,37 @@ public class PrometheusTimeSeries {
         StringBuilder metricNameBuilder = new StringBuilder(sanitizeName(name, true, false));
         String suffix = isCounter ? TOTAL_SUFFIX : "";
 
-        if (unit != null && unit.startsWith("{")) {
+        if (unit == null) {
             return metricNameBuilder.append(suffix).toString();
         }
 
-        if (unit != null) {
-            if ("1".equals(unit) && isGauge) {
-                return metricNameBuilder.append(RATIO_SUFFIX).toString();
-            }
+        if (unit.startsWith("{")) {
+            return metricNameBuilder.append(suffix).toString();
+        }
 
-            String mappedUnit = otelToPrometheusUnitsMap.get(unit);
-            if (mappedUnit != null) {
-                return metricNameBuilder.append(UNDERSCORE).append(mappedUnit).append(suffix).toString();
-            }
+        if ("1".equals(unit) && isGauge) {
+            return metricNameBuilder.append(RATIO_SUFFIX).toString();
+        }
 
-            if (unit.contains("/")) {
-                String[] unitSplit = unit.split("/", 2);
-                if (unitSplit.length == 2) {
-                    String unit1 = otelToPrometheusUnitsMap.get(unitSplit[0]);
-                    String unit2 = otelToPrometheusUnitsMap.get(unitSplit[1]);
-                    if (unit1 != null && unit2 != null) {
-                        return metricNameBuilder.append(UNDERSCORE).append(unit1)
-                                .append(UNDERSCORE).append(unit2).append(suffix).toString();
-                    }
+        String mappedUnit = otelToPrometheusUnitsMap.get(unit);
+        if (mappedUnit != null) {
+            return metricNameBuilder.append(UNDERSCORE).append(mappedUnit).append(suffix).toString();
+        }
+
+        if (unit.contains("/")) {
+            String[] unitSplit = unit.split("/", 2);
+            if (unitSplit.length == 2) {
+                String unit1 = otelToPrometheusUnitsMap.get(unitSplit[0]);
+                String unit2 = otelToPrometheusUnitsMap.get(unitSplit[1]);
+                if (unit1 != null && unit2 != null) {
+                    return metricNameBuilder.append(UNDERSCORE).append(unit1)
+                            .append(UNDERSCORE).append(unit2).append(suffix).toString();
                 }
             }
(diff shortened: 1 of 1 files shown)


data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusTimeSeries.java
```diff
@@ -302,38 +302,40 @@ static String sanitizeMetricName(final Metric metric) {
         final boolean isGauge = metric.getKind().equals(Metric.KIND.GAUGE.toString());
         final boolean isCounter = metric.getKind().equals(Metric.KIND.SUM.toString()) &&
                 ((Sum) metric).isMonotonic() &&
-                ((Sum) metric).getAggregationTemporality().equals("AGGREGATION_TEMPORALITY_CUMULATIVE");
+                "AGGREGATION_TEMPORALITY_CUMULATIVE".equals(((Sum) metric).getAggregationTemporality());
```
dlvenable, line 305: Let's include a test case for when this is true and the unit is `null`.

```
@Test
public void testSumMetricCounterWithNullUnit() {
    final String name = RandomStringUtils.randomAlphabetic(10);
    Sum sum = createSumMetric(name, null, true, "AGGREGATION_TEMPORALITY_CUMULATIVE");
    String sanitizedName = PrometheusTimeSeries.sanitizeMetricName(sum);
    assertThat(sanitizedName, equalTo(name + "_total"));
}
```

srikanthpadakanti: Sure.

data-prepper-plugins/prometheus-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/prometheus/service/PrometheusTimeSeries.java
```diff
@@ -302,38 +302,40 @@ static String sanitizeMetricName(final Metric metric) {
...
 
-        if (unit.startsWith("{")) {
+        if (unit != null && unit.startsWith("{")) {
             return metricNameBuilder.append(suffix).toString();
         }
 
-        if ("1".equals(unit) && isGauge) {
-            return metricNameBuilder.append(RATIO_SUFFIX).toString();
-        }
+        if (unit != null) {
```
kkondaka: I think the right fix is to check for `unit` being null at the very beginning. Almost everything depends on unit

dlvenable: What should be the behavior when `unit` is `null`? Do we not sanitize the name? Do we drop the event to DLQ?

srikanthpadakanti: @kkondaka  sure. 

@dlvenable When unit is null we return the sanitized metric name with the counter suffix (if applicable), skipping all unit-based suffix logic. This is pretty muchsame  as the unit.startsWith("{") case.

PR #6691, opened 2026-03-31
Add index_type tsdb support to OpenSearch sink for Prometheus metrics

### LINKED ISSUE #6644: Feature Request: OpenSearch Sink index_type: tsdb for TSDB Index Support
# Feature Request: OpenSearch Sink `index_type: tsdb` for TSDB Index Support

## Is your feature request related to a problem?

With the Prometheus Remote Write source (#6533, PR #6627), Data Prepper will be able to ingest Prometheus metrics as Metric events. But there's no way to write these to OpenSearch TSDB indices. The existing `index_type` options like `metric-analytics` produce OTel-format documents. TSDB expects a completely different structure: `{labels, timestamp, value}` with space-separated label strings.

Anyone wanting to use the [OpenSearch TSDB plugin](https://github.com/opensearch-project/time-series-db) for metrics storage and M3QL queries has no path through Data Prepper today.

## Proposed solution

A new `index_type: tsdb` in the OpenSearch sink. The sink converts Metric events to TSDB document format directly and no processor in between.

```yaml
prometheus-pipeline:
  source:
    prometheus:
      port: 9090
      path: "/api/v1/write"
  sink:
    - opensearch:
        hosts: ["https://opensearch:9200"]
        index_type: tsdb
        index: metrics
```

This keeps the pipeline simple and also lets us do multi-sink --> send to both Prometheus and TSDB in one pipeline:

```yaml
prometheus-pipeline:
  source:
    prometheus:
      port: 9090
      path: "/api/v1/write"
  sink:
    - prometheus:
        url: "https://aps-workspaces.us-west-2.amazonaws.com/workspaces/ws-xxx"
    - opensearch:
        hosts: ["https://opensearch:9200"]
        index_type: tsdb
        index: metrics
```

### TSDB Document Format

TSDB expects three fields per document:

```json
{
  "labels": "__name__ http_requests_total method POST handler /api/items status 200",
  "timestamp": 1633072800000,
  "value": 1.1
}
```

- `labels` — space-separated key-value pairs, sorted by key. `__name__` is the metric name.
- `timestamp` — epoch milliseconds.
- `value` — single numeric value. One value per document.
- Spec: https://github.com/opensearch-project/time-series-db

### How metric types map to TSDB documents

The sink takes standard Data Prepper Metric events (JacksonGauge, JacksonSum, JacksonHistogram, JacksonSummary) and converts them. TSDB stores one value per document, so complex types need expansion.

**Gauge --> 1 document:**
```
Input:  Gauge{name="cpu_temp", value=72.5, attributes={host: "server-01"}}
Output: {"labels":"__name__ cpu_temp host server-01","timestamp":1706869800000,"value":72.5}
```

**Sum (Counter) --> 1 document:**
For monotonic sums, the sink appends `_total` to the metric name if not already present (Prometheus convention for counters in TSDB).
```
Input:  Sum{name="http_requests", value=42100, isMonotonic=true, attributes={method: "GET"}}
Output: {"labels":"__name__ http_requests_total method GET","timestamp":1706869800000,"value":42100.0}
```

**Histogram --> N+2 documents:**
The sink treats `bucketCountsList` as per-bucket counts and converts to cumulative — which is what TSDB and M3QL expect (Prometheus convention). Each bucket becomes a separate document with a `le` label, plus `_count` and `_sum` documents.
```
Input:  Histogram{name="request_duration", bucketCountsList=[5, 7, 13, 5],
        explicitBounds=[0.1, 0.5, 1.0], count=30, sum=15.5, attributes={method: "GET"}}

Output (6 documents):
  {"labels":"__name__ request_duration_bucket le 0.1 method GET","timestamp":...,"value":5.0}
  {"labels":"__name__ request_duration_bucket le 0.5 method GET","timestamp":...,"value":12.0}
  {"labels":"__name__ request_duration_bucket le 1.0 method GET","timestamp":...,"value":25.0}
  {"labels":"__name__ request_duration_bucket le +Inf method GET","timestamp":...,"value":30.0}
  {"labels":"__name__ request_duration_count method GET","timestamp":...,"value":30.0}
  {"labels":"__name__ request_duration_sum method GET","timestamp":...,"value":15.5}
```

**Summary --> N+2 documents:**
Each quantile becomes a separate document with a `quantile` label, plus `_count` and `_sum` documents.
```
Input:  Summary{name="rpc_latency", quantiles=[{0.5: 0.2}, {0.99: 0.8}],
        count=1000, sum=300.5, attributes={service: "api"}}

Output (4 documents):
  {"labels":"__name__ rpc_latency quantile 0.5 service api","timestamp":...,"value":0.2}
  {"labels":"__name__ rpc_latency quantile 0.99 service api","timestamp":...,"value":0.8}
  {"labels":"__name__ rpc_latency_count service api","timestamp":...,"value":1000.0}
  {"labels":"__name__ rpc_latency_sum service api","timestamp":...,"value":300.5}
```

### Implementation approach

Following the same pattern as `metric-analytics`, `log-analytics`, `trace-analytics-raw`:

1. **`IndexType.java`** --> add `TSDB("tsdb")` enum value
2. **`IndexConstants.java`** --> template filename + default alias `metrics-tsdb-v1`
3. **`IndexConfiguration.java`** --> TSDB branch in `readIndexTemplate()`
4. **`IndexManagerFactory.java`** --> `case TSDB:` with `NoIsmPolicyManagement` since TSDB manages its own lifecycle (head chunks → blocks, no ISM rollover needed)
5. **`OpenSearchSink.java`** --> TSDB branch in `doOutput()` using a new `TSDBDocumentBuilder`
6. **`TSDBDocumentBuilder.java`** --> new class: Metric → TSDB document conversion with histogram/summary expansion
7. **`tsdb-index-template.json`** --> index template with TSDB mappings

### Index template

```json
{
  "version": 1,
  "mappings": {
    "properties": {
      "series_ref": {"type": "long", "doc_values": false},
      "labels": {"type": "keyword", "ignore_above": 4096},
      "value": {"type": "double", "doc_values": false},
      "timestamp": {"type": "date", "format": "epoch_millis"},
      "timestamp_range": {"type": "long_range"}
    }
  },
  "settings": {
    "index.translog.durability": "async",
    "index.translog.sync_interval": "1s",
    "refresh_interval": "1s"
  }
}
```

Notes on deviations from the [TSDB README example](https://github.com/opensearch-project/time-series-db#create-a-tsdb-engine-index):
- `value` uses `double` instead of `float` --> Prometheus uses float64 internally, and OpenSearch `float` (32-bit) would silently lose precision on large counter values.
- `labels` has `ignore_above: 4096` --> guards against indexing failures if a metric has many labels. Typical Prometheus label strings are ~200 chars.
- TSDB-engine settings (`tsdb_engine.enabled`, `tsdb_store`, `labels.storage_type`) are intentionally not in the template and those are cluster-level settings the admin configures when installing the [TSDB plugin](https://github.com/opensearch-project/time-series-db). Keeping only standard OpenSearch mappings means the template works on any OpenSearch version.

### Things worth noting

- **No processor needed** --> histogram/summary expansion happens at the sink level in `doOutput()`, so the pipeline stays simple (source --> sink) and multi-sink works.
- **No routing** --> TSDB's `TSDBAutoRoutingActionFilter` handles shard placement from labels automatically. The sink sends documents without routing.
- **No ISM** --> TSDB manages lifecycle through head chunks and blocks. ISM rollover would conflict with that.
- **Label sanitization** --> spaces in label values get replaced with underscores since space is TSDB's delimiter.
- **Histogram bucket assumption** --> the sink treats `bucketCountsList` as per-bucket (delta) counts and converts to cumulative via running sum. This matches what the prometheus source (#6533) produces. If Metric events from other sources already have cumulative bucket counts, the conversion would need to be aware of that.

### Compatibility

- The TSDB plugin (built against OpenSearch `3.5.0-SNAPSHOT` per its [build.gradle](https://github.com/opensearch-project/time-series-db/blob/main/build.gradle)) requires OpenSearch 3.5.0+. But our index template uses standard mappings only, so it works on any OpenSearch version. TSDB features become available when the plugin is installed.
- All changes would be behind `indexType == IndexType.TSDB`. Existing index types are not affected.
- The source (#6533) emits Data Prepper Metric events (JacksonGauge, JacksonSum, JacksonHistogram, JacksonSummary). The TSDB sink converts these to TSDB's `{labels, timestamp, value}` format and no processor needed between source and sink.

## Related content

- Prometheus Remote Write source: #6533
- Source PR: #6627
- TSDB plugin: https://github.com/opensearch-project/time-series-db
- TSDB label format: https://github.com/opensearch-project/time-series-db#index-some-metrics
- Prometheus Remote Write spec: https://prometheus.io/docs/specs/remote_write_spec/

srikanthpadakanti:
### Description
Add `index_type: tsdb` support to the OpenSearch sink for writing Prometheus metrics to OpenSearch TSDB indices. The sink converts Data Prepper Metric events (Gauge, Sum, Histogram, Summary) into TSDB's `{labels, timestamp, value}` document format, expanding complex types (histograms into per-bucket documents, summaries into per-quantile documents) so no intermediate processor is needed.
 
### Issues Resolved
https://github.com/opensearch-project/data-prepper/issues/6644
Resolves #6644 
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSinkTsdbIT.java
+++ b/data-prepper-plugins/opensearch/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSinkTsdbIT.java
@@ -1,6 +1,10 @@
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
  */
 
 package org.opensearch.dataprepper.plugins.sink.opensearch;
--- a/data-prepper-plugins/opensearch/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSinkTsdbIT.java
+++ b/data-prepper-plugins/opensearch/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSinkTsdbIT.java
@@ -153,7 +153,6 @@ class OpenSearchSinkTsdbIT {
         }
     }
 
-    // --- Index Initialization ---
 
     @Test
     @DisabledIf(value = "isES6", disabledReason = "TSDB is not supported for ES 6")
@@ -169,7 +168,6 @@ class OpenSearchSinkTsdbIT {
         final Response response = client.performRequest(request);
         assertThat(response.getStatusLine().getStatusCode(), equalTo(SC_OK));
 
-        // Query mappings via alias (TSDB uses NoIsmPolicyManagement so index name may not follow -000001 pattern)
         final String extraURI = DeclaredOpenSearchVersion.OPENDISTRO_0_10.compareTo(
                 OpenSearchIntegrationHelper.getVersion()) >= 0 ? INCLUDE_TYPE_NAME_FALSE_URI : "";
         final Request mappingRequest = new Request(HttpMethod.GET, indexAlias + "/_mappings" + extraURI);
@@ -182,7 +180,6 @@ class OpenSearchSinkTsdbIT {
         final Map<String, Object> mappings = (Map<String, Object>) ((Map<String, Object>) mappingResult.get(actualIndex)).get("mappings");
         assertThat(mappings, notNullValue());
 
-        // Verify TSDB-specific mapping fields
         @SuppressWarnings("unchecked")
         final Map<String, Object> properties = (Map<String, Object>) mappings.get("properties");
         assertThat(properties, notNullValue());
@@ -190,11 +187,8 @@ class OpenSearchSinkTsdbIT {
         assertThat(properties.containsKey("timestamp"), equalTo(true));
         assertThat(properties.containsKey("value"), equalTo(true));
 
-        // TSDB uses NoIsmPolicyManagement — no ISM policy should be attached
-        // (unlike metric-analytics which has ISM)
     }
 
-    // --- Gauge Output ---
 
     @Test
     @DisabledIf(value = "isES6", disabledReason = "TSDB is not supported for ES 6")
@@ -222,7 +216,6 @@ class OpenSearchSinkTsdbIT {
         assertThat(((Number) doc.get("value")).doubleValue(), closeTo(72.5, 0.001));
         assertThat(doc.get("timestamp"), notNullValue());
 
-        // Verify metrics
         final List<Measurement> bulkRequestErrors = MetricsTestUtil.getMeasurementList(
                 new StringJoiner(MetricNames.DELIMITER).add(PIPELINE_NAME).add(PLUGIN_NAME)
                         .add(OpenSearchSink.BULKREQUEST_ERRORS).toString());
@@ -230,7 +223,6 @@ class OpenSearchSinkTsdbIT {
         Assert.assertEquals(0.0, bulkRequestErrors.get(0).getValue(), 0);
     }
 
-    // --- Sum (Counter) Output ---
 
     @Test
     @DisabledIf(value = "isES6", disabledReason = "TSDB is not supported for ES 6")
@@ -255,11 +247,9 @@ class OpenSearchSinkTsdbIT {
         assertThat(sources, hasSize(1));
 
         final Map<String, Object> doc = sources.get(0);
-        // Monotonic sum should have _total suffix
         assertThat(doc.get("labels"), equalTo("__name__ http_requests_total method GET"));
         assertThat(((Number) doc.get("value")).doubleValue(), closeTo(100.0, 0.001));
 
-        // Verify metrics
         final List<Measurement> bulkRequestErrors = MetricsTestUtil.getMeasurementList(
                 new StringJoiner(MetricNames.DELIMITER).add(PIPELINE_NAME).add(PLUGIN_NAME)
                         .add(OpenSearchSink.BULKREQUEST_ERRORS).toString());
@@ -267,7 +257,6 @@ class OpenSearchSinkTsdbIT {
         Assert.assertEquals(0.0, bulkRequestErrors.get(0).getValue(), 0);
     }
 
-    // --- Histogram Expansion ---
 
     @Test
     @DisabledIf(value = "isES6", disabledReason = "TSDB is not supported for ES 6")
@@ -291,20 +280,16 @@ class OpenSearchSinkTsdbIT {
 
         final String indexAlias = IndexConstants.TYPE_TO_DEFAULT_ALIAS.get(IndexType.TSDB);
         final List<Map<String, Object>> sources = getSearchResponseDocSources(indexAlias);
-        // 4 bucket docs + 1 _count + 1 _sum = 6
         assertThat(sources, hasSize(6));
 
-        // Collect labels for verification
         final List<String> labels = sources.stream()
                 .map(s -> (String) s.get("labels"))
                 .collect(Collectors.toList());
 
-        // Verify bucket docs exist with cumulative counts
         assertThat(labels.stream().filter(l -> l.contains("request_duration_bucket")).count(), equalTo(4L));
         assertThat(labels.stream().filter(l -> l.contains("request_duration_count")).count(), equalTo(1L));
         assertThat(labels.stream().filter(l -> l.contains("request_duration_sum")).count(), equalTo(1L));
 
-        // Verify cumulative bucket values
         final Map<String, Double> labelToValue = sources.stream()
                 .collect(Collectors.toMap(s -> (String) s.get("labels"), s -> ((Number) s.get("value")).doubleValue()));
 
@@ -315,14 +300,12 @@ class OpenSearchSinkTsdbIT {
         assertThat(labelToValue.get("__name__ request_duration_count method GET"), closeTo(20.0, 0.001));
         assertThat(labelToValue.get("__name__ request_duration_sum method GET"), closeTo(5.5, 0.001));
 
-        // Verify document success count matches expanded document count
         final List<Measurement> documentsSuccess = MetricsTestUtil.getMeasurementList(
                 new StringJoiner(MetricNames.DELIMITER).add(PIPELINE_NAME).add(PLUGIN_NAME)
                         .add(BulkRetryStrategy.DOCUMENTS_SUCCESS).toString());
         assertThat(documentsSuccess.size(), equalTo(1));
         assertThat(documentsSuccess.get(0).getValue(), closeTo(6.0, 0));
 
-        // Verify no errors
         final List<Measurement> bulkRequestErrors = MetricsTestUtil.getMeasurementList(
                 new StringJoiner(MetricNames.DELIMITER).add(PIPELINE_NAME).add(PLUGIN_NAME)
                         .add(OpenSearchSink.BULKREQUEST_ERRORS).toString());
@@ -330,7 +313,6 @@ class OpenSearchSinkTsdbIT {
         Assert.assertEquals(0.0, bulkRequestErrors.get(0).getValue(), 0);
     }
 
-    // --- Summary Expansion ---
 
     @Test
     @DisabledIf(value = "isES6", disabledReason = "TSDB is not supported for ES 6")
@@ -358,7 +340,6 @@ class OpenSearchSinkTsdbIT {
 
         final String indexAlias = IndexConstants.TYPE_TO_DEFAULT_ALIAS.get(IndexType.TSDB);
         final List<Map<String, Object>> sources = getSearchResponseDocSources(indexAlias);
-        // 2 quantile docs + 1 _count + 1 _sum = 4
         assertThat(sources, hasSize(4));
 
         final Map<String, Double> labelToValue = sources.stream()
@@ -369,7 +350,6 @@ class OpenSearchSinkTsdbIT {
         assertThat(labelToValue.get("__name__ rpc_latency_count service api"), closeTo(1000.0, 0.001));
         assertThat(labelToValue.get("__name__ rpc_latency_sum service api"), closeTo(300.5, 0.001));
 
-        // Verify metrics
         final List<Measurement> bulkRequestErrors = MetricsTestUtil.getMeasurementList(
                 new StringJoiner(MetricNames.DELIMITER).add(PIPELINE_NAME).add(PLUGIN_NAME)
                         .add(OpenSearchSink.BULKREQUEST_ERRORS).toString());
@@ -377,7 +357,6 @@ class OpenSearchSinkTsdbIT {
         Assert.assertEquals(0.0, bulkRequestErrors.get(0).getValue(), 0);
     }
 
-    // --- Multiple Metrics in Single Batch ---
 
     @Test
     @DisabledIf(value = "isES6", disabledReason = "TSDB is not supported for ES 6")
@@ -408,10 +387,8 @@ class OpenSearchSinkTsdbIT {
 
         final String indexAlias = IndexConstants.TYPE_TO_DEFAULT_ALIAS.get(IndexType.TSDB);
         final List<Map<String, Object>> sources = getSearchResponseDocSources(indexAlias);
-        // 1 gauge + 1 sum = 2 documents
         assertThat(sources, hasSize(2));
 
-        // Verify metrics
         final List<Measurement> bulkRequestErrors = MetricsTestUtil.getMeasurementList(
                 new StringJoiner(MetricNames.DELIMITER).add(PIPELINE_NAME).add(PLUGIN_NAME)
                         .add(OpenSearchSink.BULKREQUEST_ERRORS).toString());
@@ -419,7 +396,6 @@ class OpenSearchSinkTsdbIT {
         Assert.assertEquals(0.0, bulkRequestErrors.get(0).getValue(), 0);
     }
 
-    // --- Re-instantiation (no duplicate index) ---
 
     @Test
     @DisabledIf(value = "isES6", disabledReason = "TSDB is not supported for ES 6")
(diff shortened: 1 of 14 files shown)


srikanthpadakanti:
@dlvenable @kkondaka Could you please trigger the remaining workflows and review this?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -219,6 +223,7 @@ public OpenSearchSink(final PluginSetting pluginSetting,
     this.lastFlushTimeMap = new ConcurrentHashMap<>();
     this.pluginConfigObservable = pluginConfigObservable;
     this.objectMapper = new ObjectMapper();
+    this.tsdbDocumentBuilder = (this.indexType == IndexType.TSDB) ? new TSDBDocumentBuilder() : null;
```
dlvenable: We should add an interface like `DocumentBuilder` or `CustomDocumentBuilder`. This way we can use the same pattern for any future changes.

Maybe it would work like this:

```
this.documentBuilder = this.indexType.getDocumentBuilder()
```

Or it could use a factory.

```
this.documentBuilder = documentBuilderFactory.create(this.indexType);
```

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilder.java
```diff
@@ -0,0 +1,267 @@
...
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+
+public final class TSDBDocumentBuilder {
```
dlvenable: As noted below, let's make an interface and then this class can inherit that interface.

data-prepper-plugins/opensearch/src/test/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilderTest.java
```diff
@@ -0,0 +1,421 @@
...
+    private static final long TEST_TIMESTAMP_MILLIS = 1706869800000L;
+
+    private TSDBDocumentBuilder builder;
+
+    @BeforeEach
+    void setUp() {
+        builder = new TSDBDocumentBuilder();
+    }
+
+    // --- Gauge Tests ---
```
dlvenable: Please avoid these types of comments. You can use `@Nested` if you really want to keep the code together.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilder.java
```diff
@@ -0,0 +1,267 @@
...
+        }
+        if (!extraInserted) {
+            sb.append(' ').append(extraKey).append(' ').append(extraValue);
+        }
+
+        return sb.toString();
+    }
+
+    private static String buildJsonDoc(final String labels, final long timestamp, final double value) {
+        final StringBuilder sb = new StringBuilder(labels.length() + 64);
```
dlvenable: We should build the JSON using Jackson instead of `StringBuilder` to avoid errors. You should be able to use `JsonNode` for fast creation and serialization.

There are a few other instances of this here.

data-prepper-plugins/opensearch/src/test/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilderTest.java
```diff
@@ -0,0 +1,421 @@
...
+                .build();
+
+        final List<String> docs = builder.build(histogram);
+
+        // 4 bucket docs + 1 _count + 1 _sum = 6
+        assertEquals(6, docs.size());
+
+        // Bucket 1: cumulative = 5
+        final Map<String, Object> bucket1 = parseJson(docs.get(0));
+        assertEquals("__name__ request_duration_bucket le 0.1 method GET", bucket1.get("labels"));
```
dlvenable, line 157: What motivates this label convention of using spaces here? 

Can we store labels as a nested field?

```
"labels": {
      "__name__": "request_duration_bucket",
      "le": "0.1",
      "method": "GET"
},
```

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -470,6 +475,32 @@ public void doOutput(final Collection<Record<Event>> records) {
...
+      if (indexType == IndexType.TSDB) {
+        try {
+          final List<String> tsdbDocs = tsdbDocumentBuilder.build(event);
+          final String tsdbAction = resolveEventAction(event);
+          final List<BulkOperationWrapper> wrappers = new ArrayList<>(tsdbDocs.size());
+          for (int i = 0; i < tsdbDocs.size(); i++) {
+            final SerializedJson doc = SerializedJson.fromStringAndOptionals(tsdbDocs.get(i), null, null, null);
+            final BulkOperation op = getBulkOperationForAction(tsdbAction, doc, null, indexName, null);
+            final BulkOperationWrapper wrapper = (i == tsdbDocs.size() - 1)
+                    ? new BulkOperationWrapper(op, event.getEventHandle(), null, null)
```
dlvenable: We support aggregate event handles which create aggregations. It may be ideal to have a split or multiple event handle which handles additional acquires so that we can have multiple releases here.

The current approach will acknowledge with only the last event handle succeeding.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/IndexManagerFactory.java
```diff
@@ -56,6 +56,7 @@ public final IndexManager getIndexManager(final IndexType indexType,
                         restHighLevelClient, openSearchClient, openSearchSinkConfiguration, clusterSettingsParser, templateStrategy, indexAlias);
                 break;
             case TRACE_ANALYTICS_SERVICE_MAP:
+            case TSDB:
```
dlvenable: Why is this connected with the `TraceAnalyticsServiceMapIndexManager`? Why not the default?

data-prepper-plugins/opensearch/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSinkTsdbIT.java
```diff
@@ -0,0 +1,532 @@
...
+    }
+
+    @AfterEach
+    void closeClient() throws IOException {
+        if (client != null) {
+            client.close();
+        }
+    }
+
+    // --- Index Initialization ---
```
dlvenable: Let's remove these as well.

dlvenable (changes requested):
Thank you @srikanthpadakanti for this contribution!

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
srikanthpadakanti: Agreed, will add ..

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilder.java
srikanthpadakanti: sure

data-prepper-plugins/opensearch/src/test/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilderTest.java
srikanthpadakanti: Will remove.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilder.java
srikanthpadakanti: fair enough.

data-prepper-plugins/opensearch/src/test/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilderTest.java
srikanthpadakanti: The space-separated format is what the OpenSearch TSDB plugin expects. Internally it splits the labels string, binary-encodes the pairs into ByteLabels, and runs MurmurHash3 on them to compute stable shard routing so all samples for a series land on the same shard. A nested object would not be parsed correctly by the TSDB plugin. This format matches [TSDBDocument.java](https://github.com/opensearch-project/time-series-db/blob/main/src/main/java/org/opensearch/index/engine/TSDBDocument.java) in the TSDB plugin.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
srikanthpadakanti: Will use aggregate event handles.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/IndexManagerFactory.java
srikanthpadakanti: Will move to default.

data-prepper-plugins/opensearch/src/integrationTest/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSinkTsdbIT.java
srikanthpadakanti: ok..

data-prepper-plugins/opensearch/src/test/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilderTest.java
dlvenable: This makes sense. Thanks for clarifying and linking the model.

dlvenable (dismissed):
Thank you @srikanthpadakanti for this contribution!

dlvenable:
@kkondaka , Could you take a review of this as you have worked quite a bit with Prometheus?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/index/TSDBDocumentBuilder.java
```diff
@@ -0,0 +1,240 @@
...
+                                                   final String[][] sortedAttrs, final long timestamp) {
+        documents.add(buildJsonDoc(
+                buildLabels(baseName + "_count", sortedAttrs, null, null),
+                timestamp, count != null ? count.doubleValue() : 0.0));
+        documents.add(buildJsonDoc(
+                buildLabels(baseName + "_sum", sortedAttrs, null, null),
+                timestamp, sum != null ? sum : 0.0));
+    }
+
+    private static String[][] sortAndSanitizeAttributes(final Map<String, Object> attributes) {
```
kkondaka, line 134: Looks like some of this code is similar to what is in the Prometheus sink. Should we move this code to some common library?

srikanthpadakanti: The sanitization logic here serves a different purpose. PrometheusTimeSeries sanitizes for Prometheus remote write naming conventions (unit suffixes, OpenMetrics rules), while TSDBDocumentBuilder sanitizes for the TSDB space-delimited keyword format (spaces to underscores, key prefix rules). 

The methods share a similar name but the actual logic does not overlap. Extracting to a common library would couple two different output format requirements.

dlvenable: @srikanthpadakanti , On this topic, does Prometheus have any Java libraries that might have similar logic?

srikanthpadakanti: @dlvenable Checked `PrometheusNaming` in [prometheus-metrics-model](https://github.com/prometheus/client_java/blob/main/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/PrometheusNaming.java). It validates and escapes names for Prometheus exposition formats but does not do the kind of transformation we need. Their `sanitizeLabelName` only strips leading double-underscore prefixes, it does not replace invalid characters. They have no equivalent for replacing spaces in values (our TSDB delimiter requirement) or building sorted key-value label strings. The methods solve different problems at different layers.

dlvenable:
@srikanthpadakanti , Can you rebase this and resolve conflicts? It has a conflict preventing automatic merge.

@kkondaka , Can you give it another review? It looks good to me.

srikanthpadakanti:
> @srikanthpadakanti , Can you rebase this and resolve conflicts? It has a conflict preventing automatic merge.
> 
> @kkondaka , Can you give it another review? It looks good to me.

Rebased and conflicts are resolved

dlvenable (dismissed):
Thank you for the contribution @srikanthpadakanti !

srikanthpadakanti:
Hello @kkondaka Please review this.

dlvenable (approved):

kkondaka (approved):

PR #6716, opened 2026-04-02
Fix MongoTasksRefresher to force executor restart on MongoSecurityException

dinujoh:
### Description
When DocumentDB revokes old credentials after secret rotation, the pipeline enters a permanent auth failure state because basicAuthChanged() returns false (the secret value in Secrets Manager hasn't changed).

Add forceRefresh() to MongoTasksRefresher with exponential backoff (30s, 60s, 120s, max 3 attempts) that restarts the executor with the current config.

StreamScheduler now walks the exception cause chain for MongoSecurityException and calls forceRefresh() when detected. If all 3 attempts fail, falls back to the normal hourly scheduled credential refresh.
 
 
### Check List
- [ x] Commits are signed with a real name per the DCO

### DIFF
--- a/data-prepper-plugins/mongodb/src/test/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamSchedulerTest.java
+++ b/data-prepper-plugins/mongodb/src/test/java/org/opensearch/dataprepper/plugins/mongo/stream/StreamSchedulerTest.java
@@ -1,7 +1,6 @@
 package org.opensearch.dataprepper.plugins.mongo.stream;
 
 import com.mongodb.MongoSecurityException;
-import org.bson.BsonDocument;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;


data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/documentdb/MongoTasksRefresher.java
```diff
@@ -81,13 +85,40 @@ public void update(MongoDBSourceConfig pluginConfig) {
...
+        final long backoff = BASE_BACKOFF_MS * (1L << forceRefreshAttempts);
+        if (now - lastForceRefreshTime < backoff) {
+            return;
+        }
+        lastForceRefreshTime = now;
+        forceRefreshAttempts++;
+        LOG.info("Forcing credential refresh due to authentication failure (attempt {}/{})",
+                forceRefreshAttempts, MAX_FORCE_REFRESH_ATTEMPTS);
+        try {
+            currentExecutor.shutdownNow();
```
graytaylor0, line 113: Won't this block submission of  jobs to executor service? I don't see executor service getting reinitialized just a new StreamScheduler job getting submitted

graytaylor0: Nevermind I see this creates a new executor

```
this.currentExecutor = executorServiceFunction.apply(runnables.size());
```

dinujoh: it's initialized in refreshJobs method: https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/mongodb/src/main/java/org/opensearch/dataprepper/plugins/mongo/documentdb/MongoTasksRefresher.java#L103

PR #6718, opened 2026-04-03
Fix flaky PipelinesWithAcksIT by awaiting async ack callback

lawofcycles:
### Description
Fix flaky `PipelinesWithAcksIT` tests that intermittently fail with `NullPointerException`.

The acknowledgement callback (`setAckReceived`) is invoked asynchronously after sink processing completes. The existing tests assert the ack result immediately after confirming sink output arrival, without waiting for the async callback. Under CI load, this race condition causes `getAckReceived()` to return `null`, which triggers a `NullPointerException` when unboxed by `assertTrue(Boolean)`.

This change wraps all ack assertions in `await().untilAsserted()` to poll until the acknowledgement callback has been invoked, consistent with the existing pattern already used in `three_pipelines_with_all_unrouted_records()`.

### CI impact
This fix targets the `Gradle Build / build` workflow (`gradle.yml`), which runs `./gradlew build` including `:data-prepper-core:integrationTest`. The following test methods in `PipelinesWithAcksIT` have been observed to fail intermittently across Java 11, 17, and 21 matrix jobs.

Local verification (10 consecutive runs each):
* Before: 8/10 passed (2 failures)
* After: 10/10 passed (0 failures)

### Issues Resolved
Related #3481

### Check List
- [] New functionality includes testing.

### DIFF
--- a/data-prepper-core/src/integrationTest/java/org/opensearch/dataprepper/integration/PipelinesWithAcksIT.java
+++ b/data-prepper-core/src/integrationTest/java/org/opensearch/dataprepper/integration/PipelinesWithAcksIT.java
@@ -30,9 +30,7 @@ import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.lessThanOrEqualTo;
-import static org.junit.Assert.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @FixMethodOrder()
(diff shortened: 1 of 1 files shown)


lawofcycles:
I found another flaky test: https://github.com/opensearch-project/data-prepper/actions/runs/23929544444/job/69793605060?pr=6718
Filed https://github.com/opensearch-project/data-prepper/issues/6719

dlvenable (approved):
Thanks @lawofcycles for this fix! It is very much needed.

lawofcycles:
@dlvenable Thank you for the review. Yes, I understand in data prepper, stabilizing CI is very challenging because it includes so many dependencies and scenarios. I'll continue to try to reduce flaky tests for a better developer experience, including for myself.

PR #6720, opened 2026-04-03
Fix flaky DefaultAcknowledgementSetManagerTests

### LINKED ISSUE #6719: [BUG] Flaky DefaultAcknowledgementSetManagerTests
**Describe the bug**
`DefaultAcknowledgementSetManagerTests.testExpirations()` intermittently fails in CI with:

```
java.lang.AssertionError:
Expected: <0>
    but: was <1>
at org.opensearch.dataprepper.core.acknowledgements.DefaultAcknowledgementSetManagerTests.testExpirations(
DefaultAcknowledgementSetManagerTests.java:113)
```

**To Reproduce**
Run `./gradlew :data-prepper-core:test --tests "org.opensearch.dataprepper.core.acknowledgements.DefaultAcknowledgementSetManagerTests.testExpirations" --rerun` repeatedly. The failure is intermittent and more likely under CPU load.

**Expected behavior**
The test should pass consistently regardless of system load.

 - OS: Ubuntu (GitHub Actions runner)
 - Java: 11
 - Observed in: `Gradle Build / build (11)` on PR #6712

**Additional context**
The test sleeps for `TEST_TIMEOUT * 2` (800ms) and then immediately asserts that the acknowledgement set monitor size is 0. This assumes the monitor's cleanup thread has already removed the expired set during the sleep window. Under CI load, the cleanup may not have completed in time, leaving the size at 1.

The subsequent `await().atMost(TEST_TIMEOUT.multipliedBy(3))` only checks the `result` field, not the monitor size, so the stale monitor size is never retried.

A fix would be to move the `getSize() == 0` assertion into an `await().untilAsserted()` block, similar to the pattern used in `testBasic()`.

lawofcycles:
### Description
Fix flaky `DefaultAcknowledgementSetManagerTests.testExpirations()` that intermittently fails with `AssertionError: Expected: <0> but: was <1>`.

The test asserts that the acknowledgement set monitor size is 0 immediately after a `Thread.sleep()`, assuming the cleanup thread has already removed the expired set. Under CI load, the cleanup may not have completed in time. The subsequent `await()` block only checks the `result` field, not the monitor size, so the stale size is never retried.

This change moves the `getSize() == 0` assertion into the `await().untilAsserted()` block so both conditions are polled together.

### CI impact
This fix targets `Gradle Build / build` workflow (`gradle.yml`), specifically `DefaultAcknowledgementSetManagerTests.testExpirations()` in `:data-prepper-core:test`.

Local verification (10 consecutive runs each):
The race condition is difficult to reproduce with the default `TEST_TIMEOUT` (400ms) on a local machine. To amplify it, the `Thread.sleep` in `testExpirations` was temporarily hardcoded to 20ms (instead of `TEST_TIMEOUT * 2` = 800ms), reducing the margin for the cleanup thread to near zero.
* Before fix (getSize assertion outside await): 0/10 passed
* After fix (getSize assertion inside await): 10/10 passed

### Issues Resolved
Resolves https://github.com/opensearch-project/data-prepper/issues/6719

### Check List

### DIFF
--- a/data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetManagerTests.java
+++ b/data-prepper-core/src/test/java/org/opensearch/dataprepper/core/acknowledgements/DefaultAcknowledgementSetManagerTests.java
@@ -110,9 +110,9 @@ class DefaultAcknowledgementSetManagerTests {
     void testExpirations() throws InterruptedException {
         eventHandle2.release(true);
         Thread.sleep(TEST_TIMEOUT.multipliedBy(2).toMillis());
-        assertThat(acknowledgementSetManager.getAcknowledgementSetMonitor().getSize(), equalTo(0));
         await().atMost(TEST_TIMEOUT.multipliedBy(3))
                 .untilAsserted(() -> {
+                    assertThat(acknowledgementSetManager.getAcknowledgementSetMonitor().getSize(), equalTo(0));
                     assertThat(result, equalTo(null));
                 });
     }


lawofcycles:
found another flaky test: https://github.com/opensearch-project/data-prepper/actions/runs/23930747072/job/69797135261?pr=6720
Cut: https://github.com/opensearch-project/data-prepper/issues/6721

dlvenable (approved):

kkondaka (approved):

Issue #6726, opened 2026-04-03
[FEATURE] Support shared catalog config and per-table shuffle overrides in iceberg-source


lawofcycles:
**Is your feature request related to a problem? Please describe.**

When configuring multiple tables in iceberg-source, users must repeat the same catalog properties for each table entry even when the tables share the same catalog. Additionally, shuffle parameters (`partitions`, `target_partition_size`) cannot be tuned per table.

**Describe the solution you'd like**

1. Allow a top level `catalog` definition that applies to all tables by default. When a table specifies its own `catalog`, it fully replaces the top level definition (no partial merge).

```yaml
iceberg:
 catalog:
   type: rest
   uri: "http://iceberg-rest-catalog:8181"
   io-impl: "org.apache.iceberg.aws.s3.S3FileIO"
 tables:
    - table_name: "db.table_a"
     identifier_columns: ["id"]
    - table_name: "db.table_b"
     identifier_columns: ["id"]
     catalog:
       type: glue
       warehouse: "s3://other-bucket/warehouse"
       io-impl: "org.apache.iceberg.aws.s3.S3FileIO"
```

2. Allow per table overrides for shuffle parameters (`partitions`, `target_partition_size`). When a table specifies its own `shuffle`, it fully replaces the top level shuffle parameters. Node level settings like `server_port` and `ssl` remain top level only and are not overridable per table.

```yaml
iceberg:
 shuffle:
   partitions: 64
   target_partition_size: 64mb
 tables:
    - table_name: "db.large_table"
     shuffle:
       partitions: 256
       target_partition_size: 128mb
    - table_name: "db.small_table"
```
**Additional context**

Related PR: #6682 (source-layer shuffle implementation)

lawofcycles:
latest status:

- Shared catalog config (item 1) shipped in #6727. A per-table `catalog` fully replaces the top-level one, as proposed.
- Per-table shuffle overrides (item 2) are not implemented yet. #6682 added the top-level `shuffle` settings and has been merged, so this part is now unblocked, but `TableConfig` still has no `shuffle` field.
- Until then, writing `shuffle` under a table entry fails config deserialization.

Keeping this open for item 2.

PR #6727, opened 2026-04-04
Support shared catalog config across tables in iceberg-source

lawofcycles:
### Description

Adds support for a top level `catalog` configuration in iceberg-source that applies to all tables by default. When a table specifies its own `catalog`, it fully replaces the top level definition.

This reduces configuration duplication when multiple tables share the same catalog.

```yaml
iceberg:
 catalog:
   type: rest
   uri: "http://iceberg-rest-catalog:8181"
   io-impl: "org.apache.iceberg.aws.s3.S3FileIO"
 tables:
    - table_name: "db.table_a"
     identifier_columns: ["id"]
    - table_name: "db.table_b"
     identifier_columns: ["id"]
     catalog:
       type: glue
       warehouse: "s3://other-bucket/warehouse"
       io-impl: "org.apache.iceberg.aws.s3.S3FileIO"
```
### Issues Resolved
Partially addresses #6726 (catalog config sharing). Per-table shuffle overrides will be addressed after #6682 is merged.

### Check List

### DIFF
--- a/data-prepper-plugins/iceberg-source/src/integrationTest/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergSourceIT.java
+++ b/data-prepper-plugins/iceberg-source/src/integrationTest/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergSourceIT.java
@@ -141,6 +141,74 @@ public class IcebergSourceIT {
     }
 
     @Test
+    void export_with_shared_catalog_config() throws Exception {
+        final Table table = helper.createTable(TEST_NAMESPACE, TEST_TABLE, TEST_SCHEMA);
+        helper.appendRows(table, List.of(
+                helper.newRecord(TEST_SCHEMA, 1, "Alice", 30),
+                helper.newRecord(TEST_SCHEMA, 2, "Bob", 25)
+        ));
+
+        final IcebergService service = createService(false, true);
+        final Buffer<org.opensearch.dataprepper.model.record.Record<Event>> buffer = createMockBuffer();
+        service.start(buffer);
+
+        try {
+            await().atMost(30, TimeUnit.SECONDS)
+                    .untilAsserted(() -> assertThat(receivedRecords, hasSize(greaterThanOrEqualTo(2))));
+            assertThat(receivedRecords, hasSize(2));
+        } finally {
+            service.shutdown();
+        }
+    }
+
+    @Test
+    void export_with_mixed_catalog_config() throws Exception {
+        // Table A uses shared catalog, Table B overrides with its own catalog (same catalog in test, but exercises the override path)
+        final String tableNameA = "mixed_a";
+        final String tableNameB = "mixed_b";
+        final Table tableA = helper.createTable(TEST_NAMESPACE, tableNameA, TEST_SCHEMA);
+        final Table tableB = helper.createTable(TEST_NAMESPACE, tableNameB, TEST_SCHEMA);
+        helper.appendRows(tableA, List.of(helper.newRecord(TEST_SCHEMA, 1, "Alice", 30)));
+        helper.appendRows(tableB, List.of(helper.newRecord(TEST_SCHEMA, 2, "Bob", 25)));
+
+        final IcebergSourceConfig sourceConfig = mock(IcebergSourceConfig.class);
+        final TableConfig configA = mock(TableConfig.class);
+        final TableConfig configB = mock(TableConfig.class);
+
+        when(configA.getTableName()).thenReturn(TEST_NAMESPACE + "." + tableNameA);
+        when(configA.getCatalog()).thenReturn(Collections.emptyMap());
+        when(configA.getIdentifierColumns()).thenReturn(List.of("id"));
+        when(configA.isDisableExport()).thenReturn(false);
+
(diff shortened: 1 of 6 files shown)


data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergService.java
```diff
@@ -68,7 +68,8 @@ public void start(final Buffer<Record<Event>> buffer) {
             final String tableName = tableConfig.getTableName();
             LOG.info("Loading catalog and table for {}", tableName);
 
-            final Map<String, String> catalogProps = new HashMap<>(tableConfig.getCatalog());
+            final Map<String, String> catalogProps = new HashMap<>(
+                    tableConfig.getCatalog().isEmpty() ? sourceConfig.getCatalog() : tableConfig.getCatalog());
```
dlvenable: Should we use `!= null` instead? Perhaps an empty catalog can be an explicit no catalog at all.

data-prepper-plugins/iceberg-source/src/integrationTest/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergSourceIT.java
```diff
@@ -140,6 +140,74 @@ void export_writes_all_rows_to_buffer() throws Exception {
...
+        when(configA.getCatalog()).thenReturn(Collections.emptyMap());
+        when(configA.getIdentifierColumns()).thenReturn(List.of("id"));
+        when(configA.isDisableExport()).thenReturn(false);
+
+        when(configB.getTableName()).thenReturn(TEST_NAMESPACE + "." + tableNameB);
+        when(configB.getCatalog()).thenReturn(helper.catalogProperties());
+        when(configB.getIdentifierColumns()).thenReturn(List.of("id"));
+        when(configB.isDisableExport()).thenReturn(false);
+
+        when(sourceConfig.getCatalog()).thenReturn(helper.catalogProperties());
```
dlvenable, line 189: Should we use an alternative catalog to be sure that it is correctly selected?

data-prepper-plugins/iceberg-source/src/integrationTest/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergSourceIT.java
```diff
@@ -140,6 +140,74 @@ void export_writes_all_rows_to_buffer() throws Exception {
...
+        when(sourceConfig.getTables()).thenReturn(List.of(configA, configB));
+        when(sourceConfig.getPollingInterval()).thenReturn(Duration.ofSeconds(5));
+        when(sourceConfig.isAcknowledgmentsEnabled()).thenReturn(false);
+
+        final EnhancedSourceCoordinator coordinator = createInMemoryCoordinator();
+        coordinator.createPartition(new LeaderPartition());
+        final IcebergService service = new IcebergService(coordinator, sourceConfig, pluginMetrics,
+                acknowledgementSetManager, org.opensearch.dataprepper.event.TestEventFactory.getTestEventFactory());
```
dlvenable: Use an import on the class or static import. Avoid fully qualified names.

dlvenable (changes requested):
Thanks @lawofcycles for this contribution! I have a few small comments.

data-prepper-plugins/iceberg-source/src/main/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergService.java
lawofcycles: Good point. Changed to `!= null` and updated `TableConfig.catalog` default from `Collections.emptyMap()` to `null`. This way, an unset catalog falls back to the shared config, while an explicitly empty `catalog: {}` is passed through as is (which will fail at catalog initialization.

data-prepper-plugins/iceberg-source/src/integrationTest/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergSourceIT.java
lawofcycles: Fixed.

data-prepper-plugins/iceberg-source/src/integrationTest/java/org/opensearch/dataprepper/plugins/source/iceberg/IcebergSourceIT.java
lawofcycles: Since the integration test environment has a single REST catalog server, using a truly distinct catalog to verify selection is not straightforward. I added `IcebergServiceTest` with unit tests that use `mockStatic(CatalogUtil.class)` to capture and assert the exact catalog properties passed to `buildIcebergCatalog` for each table. The integration test `export_with_mixed_catalog_config` is kept as a smoke test to confirm the mixed configuration works end to end.

dlvenable (approved):
Thank you @lawofcycles !

divbok (approved):

PR #6627, opened 2026-03-10
Prometheus Remote Write v1 Source 

srikanthpadakanti:
### Description
 Adds a new Prometheus Remote Write v1 source plugin (prometheus) that acts as a Remote Write receiver, accepting metrics pushed from Prometheus servers or other Remote Write compatible clients.

  Key features:
  - Receives Snappy-compressed protobuf payloads over HTTP POST (Prometheus Remote Write v1 protocol)
  - Infers metric types from naming conventions since Remote Write v1 does not carry type metadata:
    - _bucket suffix with le label --> Histogram
    - quantile label --> Summary
    - _total / _created suffix--> Sum (monotonic counter)
    - Everything else → Gauge
  - Uses a two-pass grouping strategy to correctly associate _count, _sum, _bucket, and quantile TimeSeries into composite Histogram/Summary events
  without misclassifying standalone _count metrics as histogram counts
  - Converts Prometheus cumulative histogram buckets to per-bucket counts
  - Supports SSL/TLS, authentication, health checks, and configurable server options (port, path, thread count, max connections, request timeout,
  etc.)
  - Supports flat label mode via flatten_labels configuration
  - Extracts service name from labels using priority order: service.name > service_name > job
  - Includes comprehensive metrics: requestsReceived, successRequests, failedRequests, payloadSize, recordsCreated, requestProcessDuration

  Configuration example:
  prometheus-source:
    port: 9090
    path: /api/v1/write
    ssl: false

 
### Issues Resolved
https://github.com/opensearch-project/data-prepper/issues/6533
Resolves #6533
 
### Check List
- [ X ] New functionality includes testing.
- [ ] New functionality has a documentation issue. Please link to it in this PR.
  - [ X ] New functionality has javadoc added
- [ X ] Commits are signed with a real name per the DCO

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/data-prepper/blob/main/CONTRIBUTING.md).

PR #6743, opened 2026-04-08
Add Prometheus scrape/pull source to prometheus plugin [Adding in the same Prometheus Remote write source] 

### LINKED ISSUE #1997: Ingest data from Prometheus as Source
A clear and concise description of what the problem is. Ex. It would be nice to have [...]

**Additional context**

srikanthpadakanti:
### Description
Add Prometheus scrape/pull source to the existing prometheus plugin. The source periodically HTTP GETs target `/metrics` endpoints, parses the Prometheus text exposition format, and converts metrics into Data Prepper events (Gauge, Sum, Histogram, Summary). Supports static target URLs, configurable scrape interval/timeout, HTTP Basic Auth, Bearer Token, and TLS certificates. Both scrape and remote write modes can run simultaneously under the same plugin. 

Three usage modes:

```
 Mode 1: Remote write only (existing behavior, unchanged)
  source:
    prometheus:
      port: 9090

Mode 2: Scrape only
  source:
    prometheus:
      scrape:
        targets:
          - url: "http://host:9090/metrics"

Mode 3: Both simultaneously
  source:
    prometheus:
      port: 9090
      scrape:
        targets:
          - url: "http://host:9090/metrics"
```

 
### Issues Resolved
https://github.com/opensearch-project/data-prepper/issues/1997
Resolves #1997
 
### Check List


data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeAuthenticationConfig.java
```diff
@@ -0,0 +1,51 @@
...
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotBlank;
+
+public class ScrapeAuthenticationConfig {
+
+    @Valid
+    @JsonProperty("http_basic")
+    private HttpBasicConfig httpBasic;
+
+    @JsonProperty("bearer_token")
```
dlvenable, line 22: Maybe this should be an `oauth2` group with only `bearer_token` for consistency.

```
scrape:
  oauth2:
    bearer_token:
```

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeTargetScraper.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Wraps Armeria WebClient to scrape a single Prometheus /metrics endpoint.
+ */
+public class ScrapeTargetScraper {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ScrapeTargetScraper.class);
+    private static final String ACCEPT_HEADER_VALUE = "text/plain;version=0.0.4";
```
dlvenable, line 40: Does the version direct Prometheus to provide the scrape data in a certain format? Will it return Not Acceptable if it doesn't support the requested version?

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeTargetScraper.java
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * Scrapes the given Prometheus metrics endpoint URL.
+     *
+     * @param url the full URL of the metrics endpoint to scrape
+     * @return the response body as a String
+     * @throws RuntimeException if the response status is not 2xx
+     */
+    public String scrape(final String url) {
+        final URI uri = URI.create(url);
+        final String baseUri = uri.getScheme() + "://" + uri.getAuthority();
```
dlvenable: Let's use Spring's `UriComponentsBuilder` instead of string manipulation. If Armeria has something like this it would be ideal to use that instead.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeTargetScraper.java
```diff
@@ -0,0 +1,134 @@
...
+        if (isCustomFactory) {
+            clientFactory.close();
+        }
+    }
+
+    private HttpHeaders buildCommonHeaders() {
+        final HttpHeadersBuilder builder = HttpHeaders.builder()
+                .add(HttpHeaderNames.ACCEPT, ACCEPT_HEADER_VALUE);
+
+        if (config.getAuthentication() != null) {
```
dlvenable: It would probably be ideal to have small classes and a common interface for updating the request based on the authentication. This could allow more complicated auth schemes like SigV4 in the future.

I had some similar comments on an `http` sink PR that is in progress currently. https://github.com/opensearch-project/data-prepper/pull/6747#discussion_r3065089882. Though my comments went the other way. Have the code more amenable to basic auth instead of SigV4. 😄

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/TextExpositionParser.java
```diff
@@ -0,0 +1,616 @@
...
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+public class TextExpositionParser {
```
dlvenable, line 33: What is the relation between this and `RemoteWriteProtobufParser`. Is there opportunity for shared code for parsing?

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/PrometheusScrapeConfig.java
```diff
@@ -0,0 +1,70 @@
...
+    @JsonProperty("scrape_interval")
+    private Duration scrapeInterval = Duration.ofSeconds(15);
+
+    @JsonProperty("scrape_timeout")
+    private Duration scrapeTimeout = Duration.ofSeconds(10);
+
+    @JsonProperty("flatten_labels")
+    private boolean flattenLabels = false;
+
+    @JsonProperty("insecure")
```
dlvenable, line 36: If any of the URLs in `targets` has a non-HTTPS scheme we should throw a validation error unless `insecure` is true.

You can add a JSR `@AssertTrue` to handle this in this class.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/PrometheusScrapeService.java
```diff
@@ -0,0 +1,126 @@
...
+                LOG.error("Error scraping target {}", url, e);
+            }
+        });
+    }
+
+    public void stop() {
+        try {
+            executor.shutdown();
+            try {
+                final long timeoutMs = config.getScrapeTimeout().toMillis() + 1000;
```
dlvenable, line 114: This timeout grows without bounds. Armeria has an exponential backoff utility. Maybe there is a linear one as well.

dlvenable (changes requested):
Thanks @srikanthpadakanti . 

At a high-level would a scraper be expected to get data from the same Prometheus nodes pushing? I'm wondering if it makes sense to have a separate source or not.

srikanthpadakanti:
> Thanks @srikanthpadakanti .
> 
> At a high-level would a scraper be expected to get data from the same Prometheus nodes pushing? I'm wondering if it makes sense to have a separate source or not.

A scraper would not typically target the same nodes pushing via remote write. They serve different use cases: remote write is Prometheus servers configured to push, while scrape pulls from any endpoint exposing the Prometheus text exposition format, including application instances that are not Prometheus servers themselves.
                                                                                                                               
The reason they share a plugin is the precedent from kkondaka's feedback on PR #[6627](https://github.com/opensearch-project/data-prepper/pull/6627#discussion_r3001615902) where the recommendation was to keep it as a single prometheus source with config options for different modes, similar to how `s3` source handles `s3-sqs` and `s3-scan` under one plugin. Both modes can also run simultaneously which simplifies pipeline configuration when users need both ingestion paths.

That said, if you prefer a separate source I am open to splitting it out.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeAuthenticationConfig.java
srikanthpadakanti: Bearer token here is a static token passed as an HTTP Authorization header, not an OAuth2 flow. Nesting it under oauth2 would imply token endpoint and refresh support that we do not have. Kept it flat under authentication to match the actual behavior. 

Happy to add an oauth2 group later when we support the full flow with token exchange.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeTargetScraper.java
srikanthpadakanti: Yes. The `version=0.0.4` in the Accept header identifies the Prometheus text exposition format per the spec. If the version parameter is missing, the server falls back to the most recent text format version which is still `0.0.4` 

Prometheus servers do not return `406` for this, they serve the default format. Content negotiation between text, OpenMetrics, and protobuf is handled via distinct content types (`text/plain`, `application/openmetrics-text`, `application/vnd.google.protobuf`). We can add OpenMetrics support as a follow-up.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeTargetScraper.java
srikanthpadakanti: Agreed on avoiding string manipulation. Will switch to `java.net.URI` constructor to extract the base URI cleanly rather than adding a Spring dependency. Armeria's WebClient already handles URI parsing internally so we just need the base extraction to be safe.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeTargetScraper.java
srikanthpadakanti: Extracted into a `ScrapeRequestAuthenticator` interface with `BasicAuthenticator` and `BearerTokenAuthenticator` as separate classes. Factory method handles selection. Adding `SigV4` or other schemes in the future is just a new implementation without touching the scraper.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/TextExpositionParser.java
srikanthpadakanti: Different input formats with no parsing overlap. Text parser handles line-oriented exposition, protobuf parser handles Snappy-compressed binary.

Both already share the same output builders (JacksonGauge, JacksonSum, etc.) from data-prepper-api. No practical code to extract.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/PrometheusScrapeConfig.java
srikanthpadakanti: Done. 
Added @AssertTrue validation.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/PrometheusScrapeService.java
srikanthpadakanti: The timeout here is fixed, not growing. It is scrapeTimeout + 1 second, used once in awaitTermination to allow the in-flight scrape to finish before forcing shutdown. There is no retry loop or escalation.

dlvenable:
> > Thanks @srikanthpadakanti .
> > At a high-level would a scraper be expected to get data from the same Prometheus nodes pushing? I'm wondering if it makes sense to have a separate source or not.
> 
> A scraper would not typically target the same nodes pushing via remote write. They serve different use cases: remote write is Prometheus servers configured to push, while scrape pulls from any endpoint exposing the Prometheus text exposition format, including application instances that are not Prometheus servers themselves.
> 
> The reason they share a plugin is the precedent from kkondaka's feedback on PR #[6627](https://github.com/opensearch-project/data-prepper/pull/6627#discussion_r3001615902) where the recommendation was to keep it as a single prometheus source with config options for different modes, similar to how `s3` source handles `s3-sqs` and `s3-scan` under one plugin. Both modes can also run simultaneously which simplifies pipeline configuration when users need both ingestion paths.
> 
> That said, if you prefer a separate source I am open to splitting it out.

@srikanthpadakanti , @kkondaka ,

I think this should be a different source and that the S3 analogy doesn't quite fit here.

The S3 source is entirely pull-based. Whether it is scan or SQS or EventBridge, it is pulling the data from S3. In no case does it open a port. The scan and the SQS are just two different ways to discover the content. The S3 source pulls the data from S3 the exact same way and uses the same code path regardless of the discovery. Additionally we had also once wanted the ability to scan and then start using SQS for a CDC-like behavior, similar to what we do for DynamoDB/RDS.

In this case, there is a big difference between opening a port and going out to an external system to make requests. The source configuration will have a security difference depending on the configuration and may not be clear.

I recommend that we split this into a `prometheus_pull` source (maybe there is a better name). @kkondaka , please chime in.

kkondaka:
@dlvenable Whether it is pull or push is not something users/customers know about or care about, right? Why make it difficult for the users?

dlvenable:
@kkondaka ,

> Whether it is pull or push is not something users/customers know about or care about, right? Why make it difficult for the users?

I think this is not at all the case. Users must know whether it is push or pull because that is how they get the data there right? If it is pull-based, they must enable a scrape URI and provide that. If it is push-based they must configure their remote-write protocol.

I think this simplifies it for users because these are in reality two different sources with disjoint configurations. The documentation will need to say for just about every field whether it is remote-write or pull/scape. It seems more straightforward as a user to start with one model and see the configurations only for that model.

In what way would two sources make the the configuration more complicated for users?

dlvenable:
@kkondaka , I suggest that we accept this PR and discuss further on the source.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeAuthenticationConfig.java
```diff
@@ -0,0 +1,51 @@
...
+package org.opensearch.dataprepper.plugins.source.prometheus;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotBlank;
+
+public class ScrapeAuthenticationConfig {
+
+    @Valid
+    @JsonProperty("http_basic")
```
kkondaka, line 19: Looks like these auth mechanisms are common for all http. We should move them to a common location.

kkondaka (approved):

srikanthpadakanti:
> @kkondaka , I suggest that we accept this PR and discuss further on the source.

Makes sense. Opened an [issue](https://github.com/opensearch-project/data-prepper/issues/6767) to track moving these to a shared module. Will do that in a follow-up PR so other HTTP plugins can reuse them.

data-prepper-plugins/prometheus-source/src/main/java/org/opensearch/dataprepper/plugins/source/prometheus/ScrapeAuthenticationConfig.java
srikanthpadakanti: acknowledge, offer to do it as a follow-up

srikanthpadakanti: I have created a new issue for the followup - https://github.com/opensearch-project/data-prepper/issues/6766

kkondaka (approved):

dlvenable (approved):
Thank you @srikanthpadakanti for this contribution!

PR #6744, opened 2026-04-09
Add generic script support for update/upsert operations in OpenSearch sink

### LINKED ISSUE #3563: Support conditional script update of documents in the OpenSearch sink
As a user of the OpenSearch sink, I would like to conditionally update documents by comparing the existing document in OpenSearch with the updated document. 

An option to specify a `script` parameter that will apply to update and upsert operation in the sink (https://opensearch.org/docs/latest/api-reference/document-apis/update-document/#script-example).

This requires that the incoming update documents contain the fields that are referenced in the `script`. 

One alternative to having OpenSearch handle this would be to have Data Prepper handle it, which would mean an extra read of each document with an `update` or `upsert` operation, and then relying on Data Prepper expression as the condition instead of the `script`. the benefit of this approach would be that we could support comparisons on Data Prepper metadata, instead of requiring the data be in the Event.

**Additional context**

dinujoh:
### Description
Adds generic script support for update/upsert operations in the OpenSearch sink. When configured, the sink attaches an inline script (https://opensearch.org/docs/latest/api-reference/document-apis/update-document/#script-example) to bulk update/upsert requests.                                                                              
                                                                                                                                                                                                                 
Key behaviors:                                                                                                                                                                                                                                                                                                                                                                                                                
  - The event document is automatically passed as params.doc                                                                                                                                                     
  - scripted_upsert is always true — the script runs on every write including first create                                                                                                                       
  - Script params support ${} expression syntax for dynamic values from event fields or metadata                                                                                                                 
                                                                                                                                                                                                                 Configuration                                                                                                                                                                                                  
  ```                                                                                                                                                                                                               
  opensearch:                                                                                                                                                                                                    
    hosts: ["https://localhost:9200"]                                                                                                                                                                            
    index: "my-index"                                                                                                                                                                                            
    action: "upsert"                                                                                                                                                                                             
    document_id: "${/id}"                                                                                                                                                                                        
    script:                                                                                                                                                                                                      
      source: "ctx._source.putAll(params.doc); ctx._source.write_count = (ctx._source.containsKey('write_count') ? ctx._source.write_count + 1 : 1)"                                                                                                                                                                                                                                                 
      params:                                                                                                                                                                                                    
        table: "${/source_table}"                                                                                                                                                                                
```                                                                                                                                                                                                                 
Examples                                                                                                                                                                                                       
                                                                                                                                                                                                                 
1. Merge with write tracking                                                                                                                                                                                   
  ```                                                                                                                                                                                                               
  script:                                                                                                                                                                                                        
    source: "ctx._source.putAll(params.doc); ctx._source.write_count = (ctx._source.containsKey('write_count') ? ctx._source.write_count + 1 : 1)"                                                                                                                                                                                                                                                 
 ```                                                                                                                                                                                                                
  First write → {"id": 1, "name": "test", "write_count": 1}                                                                                                                                                      
  Second write → {"id": 1, "name": "updated", "write_count": 2}                                                                                                                                                  
                                                                                                                                                                                                                 2. Conditional price update — only update if lower                                                                                                                                                             
                                                                                                                                                                                                            
```
  script:                                                                                                                                                                                                        
    source: "if (params.doc.price < ctx._source.price) { ctx._source.price = params.doc.price }"                                                                                                                 
```                                                                                                                                                                                        

3. Version-gated full replace    
                                                                                                                                                                          
```                                                                                                                                                                                                                 
  script:                                                                                                                                                                                                        
    source: "if (params.doc.version > ctx._source.version) { ctx._source = params.doc }"                                                                                                                                                                                                                                                                                                                  
```
                                                                                                                                                                                                              
4. Multi-table denormalization with dynamic params 
                                                                                                                                                 
```                                                                                                                                                                                                                 
  script:                                                                                                                                                                                                        
    source: |                                                                                                                                                                                                    
      String[] fields = params.fields.splitOnToken(',');                                                                                                                                                         
      if (params.is_delete == 'true') {                                                                                                                                                                          
        for (String field : fields) { ctx._source.remove(field) }                                                                                                                                                
      } else {                                                                                                                                                                                                   
        for (String field : fields) {                                                                                                                                                                            
          if (params.doc.containsKey(field)) { ctx._source[field] = params.doc[field] }                                                                                                                          
        }                                                                                                                                                                                                        
      }                                                                                                                                                                                                                                                                                                                                                                                                   
    params:                                                                                                                                                                                                      
      table: "${/_table}"                                                                                                                                                                                        
      fields: "${/_fields}"                                                                                                                                                                                      
      is_delete: "${/_is_delete}"                                                                                                                                                                                
```
                                                                                                                                                                                          
Events from products table:                                                                                                                                                                                    
 ```                                                                                                                                                                                                                
  {"primary_key": "prod-1", "title": "Widget A", "_table": "products", "_fields": "title,description", "_is_delete": "false"}                                                                                    
 ```                                                                                                                                                                                                                
Events from pricing table:                                                                                                                                                                                     
 ```                                                                                                                                                                                                                
  {"primary_key": "prod-1", "price": 9.99, "_table": "pricing", "_fields": "price,currency", "_is_delete": "false"}                                                                                              
```                                                                                                                                                                                                                 
Result in OpenSearch:                                                                                                                                                                                          
```                                                                                                                                                                                                                 
  {"title": "Widget A", "price": 9.99}                                                                                                                                                                           
```                                                                                                                                                                                                                 
Delete pricing:                                                                                                                                                                                                
```                                                                                                                                                                                                                 
  {"primary_key": "prod-1", "_table": "pricing", "_fields": "price,currency", "_is_delete": "true"}                                                                                                              
```                                                                                                                                                                                                                 
Result:                                                                                                                                                                                                        
```                                                                                                                                                                                                                 
  {"title": "Widget A"}    
``` 
### Issues Resolved
Resolves #3563
 
### Check List

### DIFF
--- a/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
+++ b/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
@@ -61,7 +61,6 @@ import org.opensearch.dataprepper.plugins.sink.opensearch.bulk.JavaClientAccumul
 import org.opensearch.dataprepper.plugins.sink.opensearch.bulk.JavaClientAccumulatingUncompressedBulkRequest;
 import org.opensearch.dataprepper.plugins.sink.opensearch.bulk.SerializedJson;
 import org.opensearch.dataprepper.plugins.sink.opensearch.configuration.ActionConfiguration;
-import org.opensearch.dataprepper.plugins.sink.opensearch.configuration.ScriptConfiguration;
 import org.opensearch.dataprepper.plugins.sink.opensearch.configuration.DlqConfiguration;
 import org.opensearch.dataprepper.plugins.sink.opensearch.configuration.OpenSearchSinkConfig;
 import org.opensearch.dataprepper.plugins.sink.opensearch.dlq.FailedBulkOperation;
@@ -91,10 +90,8 @@ import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
--- a/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
+++ b/data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
@@ -61,6 +61,7 @@ import org.opensearch.dataprepper.plugins.sink.opensearch.bulk.JavaClientAccumul
 import org.opensearch.dataprepper.plugins.sink.opensearch.bulk.JavaClientAccumulatingUncompressedBulkRequest;
 import org.opensearch.dataprepper.plugins.sink.opensearch.bulk.SerializedJson;
 import org.opensearch.dataprepper.plugins.sink.opensearch.configuration.ActionConfiguration;
+import org.opensearch.dataprepper.plugins.sink.opensearch.configuration.ScriptConfiguration;
 import org.opensearch.dataprepper.plugins.sink.opensearch.configuration.DlqConfiguration;
 import org.opensearch.dataprepper.plugins.sink.opensearch.configuration.OpenSearchSinkConfig;
 import org.opensearch.dataprepper.plugins.sink.opensearch.dlq.FailedBulkOperation;
@@ -90,8 +91,10 @@ import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
@@ -139,6 +142,7 @@ public class OpenSearchSink extends AbstractSink<Record<Event>> {
   private final String pipeline;
   private final String action;
   private final List<ActionConfiguration> actions;
+  private final ScriptManager scriptManager;
   private final String documentRootKey;
   private String configuredIndexAlias;
   private final ReentrantLock lock;
@@ -208,6 +212,7 @@ public class OpenSearchSink extends AbstractSink<Record<Event>> {
     this.routing = openSearchSinkConfig.getIndexConfiguration().getRouting();
     this.action = openSearchSinkConfig.getIndexConfiguration().getAction();
     this.actions = openSearchSinkConfig.getIndexConfiguration().getActions();
+    this.scriptManager = new ScriptManager(openSearchSinkConfig.getIndexConfiguration().getScriptConfiguration(), expressionEvaluator);
     this.documentRootKey = openSearchSinkConfig.getIndexConfiguration().getDocumentRootKey();
     this.versionType = openSearchSinkConfig.getIndexConfiguration().getVersionType();
     this.versionExpression = openSearchSinkConfig.getIndexConfiguration().getVersionExpression();
@@ -342,7 +347,7 @@ public class OpenSearchSink extends AbstractSink<Record<Event>> {
     return initialized;
   }
 
-  private BulkOperation getBulkOperationForAction(final String action,
+  BulkOperation getBulkOperationForAction(final String action,
                                                   final SerializedJson document,
                                                   final Long version,
                                                   final String indexName,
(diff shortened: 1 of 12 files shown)


data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -535,7 +560,19 @@ public void doOutput(final Collection<Record<Event>> records) {
       BulkOperation bulkOperation;
 
       try {
-        bulkOperation = getBulkOperationForAction(eventAction, document, version, indexName, event.getJsonNode());
+        Map<String, Object> resolvedScriptParams = null;
+        if (scriptConfiguration != null && scriptConfiguration.getParams() != null) {
+          resolvedScriptParams = new HashMap<>();
+          for (final Map.Entry<String, Object> entry : scriptConfiguration.getParams().entrySet()) {
+            final Object value = entry.getValue();
+            if (value instanceof String && ((String) value).contains("${")) {
```
dlvenable: I believe we have other instances of this in this sink. Can we consolidate this somehow? Perhaps something like:

````
resovledScriptParams.put(entry.getKey(), eventEvaluator.get(event, value)
```

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -380,18 +391,32 @@ private BulkOperation getBulkOperationForAction(final String action,
...
-                  .upsert(filteredJsonNode)
-                  .versionType(versionType)
-                  .version(version) :
-              new UpdateOperation.Builder<>()
-                  .index(indexName)
-                  .document(filteredJsonNode)
                   .versionType(versionType)
                   .version(version);
+
+          if (scriptConfiguration != null) {
```
dlvenable: Is there a good design that moves this logic out of here and into a class for handling scripting? e.g. `ScriptManager`?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -535,7 +560,19 @@ public void doOutput(final Collection<Record<Event>> records) {
       BulkOperation bulkOperation;
 
       try {
-        bulkOperation = getBulkOperationForAction(eventAction, document, version, indexName, event.getJsonNode());
+        Map<String, Object> resolvedScriptParams = null;
```
dlvenable: Also, can this go into `ScriptManager`?

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -342,11 +352,12 @@ public boolean isReady() {
...
   }
 
-  private BulkOperation getBulkOperationForAction(final String action,
+  BulkOperation getBulkOperationForAction(final String action,
                                                   final SerializedJson document,
                                                   final Long version,
                                                   final String indexName,
-                                                  final JsonNode jsonNode) {
+                                                  final JsonNode jsonNode,
+                                                  final Map<String, Object> resolvedScriptParams) {
```
dlvenable: How about adding the following to `SerializedJson`?

```
Optional<Map<String, Object>> getResolvedScriptParameters();
```

That class should also be updated to use a builder pattern (with some required fields) so that these optionals don't have be added each time.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
```diff
@@ -342,11 +352,12 @@ public boolean isReady() {
     return initialized;
   }
 
-  private BulkOperation getBulkOperationForAction(final String action,
+  BulkOperation getBulkOperationForAction(final String action,
```
dlvenable, line 347: Rather than expose this to test using package private, what would be the effort to move this to its own class - say `BulkOperationFactory`?

dlvenable (changes requested):
Thanks @dinujoh !

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/OpenSearchSink.java
dinujoh: Will refactor `BulkOperationFactory` as a follow-up to keep this PR scoped.

dlvenable (dismissed):

dlvenable (approved):

oeyh (approved):

PR #6746, opened 2026-04-09
Drop Kinesis records with invalid UTF-8 bytes

sb2k16:
Validate UTF-8 encoding of Kinesis record bytes before passing to the codec. Records containing malformed UTF-8 (e.g. unpaired surrogates like 0xDBC8) are dropped with a warning log instead of crashing the pipeline with a Jackson JsonParseException.

### DIFF
--- a/data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/converter/KinesisRecordConverter.java
+++ b/data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/converter/KinesisRecordConverter.java
@@ -10,6 +10,8 @@
 
 package org.opensearch.dataprepper.plugins.kinesis.source.converter;
 
+import io.micrometer.core.instrument.Counter;
+import org.opensearch.dataprepper.metrics.PluginMetrics;
 import org.opensearch.dataprepper.model.codec.DecompressionEngine;
 import org.opensearch.dataprepper.model.codec.InputCodec;
 import org.opensearch.dataprepper.model.event.Event;
@@ -17,6 +19,9 @@ import org.opensearch.dataprepper.model.event.EventMetadata;
 import org.opensearch.dataprepper.model.record.Record;
 import org.opensearch.dataprepper.plugins.kinesis.source.processor.KinesisInputOutputRecord;
 import software.amazon.kinesis.retrieval.KinesisClientRecord;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/converter/KinesisRecordConverter.java
```diff
@@ -63,8 +71,26 @@ private void processRecord(final DecompressionEngine decompressionEngine,
         // Read bytebuffer
         byte[] arr = new byte[record.data().remaining()];
         record.data().get(arr);
-        ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(arr);
 
+        if (!isValidUtf8(arr)) {
```
dlvenable: Is this the right solution? There is no reason that we cannot have codecs that parse non UTF-8 bytes. Should the codec itself do this check? 

I believe that our Avro codec wouldn't work here. Also we could support Protobuf or other formats in the future.

dlvenable (changes requested):

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/converter/KinesisRecordConverter.java
dlvenable: Can we follow this approach?

https://github.com/opensearch-project/data-prepper/blob/638b049fe4490c6253fa7d4961ee2d61ca83793c/data-prepper-plugins/s3-source/src/main/java/org/opensearch/dataprepper/plugins/source/s3/S3ObjectWorker.java#L161-L169

dlvenable (dismissed):
Thanks @sb2k16 !

divbok (approved):

data-prepper-plugins/kinesis-source/src/main/java/org/opensearch/dataprepper/plugins/kinesis/source/converter/KinesisRecordConverter.java
```diff
@@ -65,6 +69,11 @@ private void processRecord(final DecompressionEngine decompressionEngine,
         record.data().get(arr);
         ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(arr);
 
-        codec.parse(decompressionEngine.createInputStream(byteArrayInputStream), eventConsumer);
+        try {
+            codec.parse(decompressionEngine.createInputStream(byteArrayInputStream), eventConsumer);
+        } catch (final Exception e) {
+            LOG.error("Failed to parse Kinesis record. sequenceNumber={}, partitionKey={}",
```
graytaylor0, line 81: Should we have a metric for this?

graytaylor0 (dismissed):

dlvenable (approved):

graytaylor0 (approved):

PR #6747, opened 2026-04-09
Enable HTTP Sink with aws sigv4 auth mode only

kkondaka:
### Description
Enable HTTP Sink with aws sigv4 auth mode only. 
Modified HTTP sink code to use only aws auth mode only.
Removed other auth options
 
### Issues Resolved
 
### Check List


data-prepper-plugins/http-sink/build.gradle
```diff
@@ -28,13 +28,25 @@ dependencies {
...
+    testImplementation project(':data-prepper-test:test-event')
+    testImplementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
 }
 
 test {
     useJUnitPlatform()
 }
 
 sourceSets {
+    main {
```
dlvenable: You already deleted these files so you can remove this change.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/HttpSinkConfiguration.java
```diff
@@ -7,87 +7,26 @@
...
     private String url;
 
-    @JsonProperty("workers")
-    private Integer workers = DEFAULT_WORKERS;
-
     @JsonProperty("codec")
     private PluginModel codec;
 
-    @JsonProperty("http_method")
-    private HTTPMethodOptions httpMethod = HTTPMethodOptions.POST;
```
dlvenable, line 67: This seems worth keeping.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/service/HttpSinkFlushableBuffer.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public SinkFlushResult flush() {
+        if (buffer.isEmpty()) {
+            return null;
+        }
+
+        final HttpSinkSender httpSender = flushContext.getHttpSender();
+        final OutputCodec codec = flushContext.getCodec();
```
dlvenable, line 48: Use `OutputCodec.getWriter()` instead of the legacy approach. This supports threading better.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
```diff
@@ -0,0 +1,173 @@
...
+        return WebClient.builder()
+                .factory(ClientFactory.builder()
+                        .connectTimeout(config.getConnectionTimeout())
+                        .build())
+                .options(ClientOptions.builder().build())
+                .build();
+    }
+
+    public HttpEndPointResponse send(final byte[] payload) {
+        HttpEndPointResponse response = null;
```
dlvenable: Let's also rename this class to `HttpEndpointResponse` since this is a single word - "endpoint."

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
```diff
@@ -0,0 +1,173 @@
...
+                    return new HttpEndPointResponse(config.getUrl(), 0, "Retry interrupted: " + ie.getMessage());
+                }
+            }
+            attempt++;
+        }
+        
+        return response != null ? response : new HttpEndPointResponse(config.getUrl(), 0, "Max retries exceeded");
+    }
+
+    private SdkHttpFullRequest createSdkHttpRequest(final String url, @Nonnull final byte[] payload) {
```
dlvenable: Let's put everything related to AWS into a different class. Maybe give an `AuthenticationDecorator` interface that is more generic. Then you could have an `AwsAuthenticationDecorator`.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/AwsAuthenticationDecorator.java
```diff
@@ -0,0 +1,55 @@
...
+import software.amazon.awssdk.auth.signer.Aws4Signer;
+import software.amazon.awssdk.auth.signer.params.Aws4SignerParams;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.regions.Region;
+
+import javax.annotation.Nonnull;
+import java.net.URI;
+
+class HttpSinkSigV4Signer {
+    private static final String SERVICE_NAME = "execute-api";
```
dlvenable: I'm not sure if this is the right default. This is API Gateway, but why should that be the preference? We should have a configuration in the `aws:` block for `service_name`.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
```diff
@@ -0,0 +1,173 @@
...
+        if (config.getCustomHeaderOptions() != null) {
+            config.getCustomHeaderOptions().forEach((key, values) -> 
+                values.forEach(value -> builder.appendHeader(key, value))
+            );
+        }
+        return builder.build();
+    }
+
+    private HttpRequest buildHttpRequest(final byte[] payload) {
+        SdkHttpFullRequest sdkHttpRequest = createSdkHttpRequest(config.getUrl(), payload);
```
dlvenable: Replace this with some sort of decorator or builder here. For AWS, it should have both of these methods - create the SDK request and sign.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
dlvenable: This could also be common code and shared across projects. Maybe for a follow on PR.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/AwsAuthenticationDecorator.java
```diff
@@ -0,0 +1,55 @@
+/*
```
dlvenable, line 1: Please update all the license headers in this project.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/service/HttpSinkBufferEntry.java
```diff
@@ -0,0 +1,47 @@
...
+import org.opensearch.dataprepper.model.sink.OutputCodecContext;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+public class HttpSinkBufferEntry implements SinkBufferEntry {
+    private final Event event;
+    private final long estimatedSize;
+
+    public HttpSinkBufferEntry(final Event event, final OutputCodec codec, final OutputCodecContext codecContext) throws IOException {
```
dlvenable, line 25: Use the `OutputCodec.Writer` class.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/service/HttpSinkBufferEntry.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.opensearch.dataprepper.common.sink.SinkBufferEntry;
+import org.opensearch.dataprepper.model.codec.OutputCodec;
+import org.opensearch.dataprepper.model.event.Event;
+import org.opensearch.dataprepper.model.sink.OutputCodecContext;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+public class HttpSinkBufferEntry implements SinkBufferEntry {
```
dlvenable, line 21: Can this be a common class? Perhaps put in the common plugins sub-project as `ByteArraySinkBufferEntry`.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/HttpSinkConfiguration.java
```diff
@@ -96,211 +35,57 @@ public class HttpSinkConfiguration {
...
-    private String acmPrivateKeyPassword;
-
-    @JsonProperty("acm_certificate_arn")
-    private String acmCertificateArn;
-
-    @JsonProperty("acm_cert_issue_time_out_millis")
-    private long acmCertIssueTimeOutMillis = DEFAULT_ACM_CERT_ISSUE_TIME_OUT_MILLIS;
-
-    @JsonProperty("insecure")
-    private boolean insecure = DEFAULT_INSECURE;
```
dlvenable, line 126: We should keep this. If `insecure` is true only then allow for `http` schemes. You use this in the IT, so I think we need it now.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/HttpSinkConfiguration.java
```diff
@@ -96,211 +35,57 @@ public class HttpSinkConfiguration {
...
-
-    @JsonProperty("insecure_skip_verify")
```
dlvenable, line 128: Don't we also need this for testing?

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSink.java
```diff
@@ -0,0 +1,97 @@
...
+                codecConfiguration.getPluginSettings());
+        codecPluginSettings.setPipelineName(pipelineDescription.getPipelineName());
+
+        final OutputCodec codec = pluginFactory.loadPlugin(OutputCodec.class, codecPluginSettings);
+        final OutputCodecContext codecContext = OutputCodecContext.fromSinkContext(context);
+
+        final SinkMetrics sinkMetrics = new DefaultSinkMetrics(pluginMetrics, "Event");
+
+        final HttpSinkSender httpSender = new HttpSinkSender(
+                    httpSinkConfiguration.getAwsConfig() != null ? awsCredentialsSupplier : null,
```
dlvenable: Construct some decorator in here and then pass that.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/BasicAuthCredentials.java
```diff
@@ -1,25 +0,0 @@
-/*
- * Copyright OpenSearch Contributors
- * SPDX-License-Identifier: Apache-2.0
- */
-package org.opensearch.dataprepper.plugins.sink.http.configuration;
-
-import com.fasterxml.jackson.annotation.JsonProperty;
-
-public class BasicAuthCredentials {
```
dlvenable, line 9: I think removing a lot of the authentication options is good. But we should at least support basic auth. Otherwise nobody can use this other than with AWS. If we start here, we can look for community contributions for expanded support.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSigV4Signer.java
```diff
@@ -0,0 +1,55 @@
...
+                .withRegion(awsConfig.getAwsRegion())
+                .withStsRoleArn(awsConfig.getAwsStsRoleArn())
+                .withStsExternalId(awsConfig.getAwsStsExternalId())
+                .withStsHeaderOverrides(awsConfig.getAwsStsHeaderOverrides())
+                .build();
+    }
+
+    SdkHttpFullRequest signRequest(final SdkHttpFullRequest unsignedRequest) {
+        if (credentialsProvider == null || credentialsProvider.resolveCredentials() == null) {
+            return null;
```
dlvenable: If we fail to load credentials this will give an NPE, right? We should have a better way to fail.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
```diff
@@ -0,0 +1,173 @@
...
+            }
+            attempt++;
+        }
+        
+        return response != null ? response : new HttpEndPointResponse(config.getUrl(), 0, "Max retries exceeded");
+    }
+
+    private SdkHttpFullRequest createSdkHttpRequest(final String url, @Nonnull final byte[] payload) {
+        final SdkHttpFullRequest.Builder builder = SdkHttpFullRequest.builder()
+                .method(SdkHttpMethod.POST)
```
dlvenable: This should be configurable. It is an easy config to offer.

Also, this and the URI should be provided by the `HttpSinkSender` and then used by the suggested decorator. That is, don't mix the concerns of determining the method/URI with the concern of signing.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
```diff
@@ -0,0 +1,173 @@
...
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.http.SdkHttpMethod;
+
+import javax.annotation.Nonnull;
+import java.net.URI;
+import java.util.Set;
+
+public class HttpSinkSender {
+    private static final Logger LOG = LoggerFactory.getLogger(HttpSinkSender.class);
+    private static final String HTTP_METHOD = "post";
```
dlvenable: I think this is unused.

dlvenable (changes requested):
Nice improvement @kkondaka . It will be good to get this sink ready to use! I have a number of suggestions. Overall we should be sure this is a generic sink and not so tightly coupled with AWS/SigV4.

One alternative is to create an `aws_http` sink. But I think some code clean up will make that unnecessary.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/BasicAuthCredentials.java
dlvenable: We should also support unauthenticated requests.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
```diff
@@ -0,0 +1,173 @@
...
+                .method(SdkHttpMethod.POST)
+                .uri(URI.create(url))
+                .contentStreamProvider(() -> SdkBytes.fromByteArray(payload).asInputStream());
+
+        if (signer != null) {
+            builder.putHeader("x-amz-content-sha256", "required");
+        }
+
+
+        if (config.getCustomHeaderOptions() != null) {
```
dlvenable: I think the Codecs should ideally offer a MIME type interface.

```
Optional<String> getMimeType();
```

Then the codec can provide a default MIME type to use in the `Content-Type` header.

This would work well for the `json` codec.

However it might not work for all. I think ND-JSON doesn't have a MIME type. Maybe `application/jsonl` or `application/x-ndjson`. We should still override it with custom headers.

dlvenable (changes requested):

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/HttpSinkConfiguration.java
kkondaka: Agreed. WIll add them back.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/HttpSinkConfiguration.java
kkondaka: I thought about it but since we are not using now, decided to keep it out. We can always add things that are needed back when they are needed.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/AwsAuthenticationDecorator.java
kkondaka: "execute-api" is correct for aws services and this is aws specific code.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/service/HttpSinkBufferEntry.java
kkondaka: I will move it to common sink framework code we already in a follow-up PR.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/BasicAuthCredentials.java
kkondaka: Requests are unauthenticated when "aws:" is not used. There is already integration tests that shows that path works.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
kkondaka: Agreed. But do not need it initially. I will do it in a follow-up PR

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
kkondaka: Let's look at this in a follow-up PR

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/service/HttpSinkBufferEntry.java
```diff
@@ -0,0 +1,47 @@
...
+    private final Event event;
+    private final long estimatedSize;
+
+    public HttpSinkBufferEntry(final Event event, final OutputCodec codec, final OutputCodecContext codecContext) throws IOException {
+        this.event = event;
+        this.estimatedSize = calculateSize(event, codec, codecContext);
+    }
+
+    private long calculateSize(final Event event, final OutputCodec codec, final OutputCodecContext codecContext) throws IOException {
```
srikanthpadakanti, line 30: The event is serialized once in calculateSize() for size estimation, then again in flush() for actual sending. Consider caching the serialized bytes from the size calculation and reusing them during flush, or using a cheaper estimation like event.toJsonString().length()

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/service/HttpSinkBufferEntry.java
dlvenable: A follow-on PR makes sense.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/AwsAuthenticationDecorator.java
dlvenable: The `execute-api` is not a generic AWS services service. It is quite specific to API Gateway with `AWS_IAM` authentication. I think we should require this to be set by the user.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
dlvenable: Yea, that's good.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSinkSender.java
dlvenable: I think we can add additional configurations later. But we shouldn't leave the code in a state where two things that don't need to be connected are. Specifically having the AWS SDK code in `HttpSinkSender`. Let's just decouple that and we should be good to go.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/service/HttpSinkBufferEntry.java
kkondaka: Yes, this is a known issue. We cannot use `event.toJsonString().length()` because the codec is not always json.

dlvenable (commented):
Two things before merging:

* Let's require the `service_name` because this is a generic `http` sink that should be able to send to any AWS SigV4 service. It isn't just for API Gateway.
* Let's remove the AWS code from `HttpSinkSender`. This way if the community wants to extend the authentication options there is no need to extract this AWS code which isn't tested in an integration environment. This will give more confidence.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/service/HttpSinkBufferEntry.java
kkondaka: One way to improve this is to store the OutputStream from calculateSize() and reuse it in flush(), not sure if that would work. But something to look into in future.

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/HttpSinkConfiguration.java
```diff
@@ -58,6 +58,9 @@ public class HttpSinkConfiguration {
     @JsonProperty("connection_timeout")
     private Duration connectionTimeout = DEFAULT_CONNECTION_TIMEOUT;
 
+    @JsonProperty("aws_sigv4_service_name")
```
dlvenable, line 61: Do we have an `aws: ` block? If so, I think we can just make this `service_name` under that block. We can follow on with this change or discuss further before the PR.

dlvenable (approved):

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/configuration/HttpSinkConfiguration.java
graytaylor0: There is an aws block. Would be better to move this under that

data-prepper-plugins/http-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/http/HttpSink.java
```diff
@@ -0,0 +1,105 @@
...
+                httpSinkConfiguration,
+                sinkMetrics,
+                httpSender,
+                pipelineDescription,
+                codec,
+                codecContext);
+    }
+
+    @Override
+    public boolean isReady() {
```
graytaylor0, line 88: Doesn't have to be in this PR but would be good to consider implementing graceful shutdown where we flush everything in the sink before exiting.

graytaylor0 (approved):

PR #6750, opened 2026-04-10
Add support for zero dates in mysql mapping

divbok:
### Description
Add support for MySQL zero-in-date values in `TemporalTypeHandler`.

MySQL allows dates with zero components (e.g., 1999-00-20, 2024-05-00, 0000-01-12) when `NO_ZERO_IN_DATE` SQL mode is not enabled. Currently, TemporalTypeHandler only handles the full zero date `0000-00-00` and throws IllegalArgumentException for partial zero dates, causing the entire Parquet file to fail during RDS S3 export ingestion.

This change:

Extends handleDate() to return null for any date with a zero year (0000), zero month (-00-), or zero day (-00)
Extends handleDateTime() to apply the zero-component check on the date portion for DATETIME values
since TIMESTAMP only permits the full zero timestamp (0000-00-00 00:00:00), not partial zeros. 
 
 Reference: https://dev.mysql.com/doc/refman/9.6/en/datetime.html
 
### Check List


dlvenable (approved):
Nice, thanks!

dinujoh (approved):

PR #6757, opened 2026-04-14
Add JDBC-based source coordination store plugin

### LINKED ISSUE #6740: Add JDBC-based source coordination store

The source coordination store currently has only two implementations: `dynamodb` for production multi-node environments and `in_memory` for single-node development use. Running Data Prepper in a multi-node cluster requires Amazon DynamoDB, so there is no cloud-agnostic option for the coordination store backend.

A new `jdbc` source coordination store plugin backed by a relational database. The coordination store's requirements (key-value CRUD with optimistic locking, status-based queries with priority ordering, and TTL for completed partitions) map naturally to standard SQL operations.

| DynamoDB concept | JDBC equivalent |
|---|---|
| Partition key + sort key | Composite primary key (`source_identifier`, `source_partition_key`) |
| Conditional put (`attribute_not_exists`) | Conditional insert (e.g. `INSERT ... WHERE NOT EXISTS`) |
| Conditional update (version check) | `UPDATE ... WHERE version = ?` |
| GSI query by status + priority | `SELECT ... WHERE source_status = ? ORDER BY partition_priority` with an index |
| TTL (automatic expiration) | TTL-based cleanup of completed partitions is supported via the `ttl` configuration option. When configured, expired partitions are periodically deleted. |

Configuration example:
```yaml
source_coordination:
 store:
   jdbc:
     url: "jdbc:postgresql://localhost:5432/dataprepper"
     username: "dp_user"
     password: "..."
```

The initial implementation will be tested with PostgreSQL and MySQL.

Redis: Offers native TTL support and lightweight operation, but does not guarantee durability by default. The coordination store requires that a successful write is durable, since lost partition state updates could cause duplicate processing or missed partitions. RDBMS provides this guarantee by default.

**Additional context**

The original source coordination design (#2412) noted that the store should be pluggable and mentioned several potential backends including "Remote/Local File DB, Apache Zookeeper, MySQL, DynamoDB, and more." The `SourceCoordinationStore` interface is already designed as a plugin, so adding a new implementation requires no changes to the existing framework or DynamoDB plugin.

I am willing to implement this.

lawofcycles:
### Description
Adds a new `jdbc` source coordination store plugin backed by a relational database.

The plugin uses standard SQL operations to implement the `SourceCoordinationStore` interface.

| Capability | Implementation |
|---|---|
| Conditional create | `INSERT` with duplicate key check |
| Optimistic locking | `UPDATE ... WHERE version = ?` |
| Partition acquisition | Status-based query with priority ordering |
| TTL cleanup | Scheduled deletion of expired `COMPLETED` partitions |

Configuration example
```yaml
source_coordination:
 store:
   jdbc:
     url: "jdbc:postgresql://localhost:5432/dataprepper"
     username: "dp_user"
     password: "dp_pass"
```

Tested with PostgreSQL and MySQL. Integration tests run against PostgreSQL by default and can be configured for MySQL via system properties. Also verified end-to-end with OpenSearch Source and Iceberg Source pipelines on both databases, including TTL-based cleanup of completed partitions.

### Issues Resolved
Resolves #6740

### Check List


data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
```diff
@@ -0,0 +1,452 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
```
srikanthpadakanti, line 9: No unit tests for `JdbcSourceCoordinationStore` itself (only integration tests). The DynamoDB store has unit tests with mocked clients. Add unit tests that mock DataSource/Connection to test error handling paths, version mismatch behavior, and SQL exception translation without requiring a running database.

data-prepper-plugins/jdbc-source-coordination-store/src/integrationTest/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStoreIT.java
```diff
@@ -0,0 +1,236 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.Statement;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
```
srikanthpadakanti: Uses JUnit assertions (assertEquals, assertFalse, assertTrue). Project standard requires Hamcrest matchers (assertThat with equalTo, is(true), is(false)).

data-prepper-plugins/jdbc-source-coordination-store/src/integrationTest/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStoreIT.java
```diff
@@ -0,0 +1,236 @@
...
+                SOURCE_ID, "node-1", Duration.ofMinutes(10)).isPresent());
+    }
+
+    @Test
+    void tryAcquireAvailablePartition_acquires_expired_assigned() throws InterruptedException {
+        store.tryCreatePartitionItem(SOURCE_ID, "p1", SourcePartitionStatus.UNASSIGNED, 0L, null, false);
+
+        final Optional<SourcePartitionStoreItem> first = store.tryAcquireAvailablePartition(
+                SOURCE_ID, "node-1", Duration.ofSeconds(2));
+        assertTrue(first.isPresent());
```
srikanthpadakanti: Thread.sleep(3000) in test. Use `org.awaitility:awaitility` for polling-based assertions instead of fixed sleeps. The DynamoDB store tests already use `awaitility` as a test dependency.

data-prepper-plugins/jdbc-source-coordination-store/src/test/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcPartitionItemTest.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package org.opensearch.dataprepper.plugins.sourcecoordinator.jdbc;
+
+import org.junit.jupiter.api.Test;
+import org.opensearch.dataprepper.model.source.coordinator.SourcePartitionStatus;
+
+import java.time.Instant;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
srikanthpadakanti: uses JUnit assertions instead of Hamcrest matchers

data-prepper-plugins/jdbc-source-coordination-store/src/test/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcStoreSettingsTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+package org.opensearch.dataprepper.plugins.sourcecoordinator.jdbc;
+
+import org.junit.jupiter.api.Test;
+
+import java.time.Duration;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
```
srikanthpadakanti: uses JUnit assertions instead of Hamcrest matchers

data-prepper-plugins/jdbc-source-coordination-store/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+dependencies {
+    implementation project(':data-prepper-api')
+    implementation 'com.zaxxer:HikariCP:5.1.0'
+    implementation 'com.fasterxml.jackson.core:jackson-databind'
+    implementation 'javax.inject:javax.inject:1'
+    implementation 'org.postgresql:postgresql:42.7.7'
```
srikanthpadakanti: Any reason why both postgresql and mysql included as implementation dependencies? 
As they should be runtimeOnly. No compile-time references to driver-specific classes exist in the source as they're loaded via DriverManager at runtime.

data-prepper-plugins/jdbc-source-coordination-store/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+    implementation 'com.mysql:mysql-connector-j:8.4.0'
+}
+
```
srikanthpadakanti, line 19: No JaCoCo configuration. New plugins should have jacocoTestCoverageVerification with minimum 1.0 ratio per project standards. 
@dlvenable correct me if I am wrong.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcStoreSettings.java
```diff
@@ -0,0 +1,88 @@
...
+    private final String tableName;
+    private final boolean skipTableCreation;
+    private final int maxPoolSize;
+    private final Duration ttl;
+    private final Map<String, String> connectionProperties;
+
+    @JsonCreator
+    public JdbcStoreSettings(
+            @JsonProperty("url") final String url,
+            @JsonProperty("username") final String username,
```
srikanthpadakanti, line 36: password is stored as plain String. Consider documenting that this field supports Data-Prepper's secret management integration, or note it as a known limitation for the initial version

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
```diff
@@ -0,0 +1,452 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(JdbcSourceCoordinationStore.class);
+    private static final String UNIQUE_VIOLATION_SQLSTATE = "23505";
+    private static final String DUPLICATE_RELATION_SQLSTATE = "42P07";
+    private static final int MYSQL_DUPLICATE_KEY_NAME_ERROR = 1061;
+    private static final Duration TTL_CLEANUP_INTERVAL = Duration.ofHours(1);
+
+    private final JdbcStoreSettings settings;
+    private HikariDataSource dataSource;
+    private ScheduledExecutorService ttlExecutor;
```
srikanthpadakanti, line 56: `HikariDataSource` and `ScheduledExecutorService` are never shut down. Implement `AutoCloseable` with a `close()` that calls `dataSource.close()` and `ttlExecutor.shutdownNow()`.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
```diff
@@ -0,0 +1,452 @@
...
+
+    @DataPrepperPluginConstructor
+    public JdbcSourceCoordinationStore(final JdbcStoreSettings settings) {
+        this.settings = settings;
+    }
+
+    @Override
+    public void initializeStore() {
```
srikanthpadakanti, line 64: `initializeStore()` is not guarded against concurrent calls. If called twice, it creates a second HikariDataSource without closing the first. Add an AtomicBoolean initialized guard or null check on dataSource.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
```diff
@@ -0,0 +1,452 @@
...
+        hikariConfig.setDataSourceProperties(props);
+
+        this.dataSource = new HikariDataSource(hikariConfig);
+
+        if (!settings.skipTableCreation()) {
+            createTable();
+        }
+
+        if (settings.getTtl() != null) {
+            ttlExecutor = Executors.newSingleThreadScheduledExecutor(r -> {
```
srikanthpadakanti, line 88: The `ScheduledExecutorService` for TTL cleanup is never shut down. On pipeline stop this daemon thread keeps running. Store a reference and shut it down in the `close()` method.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
lawofcycles: Addressed. Implemented `AutoCloseable` with `close()` that shuts down both `ttlExecutor` and `dataSource`.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
lawofcycles: Addressed. Added `AtomicBoolean` guard in `initializeStore()` to prevent double initialization.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcStoreSettings.java
lawofcycles: The `password` field already supports Data Prepper's secret management through `VariableExpander`. Users can use `${{aws_secrets:secretId:secretKey}}` syntax, which is automatically resolved during deserialization via `@JsonProperty`. 
However, there is currently only `AwsSecretsPluginConfigValueTranslator`, so it would be worth adding another implementation such as HashiCorp Vault for cloud-agnostic use cases in the future.

data-prepper-plugins/jdbc-source-coordination-store/build.gradle
lawofcycles: addressed.

data-prepper-plugins/jdbc-source-coordination-store/src/integrationTest/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStoreIT.java
lawofcycles: Addressed

data-prepper-plugins/jdbc-source-coordination-store/src/integrationTest/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStoreIT.java
lawofcycles: Addressed

data-prepper-plugins/jdbc-source-coordination-store/src/test/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcPartitionItemTest.java
lawofcycles: Addressed

data-prepper-plugins/jdbc-source-coordination-store/src/test/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcStoreSettingsTest.java
lawofcycles: Addressed

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
lawofcycles: Addressed

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
lawofcycles: Added `JdbcSourceCoordinationStoreTest`.

data-prepper-plugins/jdbc-source-coordination-store/build.gradle
lawofcycles: Added `jacocoTestCoverageVerification` with minimum 0.99. The threshold varies across plugins (from 0.84 to 1.0), and the current instruction coverage is 99%. The only uncovered line is the `HikariDataSource` constructor in `createDataSource()`, which is overridden in tests to inject a mock.

lawofcycles:
Thank you for the thorough review, @srikanthpadakanti. All comments have been addressed.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    HikariDataSource createDataSource(final HikariConfig hikariConfig) {
+        return new HikariDataSource(hikariConfig);
+    }
+
+    private void createTable() {
+        final String tableName = settings.getTableName();
+        final String createTableSql = "CREATE TABLE IF NOT EXISTS " + tableName + " ("
+                + "source_identifier VARCHAR(256) NOT NULL, "
```
srikanthpadakanti, line 105: `source_identifier` and `source_partition_key` are `VARCHAR(256)`. S3 object keys can be up to 1024 characters and some source plugins may use them as partition keys. Consider increasing to `VARCHAR(1024)` or documenting the limitation.

srikanthpadakanti:
> Thank you for the thorough review, @srikanthpadakanti. All comments have been addressed.

Thanks for addressing them. Added one more comment. Everything else looks good to me.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
lawofcycles: Increased `source_partition_key` from `VARCHAR(256)` to `VARCHAR(512)`. MySQL InnoDB with utf8mb4 has a 3072 byte composite primary key limit (`N * 4` bytes per `VARCHAR(N)` column), so `VARCHAR(1024)` is not possible. `VARCHAR(256) + VARCHAR(512)` fits exactly at the 3072 byte limit. Will note this limitation on the document.

srikanthpadakanti (approved):
Thanks for your contribution @lawofcycles

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcStoreSettings.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public String getUsername() {
+        return username;
+    }
+
+    public String getPassword() {
+        return password;
+    }
+
+    public String getTableName() {
```
dlvenable, line 69: You are using this below in string concatenation. To prevent SQL injection, we need regex validation on the table name.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcStoreSettings.java
```diff
@@ -0,0 +1,88 @@
...
+    public JdbcStoreSettings(
+            @JsonProperty("url") final String url,
+            @JsonProperty("username") final String username,
+            @JsonProperty("password") final String password,
+            @JsonProperty("table_name") final String tableName,
+            @JsonProperty("skip_table_creation") final Boolean skipTableCreation,
+            @JsonProperty("max_pool_size") final Integer maxPoolSize,
+            @JsonProperty("ttl") final Duration ttl,
+            @JsonProperty("connection_properties") final Map<String, String> connectionProperties) {
+        Objects.requireNonNull(url, "url is required for JDBC store settings");
```
dlvenable: We should use JSR bean validations for plugin configurations.

Is there a reason you took the `@JsonCreator` approach here rather than the approach used by other configs?

See the `JsonInputCodecConfig` for an example. It has private fields which are set and annotated. And has getter methods.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
```diff
@@ -0,0 +1,473 @@
...
+        }
+    }
+
+    HikariDataSource createDataSource(final HikariConfig hikariConfig) {
+        return new HikariDataSource(hikariConfig);
+    }
+
+    private void createTable() {
+        final String tableName = settings.getTableName();
+        final String createTableSql = "CREATE TABLE IF NOT EXISTS " + tableName + " ("
```
dlvenable: I think it would be ideal to put this into the resources as a file. It would make it more readable.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
dlvenable: Maybe the key should be more decomposable. But that might be for a future PR.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
```diff
@@ -0,0 +1,473 @@
...
+                + " SET partition_owner = ?, partition_progress_state = ?,"
+                + " source_partition_status = ?, partition_ownership_timeout = ?,"
+                + " reopen_at = ?, closed_count = ?, partition_priority = ?,"
+                + " version = version + 1, expiration_time = ?"
+                + " WHERE source_identifier = ? AND source_partition_key = ? AND version = ?";
+
+        try (Connection conn = dataSource.getConnection();
+             PreparedStatement stmt = conn.prepareStatement(sql)) {
+            int idx = 1;
+            stmt.setString(idx++, item.getPartitionOwner());
```
dlvenable: I tend to think that setting the indices here is clearer.

I also think it might be nice to include Spring and use `NamedParameterJdbcTemplate` to make the code more readable. I'm unsure what packages that brings in. But Data Prepper already uses Spring core and web client.

data-prepper-plugins/jdbc-source-coordination-store/docker/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+services:
```
dlvenable, line 8: We should have a new GitHub Action to run the Docker compose and the integration tests. Kafka and OpenSearch sinks have examples of this approach.

dlvenable (changes requested):
Thank you @lawofcycles for this contribution!

graytaylor0 (commented):
I see you manually tested opensearch source and iceberg with this store, which is great. Please also test with DynamoDB as a source and S3 Scan as a source to make sure there are no weird issues there. Thanks!

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcStoreSettings.java
lawofcycles: No particular reason. Rewrote it to the annotated-fields style with JSR validations, following JsonInputCodecConfig.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcStoreSettings.java
lawofcycles: Added Pattern validation: ([A-Za-z_][A-Za-z0-9_]*\.)?[A-Za-z_][A-Za-z0-9_]* (identifier characters only, one optional schema qualifier).

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
lawofcycles: Moved the CREATE TABLE and CREATE INDEX statements to SQL files under resources.

data-prepper-plugins/jdbc-source-coordination-store/src/main/java/org/opensearch/dataprepper/plugins/sourcecoordinator/jdbc/JdbcSourceCoordinationStore.java
lawofcycles: Switched to explicit indices. On NamedParameterJdbcTemplate: it would pull in spring-jdbc and spring-tx. Given this plugin is a single table with 8 static statements, I lean toward staying with plain JDBC for now, but happy to do the Spring migration as a follow-up if you prefer.

data-prepper-plugins/jdbc-source-coordination-store/docker/docker-compose.yml
lawofcycles: Added a workflow that runs the integration tests against both PostgreSQL and MySQL via the docker-compose file, with test result artifacts uploaded for diagnosis.

lawofcycles:
https://github.com/opensearch-project/data-prepper/pull/6757#discussion_r3190166231

Filed https://github.com/opensearch-project/data-prepper/issues/6982 to track this as a follow-up. The main design question there is who owns the key format

lawofcycles:
@graytaylor0
> I see you manually tested opensearch source and iceberg with this store, which is great. Please also test with DynamoDB as a source and S3 Scan as a source to make sure there are no weird issues there. Thanks!

Tested both, and this was a good call.
S3 Scan worked as is: 20 objects processed exactly once across two concurrent Data Prepper instances sharing the store (a clean 20/20 split, 40 unique records, no duplicates), with all object partitions reaching COMPLETED.

DynamoDB found a real bug. The enhanced source coordinator creates global state items with a null status (they are not for lease), and this store threw an NPE on them and declared the status column NOT NULL, so the dynamodb source's leader loop could not even start. 

Fixed in [4ecde3653](https://github.com/opensearch-project/data-prepper/pull/6757/changes/4ecde36533649f39a3544129e57459f0c3faa2a2): the column is now nullable and the store handles null status on create, read, and update, with unit and integration tests added. After the fix, export (5 items) and stream (3 items) both flowed through, with EXPORT and DATAFILE partitions completing and the GLOBAL, STREAM, and LEADER items managed as expected.

Since the fix touches paths shared by all sources, I re-verified the others against it end to end: opensearch source (30 documents, index partition lifecycle correct), iceberg source (initial load plus an appended snapshot picked up incrementally, each record delivered exactly once, and its null-status GlobalState completion trackers stored correctly, so that source was affected by the same bug), the S3 Scan concurrency test again, and the PostgreSQL and MySQL integration suites. All green.

lawofcycles:
@dlvenable @graytaylor0 Sorry for the delay. I've addressed all the review feedback. Could you take another look when you get a chance?

kkondaka (approved):

kkondaka:
@dlvenable please take a look

dlvenable (approved):
Thank you @lawofcycles for this contribution!

PR #6762, opened 2026-04-16
Support denormalized document joins for RDS source

dinujoh:
### Description
Adds support for joining multiple RDS tables into a single denormalized OpenSearch document during ingestion using scripted upserts. Customers with normalized relational schemas can produce search-optimized documents without query-time joins.

Configuration Example                                                                                                                                                                                                                                                                                                                                                 
```                                                                                                                                                                                                                                                                                                                                                                   
  source:                                                                                                                                                                                                                                                                                                                                                               
    rds:                                                                                                                                                                                                                                                                                                                                                                
      joins:                                                                                                                                                                                                                                                                                                                                                            
        version_field: "__versions"                                                                                                                                                                                                                                                                                                                                     
        relations:                                                                                                                                                                                                                                                                                                                                                      
          - parent: "orders"                                                                                                                                                                                                                                                                                                                                            
            child: "order_items"                                                                                                                                                                                                                                                                                                                                        
            parent_key: "order_id"           # single key (backward compatible)                                                                                                                                                                                                                                                                                         
            child_key: "order_id"                                                                                                                                                                                                                                                                                                                                       
            child_primary_key: "item_id"                                                                                                                                                                                                                                                                                                                                
            join_type: "one_to_many"                                                                                                                                                                                                                                                                                                                                    
            max_child_records: 1000                                                                                                                                                                                                                                                                                                                                     
          - parent: "orders"                                                                                                                                                                                                                                                                                                                                            
            child: "shipping"                                                                                                                                                                                                                                                                                                                                           
            parent_key: "order_id"                                                                                                                                                                                                                                                                                                                                      
            child_key: "order_id"                                                                                                                                                                                                                                                                                                                                       
            child_primary_key: "shipping_id"                                                                                                                                                                                                                                                                                                                            
            join_type: "one_to_one"                                                                                                                                                                                                                                                                                                                                     
          - parent: "product_catalog"                                                                                                                                                                                                                                                                                                                                   
            child: "product_reviews"                                                                                                                                                                                                                                                                                                                                    
            parent_key: ["category", "product_code"]  # composite key                                                                                                                                                                                                                                                                                                   
            child_key: ["category", "product_code"]                                                                                                                                                                                                                                                                                                                     
            child_primary_key: "review_id"                                                                                                                                                                                                                                                                                                                              
```                                                                                           
 
### Issues Resolved
Resolves #[6759]
 
### Check List


data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/stream/BinlogEventListener.java
```diff
@@ -319,6 +337,23 @@ void handleUpdateEvent(com.github.shyiko.mysql.binlog.event.Event event) {
                     break;
                 }
             }
+
+            // Detect FK (join key) change — emit DELETE from old parent doc
+            if (childKeyIndices != null) {
+                boolean fkChanged = false;
+                for (int idx : childKeyIndices) {
+                    if (idx >= 0 && !row.getKey()[idx].equals(row.getValue()[idx])) {
```
oeyh, line 345: FK can be NULL before or after the update, causing NPE. Can use `!Objects.equals(row.getKey()[idx], row.getValue()[idx])` instead.

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/configuration/JoinRelation.java
```diff
@@ -0,0 +1,52 @@
...
+    @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
+    private List<String> childKey;
+
+    @NotEmpty
+    @JsonProperty("child_primary_key")
+    @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
+    private List<String> childPrimaryKey;
+
+    @JsonProperty("join_type")
+    private String joinType = "one_to_many";
```
oeyh, line 48: This can be an Enum field.

data-prepper-pipeline-parser/src/main/java/org/opensearch/dataprepper/pipeline/parser/rule/RuleEvaluator.java
```diff
@@ -91,10 +91,23 @@ private RuleFileEvaluation evaluate(String pipelinesJson) {
         try {
             Collection<RuleStream> ruleStreams = transformersFactory.loadRules();
 
-            //walk through all rules and return first valid
+            // Pre-parse all rules and sort by specificity (most conditions first)
+            // so more specific rules like rds-joins match before generic rds
+            List<ParsedRule> parsedRules = new java.util.ArrayList<>();
```
oeyh, line 96: Add import for ArrayList.

data-prepper-plugins/rds-source/src/main/resources/org/opensearch/dataprepper/transforms/templates/rds-joins-template.yaml
```diff
@@ -0,0 +1,107 @@
...
...ms.child_table_name; String[] pkNames = params.child_pk_name.splitOnToken('|'); String pkVal = params.child_pk_value; String rowKey = table + ':' + pkVal; long storedVersion = ctx._source[vf].containsKey(rowKey) ? ctx._source[vf][rowKey] : -1L; if (incomingVersion <= storedVersion) { ctx.op = 'none'; return } if (!ctx._source.containsKey(arr)) { ctx._source[arr] = [] } ctx._source[arr].removeIf(e -> { String v = ''; for (int i = 0; i < pkNames.length; i++) { if (i > 0) v += '|'; v += e[pkNames[i]].toString() } return v == pkVal }); if (params.is_delete != 'true') { def entry = [:]; for (String f : fields) { if (params.doc.containsKey(f)) { entry[f] = params.doc[f] } } for (String pk : pkNames) { entry[pk] = params.doc[pk] } ctx._source[arr].add(entry); ctx._source[vf][rowKey] = incomingVersion; int max = Integer.parseInt(params.max_child_records); if (ctx._source[arr].size() > max) { def removed = ctx._source[arr].remove(0); String rv = ''; for (int i = 0; i < pkNames.length; i++) { if (i > 0) rv += '|'; rv += removed[pkNames[i]].toString() } ctx._source[vf].remove(table + ':' + rv) } } else { ctx._source[vf].remove(rowKey); if (ctx._source[vf].isEmpty()) { ctx.op = 'delete' } } }"
```
oeyh, line 94: > String rowKey = table + ':' + pkVal;

Why do we need to create an entry in the version map for each `table:pkVal`? I thought table name alone should be sufficient.

dinujoh: Per-table versioning fails for 1:N child tables because multiple child rows from the same table can share the same CDC timestamp. For example:                                                                                                                                                                                                                        
```                                                                                                                                                                                                                                                                                                                                                                        
  Event 1: item_id=14 INSERT (v=1000) → __versions.order_items = 1000  ✅                                                                                                                                                                                                                                                                                               
  Event 2: item_id=15 INSERT (v=1000) → 1000 <= 1000                   ❌ DISCARDED                                                                                                                                                                                                                                                                                     
```                                                                                                                                                                                                                                                                                                                                                                        
Both items are from order_items with the same timestamp (same binlog second or same export snapshot). With per-table versioning, the second item is discarded as stale because incomingVersion <= storedVersion.                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                        
Per-row versioning (__versions["order_items:14"], __versions["order_items:15"]) gives each child row its own independent version history, so they don't interfere with each other:                                                                                                                                                                                    
  
```                                                                                                                                                                                                                                                                                                                                                                      
  Event 1: item_id=14 INSERT (v=1000) → __versions["order_items:14"] = 1000  ✅                                                                                                                                                                                                                                                                                         
  Event 2: item_id=15 INSERT (v=1000) → __versions["order_items:15"] = 1000  ✅ (different key)                                                                                                                                                                                                                                                                         
```                                                                                                                                                                                                                                                                                                                                                                        
Per-table versioning is still used for parent and 1:1 child tables since they overwrite the same fields — only one version per table is needed.

oeyh (dismissed):

oeyh:
One question is that for export, all records will have the same version (snapshot timestamp) and do we need to take additional measures to avoid dropping records from version conflict, for joins with one-to-N relationship?

dinujoh:
> One question is that for export, all records will have the same version (snapshot timestamp) and do we need to take additional measures to avoid dropping records from version conflict, for joins with one-to-N relationship?

Added support for monotonic counter in `ExportRecordConverter.getVersionNumber()` ensures each export record gets a unique version

data-prepper-pipeline-parser/src/main/java/org/opensearch/dataprepper/pipeline/parser/transformer/DynamicConfigTransformer.java
```diff
@@ -82,6 +82,12 @@ public class DynamicConfigTransformer implements PipelineConfigurationTransforme
 
     private static final String S3_BUFFER_PREFIX = "/buffer";
 
+    /**
+     * Pattern to match overlay directives like "<<overlay sink[*].opensearch>>"
+     * The captured group is the target path (e.g., "sink[*].opensearch")
+     */
+    private final Pattern OVERLAY_PATTERN = Pattern.compile("^<<overlay\\s+(.+?)>>$");
```
divbok: Nit: This can be static final

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkOperationFactory.java
```diff
@@ -117,6 +117,7 @@ private BulkOperation createUpdateOperation(final String action,
             if (isUpsert) {
                 builder.upsert(filteredJsonNode);
                 builder.scriptedUpsert(true);
+                builder.retryOnConflict(3);
```
divbok: Should we introduce a configurable flag for setting the number of retries?

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/configuration/JoinRelation.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @NotEmpty
+    @JsonProperty("parent_key")
+    @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
+    private List<String> parentKey;
+
+    @NotEmpty
+    @JsonProperty("child_key")
+    @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
+    private List<String> childKey;
```
divbok, line 41: We should validate that parentKey and childKey are of equal length.

data-prepper-plugins/opensearch/src/main/java/org/opensearch/dataprepper/plugins/sink/opensearch/BulkOperationFactory.java
dinujoh: Lets keep it hardcoded at 3 for now. It's the OpenSearch default, works for our use case, and avoids exposing an advanced tuning knob that most users won't understand. We can make it configurable later if we hit contention issues

divbok (approved):

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/export/DataFileLoader.java
```diff
@@ -133,7 +133,7 @@ public void run() {
                     transformEvent(event, fullTableName, EngineType.fromString(progressState.getEngineType()));
 
                     final long snapshotTime = progressState.getSnapshotTime();
-                    final long eventVersionNumber = snapshotTime - VERSION_OVERLAP_TIME_FOR_EXPORT.toMillis();
+                    final long eventVersionNumber = recordConverter.getVersionNumber(snapshotTime - VERSION_OVERLAP_TIME_FOR_EXPORT.toMillis());
```
oeyh, line 136: Given 5min (300k ms) VERSION_OVERLAP_TIME_FOR_EXPORT, we will have a cap of 300 million export events before export event version meets or exceeds the version for stream events generated after snapshot. 

This is probably OK for most use cases and we can further increase overlap to increase the cap. But we may want to document this implicit cap here.

oeyh (approved):

data-prepper-plugins/rds-source/src/main/java/org/opensearch/dataprepper/plugins/source/rds/export/DataFileLoader.java
dinujoh: I will add a comment as followup.

PR #6768, opened 2026-04-20
feat(otlp-sink): Add support for metrics and logs with configurable SigV4 and headers

### LINKED ISSUE #6763: [FEATURE] Add log signal support, configurable SigV4 signing, and additional headers to OTLP sink
### Is your feature request related to a problem?

The OTLP sink currently only supports sending traces (`ExportTraceServiceRequest`) to AWS X-Ray. It has a hardcoded SigV4 signing service name (`xray`), no way to add custom HTTP headers, and derives the AWS region solely from the endpoint URL. This limits the sink to a single use case when the OTLP/HTTP protocol supports traces, logs, and metrics with various backends.

### What solution would you like?

Extend the existing OTLP sink with:

1. **`signal_type` config option** (`traces` or `logs`) — controls whether the sink builds `ExportTraceServiceRequest` or `ExportLogsServiceRequest`. Default: `traces` for backward compatibility.
2. **`aws.service_name` config** — allows overriding the SigV4 signing service name (currently hardcoded to `xray`). Different OTLP-compatible AWS endpoints require different signing names.
3. **`additional_headers` config** — a map of custom HTTP headers injected into every outgoing request. Some endpoints require routing or metadata headers.
4. **Region from `aws.region`** — fall back to endpoint URL parsing only when not explicitly set. Some endpoints don't embed the region in the hostname.
5. **Generalize the sink** from `Record<Span>` to `Record<Event>` so it can accept log events.

**Example config (logs):**

```yaml
otlp:
  endpoint: "https://my-otlp-endpoint.example.com:443/v1/logs"
  signal_type: logs
  aws:
    service_name: my-service
    region: us-east-1
    sts_role_arn: arn:aws:iam::123456789012:role/MyRole
  additional_headers:
    x-custom-header: my-value
```

### What alternatives have you considered?

Creating a separate sink plugin for each OTLP-compatible backend. This was rejected because the only differences are authentication parameters and signal type — not enough to justify a new plugin. The existing sink already handles OTLP/HTTP, SigV4, gzip, retry, and batching.

### Do you have any additional context?

- All changes are backward compatible — existing trace pipeline configs work unchanged
- The OTLP sink is marked `@Experimental`, so the `Record<Span>` → `Record<Event>` type change is acceptable
- This follows the same pattern as other Data Prepper sinks that support configurable AWS service names

LaszloKovacs9001:
### Description

This PR extends the OTLP sink plugin to support all three signal types (traces, metrics, logs) with full OTLP protocol encoding, configurable SigV4 signing, and custom HTTP headers.

Based on the work from @viquer in #6488 (who is currently on leave), with additional enhancements.

**Multi-signal support (from #6488):**
* Extended `OTelProtoCodec` with `convertToResourceMetrics()` and `convertToResourceLogs()`
* Added generic `OtlpSignalHandler<T>` interface pattern (Trace, Metric, Log handlers)
* Per-signal-type buffer architecture (`Map<OtlpSignalType, OtlpSinkBuffer>`)
* Generalized sink from `Record<Span>` to `Record<Event>`

**Design deviation from #6763 — no `signal_type` config:**
Issue #6763 originally requested a `signal_type` configuration option. After implementation, we chose **automatic signal type detection** from the event class at runtime (`OtlpSignalType.fromEvent()`) instead. This approach:
* Eliminates misconfiguration (no mismatch between config and actual event types)
* Supports mixed signal types in a single pipeline without additional config
* Simplifies the user experience — no config needed, it just works

The `signal_type` config field was implemented initially but removed as dead code since routing is fully handled by runtime type checking.

**Configurable SigV4 signing service name:**
* Added `service_name` field to `OtlpSinkConfig` (top-level, not under `aws:` block)
* `SigV4Signer` uses configurable service name (default: `xray`)

**Additional headers:**
* Added `additional_headers` map config to `OtlpSinkConfig`
* Headers injected into every outgoing request after SigV4 signing
* Protected headers (`Authorization`, `Host`, `Content-Type`, `Content-Encoding`, `Content-Length`, `X-Amz-*`) cannot be overridden
* Note: additional headers are not included in the SigV4 signature

**Region from AWS config:**
* `getAwsRegion()` checks `aws.region` first, falls back to endpoint URL parsing

**Per-signal metrics:**
* Added per-signal-type counters: `rejectedTracesCount`, `rejectedMetricsCount`, `rejectedLogsCount`, `failedTracesCount`, `failedMetricsCount`, `failedLogsCount`
* Aggregate counters (`rejectedRecordsCount`, `failedRecordsCount`) still available

**Breaking changes (plugin is `@Experimental`):**
* `OtlpSink` changed from `AbstractSink<Record<Span>>` to `AbstractSink<Record<Event>>`
* Metric counters renamed: `rejectedSpansCount` → `rejectedRecordsCount`, `failedSpansCount` → `failedRecordsCount`

**Example config (logs):**

```yaml
otlp:
  endpoint: "https://my-otlp-endpoint.example.com:443/v1/logs"
  service_name: my-service
  aws:
    region: us-east-1
    sts_role_arn: arn:aws:iam::123456789012:role/MyRole
  additional_headers:
    x-custom-header: my-value
  threshold:
    max_events: 100
    flush_timeout: 10s
```

### Issues Resolved

Resolves #6763 (with design deviation noted above — auto-detection instead of `signal_type` config)
Supersedes #6488

### Check List
* New functionality includes testing (100% JaCoCo line coverage)
* New functionality has a documentation issue: #12289
* New functionality has javadoc added
* Commits are signed with a real name per the DCO


dlvenable (changes requested):
Thank you for picking up the work from #6488 and building on it. The overall architecture is significantly improved — the `OtlpSignalHandler<T>` interface and signal-agnostic buffer design are exactly what I was looking for in that earlier review.

However, there are several issues that need to be addressed before this can be merged.

## Issues from #6488 review that are not fully resolved

### 1. `service_name` does not belong in shared `AwsConfig`

`org.opensearch.dataprepper.aws.api.AwsConfig` is a shared API class used by many plugins (SQS sink, CloudWatch Logs, Kafka, HTTP sink, OpenSearch source, etc.). Adding a `service_name` field here means every plugin that uses `AwsConfig` now silently accepts this property in their `aws:` block, even though it's meaningless for them.

Please move `service_name` into `OtlpSinkConfig` directly, or create an OTLP-specific config extension. The `OtlpSinkConfig` already has a `getServiceName()` method that delegates to `awsConfig.getServiceName()`, so this should be a straightforward refactor.

### 2. Test still uses `instanceof JacksonHistogram` conditional

In `OTelProtoEncoderRoundTripTest.java`, the `testHistogramWithoutExplicitBounds` test still has:
```java
if (metric instanceof JacksonHistogram) {
```
My comment on #6488 was: "Let's avoid conditions in tests" and "let's base it on `Histogram` and not `JacksonHistogram`." This should assert the expected type and use the `Histogram` interface, not silently skip if the metric is a different type.

### 3. Assert collection sizes before iterating

Several tests still iterate over collections without first asserting the expected size:
- `testMultipleMetricsRoundTrip` — iterates `records` then asserts counts at the end. If `records` is empty, the test silently passes.
- `testMultipleLogsRoundTrip` — iterates `logs` without asserting size.
- `testHistogramWithoutExplicitBounds` — iterates without size assertion (combined with the `instanceof` conditional, this test could pass with 0 assertions).

Assert the expected collection size before the loop.

### 4. Nested double/int gauge tests

The original request was to split gauge data point tests into separate `@Nested` classes for double vs. int values. I see `@Nested` used elsewhere, but the gauge value test handles both in a single test. This is minor — fine to defer.

## New issues

### 5. Issue #6763 mismatch — `signal_type` config

The issue requests a `signal_type` config option. The PR instead auto-detects signal type from the event class. The auto-detection approach is arguably better (eliminates misconfiguration, supports mixed signals), but the PR should explicitly note this design deviation in the description and on the issue. As-is, someone reading #6763 would expect a `signal_type` config that doesn't exist.

### 6. `additional_headers` can override signed headers

There's no validation preventing `additional_headers` from overriding critical headers like `Authorization`, `Host`, or `Content-Type`. Since these are added after SigV4 signing, overriding them could break authentication or cause unexpected behavior. Please add a blocklist of protected header names, or at minimum document the restriction.

### 7. Breaking metric rename — consider per-signal metrics

The `rejectedSpansCount` → `rejectedRecordsCount` rename is fine since the plugin is `@Experimental`. However, since this PR adds multi-signal support, per-signal metrics (e.g., `rejectedTracesCount`, `rejectedMetricsCount`, `rejectedLogsCount`) would provide much better observability. Without them, operators can't tell which signal type is failing. Consider adding per-signal counters.

### 8. License header CI failure

The CI bot flagged 5 test files for missing license headers. I can see the headers are present in the diff — this may be a stale CI run. Please verify by re-running the check, or check if the header format matches what the bot expects exactly.

## What looks good

- The `OtlpSignalHandler<T>` interface eliminates the switches from `OtlpHttpSender` — exactly what I asked for.
- The `Map<OtlpSignalType, OtlpSinkBuffer>` with signal-agnostic buffers is clean.
- Production code in `OTelProtoStandardCodec` correctly uses interfaces (`Gauge`, `Sum`, `Histogram`) instead of Jackson classes.
- `.gitignore` additions for Eclipse files are fine.
- `build.gradle` is clean — the unnecessary comment from #6488 is gone.
- The `README.md` is thorough.

LaszloKovacs9001:
@dlvenable Thanks for the thorough review. I've addressed all the feedback in the latest force-push (`412d9c0`):

1. **`service_name` moved out of `AwsConfig`** — now a top-level field in `OtlpSinkConfig`
2. **`instanceof JacksonHistogram` conditional removed** — test now asserts `Histogram` interface type and verifies encoded output
3. **Collection size assertions added** — `testMultipleMetricsRoundTrip`, `testMultipleLogsRoundTrip`, and `testHistogramWithoutExplicitBounds` all assert size before iterating
4. **Nested gauge tests** — deferred per your suggestion
5. **`signal_type` design deviation documented** — noted in PR description and [commented on #6763](https://github.com/opensearch-project/data-prepper/issues/6763#issuecomment-4329361972)
6. **Protected header blocklist added** — `additional_headers` rejects `Authorization`, `Host`, `Content-Type`, `Content-Encoding`, `Content-Length`, and `X-Amz-*`
7. **Per-signal metrics added** — `rejectedTracesCount`, `failedMetricsCount`, etc. alongside aggregate counters
8. **License headers verified** — all pass

dlvenable (dismissed):
Thank you @LaszloKovacs9001 !

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoCodec.java
```diff
@@ -150,6 +149,10 @@ Collection<Record<? extends Metric>> parseExportMetricsServiceRequest(
 
     public interface OTelProtoEncoder {
         ResourceSpans convertToResourceSpans(final Span span) throws UnsupportedEncodingException, DecoderException;
+        
+        io.opentelemetry.proto.metrics.v1.ResourceMetrics convertToResourceMetrics(final Metric metric) throws UnsupportedEncodingException, DecoderException;
```
kiran536: Nit: Why are we using full class name here?

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoStandardCodec.java
```diff
@@ -930,6 +936,471 @@ protected AnyValue objectToAnyValue(final Object obj) throws UnsupportedEncoding
...
+            final io.opentelemetry.proto.metrics.v1.Metric.Builder metricBuilder = 
+                io.opentelemetry.proto.metrics.v1.Metric.newBuilder();
+            
+            // Set common fields
+            metricBuilder.setName(metric.getName() != null ? metric.getName() : "");
+            metricBuilder.setDescription(metric.getDescription() != null ? metric.getDescription() : "");
+            metricBuilder.setUnit(metric.getUnit() != null ? metric.getUnit() : "");
+            
+            // Handle different metric types
+            if (metric instanceof Gauge) {
```
kiran536, line 1036: Nit: Can we use pattern switch case instead of if else loop?

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/http/SigV4Signer.java
```diff
@@ -24,13 +24,14 @@
  * before sending them to the AWS OTLP endpoint.
  */
 class SigV4Signer {
-    private static final String SERVICE_NAME = "xray";
+    private static final String DEFAULT_SERVICE_NAME = "xray";
     private static final String OTLP_PATH = "/v1/traces";
```
kiran536: Since endpoint is required field in config, can we get rid of this field and the corresponding default init? Because this is now confusing considering we are supporting metrics and logs

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/metrics/OtlpSinkMetrics.java
```diff
@@ -69,21 +69,43 @@ public void registerQueueGauges(final BlockingQueue<?> queue) {
...
-     * Increments the count of spans that were explicitly rejected by the OTLP endpoint.
+     * Increments the count of records that were explicitly rejected by the OTLP endpoint.
+     * This applies to all signal types: spans, metrics, and logs.
      *
-     * @param count The number of spans rejected.
+     * @param count The number of records rejected.
      */
-    public void incrementRejectedSpansCount(final long count) {
-        pluginMetrics.counter("rejectedSpansCount").increment(count);
+    public void incrementRejectedRecordsCount(final long count) {
```
kiran536, line 98: question: Do you think this will cause issue with the current metrics that are /were published. I am also concerned because we have allowlist on metrics vended. @dlvenable do you see any issue with changing the metric names?

kiran536 (changes requested):

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/http/SigV4Signer.java
LaszloKovacs9001: This was designed this way for backwards compatibility, do we just drop backwards compatibility?

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/metrics/OtlpSinkMetrics.java
dlvenable: This is a good point. It would be a breaking change and we shouldn't do this. 

Maybe we could rename the signal type enum to use `Spans` as the name here.

kiran536: Yeah, may be it is better to just use the plural forms for all enums! I was debating if I should add that as a Nit to the Enum saying if we are using Plural forms then use that through out

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/http/SigV4Signer.java
kiran536: Was there any ways the usage statement could have been reached earlier considering endpoint is mandatory? IMO it is safe to remove because it was functionally dead code previously

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoCodec.java
LaszloKovacs9001: Fixed. `ResourceMetrics` was already imported, so the FQ name was redundant. Added imports for `ResourceLogs` and `Log` and updated both method signatures to use short names.

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoStandardCodec.java
LaszloKovacs9001: The project targets Java 11 (`build.gradle`: `JavaLanguageVersion.of(11)`) — pattern matching switch requires Java 21. The if/else instanceof chain is the standard approach for this Java version.

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/http/SigV4Signer.java
LaszloKovacs9001: Agreed — removed the `OTLP_PATH` constant and the fallback URI entirely. As you pointed out, `endpoint` is `@NotBlank` so the fallback was functionally dead code. A null endpoint now fails fast at construction time.

data-prepper-plugins/otlp-sink/src/main/java/org/opensearch/dataprepper/plugins/sink/otlp/metrics/OtlpSinkMetrics.java
LaszloKovacs9001: Good catch, and thanks @dlvenable for the input. Updated the TRACE metrics label from `Traces` to `Spans` so per-signal counters emit `rejectedSpansCount`/`failedSpansCount` — preserving backwards compatibility with existing allowlists and dashboards. New signal types use `rejectedMetricsCount`/`rejectedLogsCount`. Removed the migration note from the README since there's no longer a breaking rename.

dlvenable (approved):

data-prepper-plugins/otel-proto-common/src/main/java/org/opensearch/dataprepper/plugins/otel/codec/OTelProtoStandardCodec.java
kiran536: Thanks. For some reason I remembered this as Java 17

kiran536 (approved):
LGTM after the changes

PR #6769, opened 2026-04-20
Add Confluence Data Center support with allow_local_address and beare…

srikanthpadakanti:
### Description
Two changes to support Confluence Data Center deployments:                                                                                        
   
  **1. Configurable address validation (`allow_local_address`)**                                                                                    
                                                            
  The existing `AddressValidation.validateInetAddress()` rejects all local/private IPs, which blocks Confluence Data Center running on internal     
  networks. Added a boolean config `allow_local_address` (defaults to `false`) that skips site-local, loopback, and link-local checks when enabled.
  Multicast and any-local addresses are still rejected regardless.                                                                                  
                                                                                                                                                                                                       
```
  source:
    confluence:                                                                                                                                     
      hosts:                                                                                                                                        
        - "https://confluence.internal.company.com"                                                                                                 
      allow_local_address: true                                                                                                                    

```                                                                                                                                                    
  2. Bearer token authentication                                                                                                                    
                                                                                                                                                    
  Confluence Data Center uses Personal Access Tokens (PAT) for authentication. Added bearer_token as a third auth option alongside basic and oauth2.
   Only one auth method can be configured at a time.        
                                                                                                                                                    
```
  source:                                                   
    confluence:
      hosts:                                                                                                                                        
        - "https://confluence.internal.company.com"
      allow_local_address: true                                                                                                                     
      authentication:                                       
        bearer_token: "your-personal-access-token"           
```                                                                                       
                                                                                                                                                    
Changes are in atlassian-commons so Jira Data Center also benefits from the same allow_local_address and bearer token support. 
 
### Issues Resolved
Resolves #6496
https://github.com/opensearch-project/data-prepper/issues/6496
 
### Check List
- [ X ] New functionality includes testing.
- [ ] New functionality has a documentation issue. Please link to it in this PR.
- [ X ] New functionality has javadoc added
- [ X ] Commits are signed with a real name per the DCO

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/data-prepper/blob/main/CONTRIBUTING.md).

PR #6770, opened 2026-04-20
Add lookup_private_addresses option to GeoIP processor for internal I…

### LINKED ISSUE #6079: GeoIP processor: Option to enable private IP address lookups
Currently, the GeoIP processor in Data Prepper ignores private IP addresses. This makes it difficult to enrich internal network traffic (e.g., NetFlow or audit logs) with metadata from an internal IP Address Management (IPAM) system, such as office, datacenter, or department information

Introduce a configuration option in the [geoip_service](https://docs.opensearch.org/latest/data-prepper/managing-data-prepper/extensions/geoip-service/) that allows processing of private IP addresses. For example:
```
extensions:
  geoip_service:
    maxmind:
      lookup_private_addresses: true
```

This would require an update to [GeoInetAddress.java](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/geoip-processor/src/main/java/org/opensearch/dataprepper/plugins/geoip/processor/GeoInetAddress.java) 

```
class GeoInetAddress {

    static Optional<InetAddress> usableInetFromString(final String ipAddress) {
        final InetAddress address;
        try {
            address = InetAddresses.forString(ipAddress);
        } catch (final IllegalArgumentException e) {
            return Optional.empty();
        }
        if (isPublicIpAddress(address) OR LOOKUP_PRIVATE_ADDRESS)
            return Optional.of(address);
        return Optional.empty();
    }
```

An alternative is to use the [add_entries](https://docs.opensearch.org/latest/data-prepper/pipelines/configuration/processors/add-entries/) processor in combination with add_when and  [cidrContains()](https://docs.opensearch.org/latest/data-prepper/pipelines/cidrcontains/)
However, this approach is significantly slower and becomes impractical with large numbers of CIDR ranges, especially for NetFlow enrichment.

**Additional context**
It is straightforward to enrich MMDB files with internal IPAM data using scripts like
[Enriching MMDB files with your own data using Go](https://github.com/maxmind/mmdb-from-go-blogpost)

As a further enhancement, it would be useful to allow reading custom fields from MMDB files (not only geo-related data). This would make it possible to attach metadata such as department or business unit information to a given CIDR block.

srikanthpadakanti:
### Description
 The GeoIP processor currently rejects all private IP addresses, which blocks users who enrich custom MMDB files with internal IPAM data (office, datacenter, department info) from using the processor for internal network traffic like NetFlow or audit logs.

This adds a `lookup_private_addresses` config option (default `false`) to the geoip processor. When set to `true`, private and loopback addresses are accepted for enrichment instead of being filtered out.
```

  processor:
    - geoip:
        lookup_private_addresses: true
        entries:
          - source: "/srcIP"
            target: "/geo"


```  

Three production files changed, three test files updated. No new dependencies. Fully backward compatible since the default is false.

 
### Issues Resolved
Resolves #6079
https://github.com/opensearch-project/data-prepper/issues/6079
 
### Check List


srikanthpadakanti:
Hi @dlvenable Please review this. Thanks.

data-prepper-plugins/geoip-processor/src/main/java/org/opensearch/dataprepper/plugins/geoip/processor/GeoIPProcessorConfig.java
```diff
@@ -52,6 +52,11 @@ public class GeoIPProcessorConfig {
     })
     private String whenCondition;
 
+    @JsonProperty("lookup_private_addresses")
```
dlvenable, line 55: I think this is a better name than the one from #6769.

The verb here is right - `lookup`. But we should aim for some consistency with the Confluence change.

See:

https://github.com/opensearch-project/data-prepper/pull/6769/changes#r3196227674

dlvenable (commented):
This approach does have slightly different semantics from #6769.

That code has:
```
if (address.isMulticastAddress() || address.isAnyLocalAddress()) {
  throw new BadRequestException(INVALID_URL);
}
```

So it is disallowing multicast and link-local. I'm not sure we need to disallow that here. But should we aim for closer semantic similarity between the two?

If we are going to allow these, include some unit test cases to make this explicit.

srikanthpadakanti:
> This approach does have slightly different semantics from #6769.
> 
> That code has:
> 
> ```
> if (address.isMulticastAddress() || address.isAnyLocalAddress()) {
>   throw new BadRequestException(INVALID_URL);
> }
> ```
> 
> So it is disallowing multicast and link-local. I'm not sure we need to disallow that here. But should we aim for closer semantic similarity between the two?
> 
> If we are going to allow these, include some unit test cases to make this explicit.

For GeoIP, multicast and link-local won't match MMDB entries so they're harmless to pass through unlike Confluence where connecting to those addresses is a security concern. Added explicit test cases for multicast and link-local to document this is intentional.

data-prepper-plugins/geoip-processor/src/main/java/org/opensearch/dataprepper/plugins/geoip/processor/GeoIPProcessorConfig.java
srikanthpadakanti: Kept as `lookup_private_addresses` — semantically distinct from the Confluence `allow_internal_address` since this controls MMDB lookup passthrough, not network connectivity validation.

dlvenable (approved):

dlvenable:
@srikanthpadakanti , Can you rebase this against `main` so that we can verify the tests pass? I pushed a fix yesterday to get the build working again.

srikanthpadakanti:
> @srikanthpadakanti , Can you rebase this against `main` so that we can verify the tests pass? I pushed a fix yesterday to get the build working again.

@dlvenable Merged main into the branch.

dlvenable (approved):

data-prepper-plugins/geoip-processor/src/main/java/org/opensearch/dataprepper/plugins/geoip/processor/GeoIPProcessorConfig.java
dlvenable: Yes, this makes sense.

kkondaka (approved):

